PR #18, opened 2015-10-27
exercise for the ternary operator

ConnyOnny:
Again I didn't add this to the Readme properly as I have no easy way of generating the required Rust playground link. I hope you like the exercise and don't mind adding it to the Readme.

### DIFF
--- a/ex6.rs
+++ b/ex6.rs
@@ -1,0 +1,51 @@
+fn bigger(a: i32, b:i32) -> i32 {
+    // Complete this function to return the bigger number!
+    // Do not use:
+    // - return
+    // - another function call
+    // - additional variables
+    // Scroll down for hints.
+}
+
+fn main() {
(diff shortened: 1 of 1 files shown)


carols10cents:
It's totally not a problem for me to add it to the readme, don't worry about it :)

Just to check, this exercise is meant to showcase the fact that there *isn't* a ternary operator in rust and you'll need to use if/else, right? Just making sure I understand the exercise's intent :)

ConnyOnny:
Yes it's meant to show how you can use Rust's if-then-else as a ternary operator because it is an expression and not (like in C or Java) a statement.

ConnyOnny:
Should we make this more clear in the hint comments?

carols10cents:
Yup, that's what I ended up doing -- I also decided this would be a great start to a section on `if`, so I moved it there!!! Thank you so much for your contribution!! :gift:

PR #19, opened 2015-11-17
Fix typo in hint for move_semantics1.rs.

toddthomas:
When I compile the unedited example in the playground, I get the
error on line 10, not line 8. That's the case for stable, beta,
and nightly. I'm a rust newbie, so for all I know the error is
supposed to be on line 8, but isn't due to some bug. I'm just
correcting the hint to match what I'm seeing.

Thank you for this project! I'm having fun and learning.

### DIFF
--- a/move_semantics/move_semantics1.rs
+++ b/move_semantics/move_semantics1.rs
@@ -37,6 +37,6 @@ fn fill_vec(vec: Vec<i32>) -> Vec<i32> {
 
 
 
-// So you've got the "cannot borrow immutable local variable `vec1` as mutable" error on line 8,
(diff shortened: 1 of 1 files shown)


carols10cents:
Good catch!!! I'm sorry I didn't see your bug report earlier, I don't know what happened. But thank you so much for the pull request!!!! And I'm glad you like the project, that means a lot to me!!! :heart: :heart: :heart:

toddthomas:
No problem--that wasn't my bug report. I just saw it when I was checking to see if anyone else had reported or made a PR for this. Thanks again.

PR #31, opened 2016-02-08
Moved away from deprecated thread::sleep_ms

StrongGoat:
Refactored to use thread::sleep and time::Duration.


StrongGoat:
Forgot to update the README link.

PR #37, opened 2016-04-09
Fix typo

jwworth:
Fixes 'playground'.

Thank you for making this— it's awesome!


carols10cents:
Good catch!!! :+1: :+1: :+1: And you're very welcome!! :heart:

PR #39, opened 2016-04-14
added exercise iterator1 and its solution

ConnyOnny:
This is already quite a complex iterator example. Maybe we should some day add a simpler one, call it iterator1 and rename this one to iterator2 or something like that.

The inspiration came from here:
* https://jethrogb.github.io/rust-combinators
* https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.collect


ConnyOnny:
No wait! Now the compiler is too clever and needs no type annotation at all.

PR #44, opened 2016-05-26
Update hint to reflect new variable binding name

Zajn:
The hint in the `strings2` exercise was referencing a variable binding from a previous commit.


carols10cents:
Good catch!! Thank you Zach!!! 🎈

PR #45, opened 2016-06-07
Update dead dining philosophers link to the 1.4.0 book example.

matthewjberger:
The link to the dining philosopher's problem was dead in the readme, so I changed it to point to the example from the 1.4.0 book (the most recent version that's still around) so that an explanation and implementation of the problem in rust is still provided.


carols10cents:
Good catch!!!!!! And awesome idea to link to the 1.4.0 docs :) I still think it's a good example.

Thank you!!! 💯

PR #47, opened 2016-06-14
Create tests for if1.rs

### LINKED ISSUE #46: if1.rs silently succeeds
The `if1.rs` rustling does not provide any feedback that you succeeded when you correctly implemented a solution.

Even though I understood why, it was a bit jarring to not seeing any form of output. Maybe especially so because I worked on `if1.rs` right after the test rustlings which give nice confirmation of their success and failure.

There are a few options to change the behavior of `if1.rs` if the current behaviour is deemed problematic. The ones I could think of are listed below.

1. **Make `if1.rs` a test**. *pro*: This provides instant standardized feedback for both correct and incorrect implementations. *con*: It could muddle the lesson about `if.
2. **print some output**. *pro*: user has confirmation if they are correct. *con*: It also could muddle the lesson about `if`.

I am willing an able to make changes to `if1.rs` and send a pull request for any improvements that are suggested. I am just opening this issue to discuss if it is a problem at all, and if it is a problem how to improve `if1.rs`.

dvberkel:
This provides standardize feedback if a solution is correct. Fixes #46.

### DIFF
--- a/if/if1.rs
+++ b/if/if1.rs
@@ -1,4 +1,4 @@
-fn bigger(a: i32, b:i32) -> i32 {
+pub fn bigger(a: i32, b:i32) -> i32 {
     // Complete this function to return the bigger number!
     // Do not use:
     // - return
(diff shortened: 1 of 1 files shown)


dvberkel:
@carols10cents is this what you were expecting?

carols10cents:
Yes!!! This is awesome, thank you so much!!! ❤️ ❤️ ❤️ ❤️ ❤️

PR #48, opened 2016-07-04
Update "strings2.rs" link

pindell-matt:
Add 1 newline to the "strings2.rs" link on rust playground so that the hints at the bottom are completely obscured on the initial page load.

Used the "Shorten Link" option from the Playground to create the new url.


carols10cents:
Ooooh good catch! I'm going to keep the link long though, since there's a script I can run on the code files to create them all automatically instead of having to make shortlinks, but I've got the code from your shortlink :) Thank you!!! ❤️

PR #49, opened 2016-08-15
corrected URLs to use https scheme

magicgoose:
https://play.rust-lang.org does support https, so it's better to link directly to the https version.
same for https://pnkfelix.github.io and https://doc.rust-lang.org


carols10cents:
Thank you!!! Sorry it took me so long to merge this 😿 ⌚ ⌛ ⏳

PR #54, opened 2017-01-03
ex1 is compiling

jan-schreib:
While trying your exercises I noticed that the ex1.rs code is already compiling.

### DIFF
--- a/ex1.rs
+++ b/ex1.rs
@@ -1,5 +1,5 @@
 // Make me compile!
 
 fn main() {
-    println!();
+    println();
 }


mwilli20:
This probably existed as in the past you could not give println! no arguments. https://github.com/rust-lang/rfcs/issues/995 . I think not knowing it is a macro is also a common mistake and this PR is still good. Just a different test for the user is all.

carols10cents:
Sorry this took me so long to get to!! Looks good!!!

PR #57, opened 2017-03-17
add macros examples

QuietMisdreavus:
Wanted to make some lessons for macros, since I've seen enough confusion about those. I added some hints; do you think they're good-but-not-direct enough?

I haven't added the examples to the readme, but that should be simple enough. I couldn't get `prlink` to build, though. `>_>`

### DIFF
--- a/macros/macros1.rs
+++ b/macros/macros1.rs
@@ -1,0 +1,62 @@
+// Make me compile! Scroll down for hints :)
+
+macro_rules! my_macro {
+    () => {
+        println!("Check out my macro!");
+    };
+}
+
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
Just added one more simple example that still doesn't have to get into what matches what. I figure this can be kept simple, and if people want to know more about macros they can check the book or TLBORM.

carols10cents:
Love it!! Nice and gentle :) I think the hints are perfect! 🌺 

> I couldn't get prlink to build, though. >_>

i got u fam!!!!

PR #62, opened 2017-09-03
Changed book links to second edition.

steveslatky:
Links to the book are now updated to the second edition.

Macros is still using the first book since there is no chapter on it in the second book yet. To indicate this I changed the wording of the link from:
Check out the Macros section of the book
To: 
Check out the Macros section of the first book


carols10cents:
Thank you!! Sorry it took me so long to merge this!!

PR #65, opened 2017-12-19
Tmux based idea

svanellewee:
An idea, not to merge, but as a starting point for a conversation..

This is just a suggestion of course, please kill this PR with fire if it is without merit.

I love the terminal, tmux and I feel  these things could be leveraged to answer https://github.com/carols10cents/rustlings/issues/21. 

The idea:
- the README.md and the rust example files in the repository stores the same rust code (either as en encoded URL or as actual files) Make only the code in the README the one true source of truth, delete the rest (Not done in this PR though) in the spirit of "Don't repeat yourself".
- Instead use a script to build the rust files as examples (automated cargo build)
- Then use tmux and entr (http://entrproject.org/) to run the code in one pane and see the effects immediately in the other pane. Changes to source triggers "cargo run" or "cargo 
test"

Thoughts welcome. Not the best script code (More a pythonista myself) However, hopefully the choice in scripting engine will make this script easy to run and at least prove the idea. We can always rewrite it in something else later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
+bin/rustlings/
+*~
 *.swp
--- a/bin/rustle.sh
+++ b/bin/rustle.sh
@@ -1,0 +1,112 @@
+# Allows you to change code and see what the rust compiler does immediately in the opposite pane
+
+# Instructions:
+# - ./rustle.sh
+# - select a rustling, see the split pane view show up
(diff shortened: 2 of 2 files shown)


svanellewee:
Just going to close PR incase someone actually presses the merge button by accident :-)

PR #70, opened 2018-03-04
Fix broken link to dining philosophers

asmw:


asmw:
I overlooked there's already a PR to fix all book links.

Issue #86, opened 2018-07-16
primitive_types3.rs: linked documentation no longer contains "fixed array with size" shorthand


wittekm:
primitive_types3.rs mentions:

```
// There's a shorthand to initialize Arrays with a certain size that does not 
// require you to type in 100 items (but you certainly can if you want!)
// Check out the Primitive Types -> Arrays section of the book:
// https://doc.rust-lang.org/stable/book/second-edition/ch03-02-data-types.html#arrays
```

I suspect that the shorthand here is
```
    let a: [i32; 100] = [0; 100];
```

but that shorthand is no longer mentioned in the linked data types page.

PR #87, opened 2018-07-20
update old book references to the second edition of the book

kosiachenko:
There is a bunch of book references in the hints within exercises that lead to the outdated version of the book (pages have a message "There is a new edition of the book and this is an old link." and require a bunch of digging to find the appropriate section of the book). This change updated the outdated links with the relevant ones, where possible. In one instance (primitive_types3.rs) the content is no longer available in the book, so I provided a direct hint on how to initialize an array of a specified size with values.


kosiachenko:
also fixes #86

oe:
thanks so much!

PR #90, opened 2018-08-16
Fixing main function errors in move_semantics/

spacekookie:
These examples didn't build because `pub fn main()` in them


oe:
can you update the compiled versions in the readme as well?

PR #106, opened 2019-01-10
rustlings v2

komaeda:
this replaces old rustlings with a cli-based version that runs on the same exercises

- [x] finish up readme
- [x] improve exercise structure (add little "test exercises" after a couple of chapters)
- [x] go over every book link and update it
- [x] refactor that one us-centric exercise


exercises/error_handling/README.md
```diff
@@ -0,0 +1,5 @@
+For this exercise check out the sections:
+- [Error Handling](https://doc.rust-lang.org/book/2018-edition/ch09-02-recoverable-errors-with-result.html) 
+- [Generics](https://doc.rust-lang.org/book/2018-edition/ch10-01-syntax.html)
```
phansch: The links to the book should be without `2018-edition` as the 2018 edition book doesn't exist anymore and is incorporated into the 'normal' book now: https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html for example.

komaeda: i'm planning to go over the links one by one before merging it in!

PR #107, opened 2019-01-13
use 2018 edition for rust playground

yerke:
If we don't specify the edition, Rust playground uses 2015 edition. This PR modifies links to playground to use the 2018 edition.


yerke:
oops, README file is autogenerated.

Issue #111, opened 2019-01-27
`rustlings watch` does not work


christiansakai:
The command `rustlings watch` does not watch and recompile.

Mac OXS 10.14.1 (Mojave)

christiansakai:
Yes I am, the `rustlings verify` work just fine. The `rustlings watch` just be in some suspended mode and does not recompile. I am using Rust 2018 if that helps

komaeda:
i don't think i can reproduce this :/

christiansakai:
That's ok. Maybe my environment is whacked up. Closing it now.

PR #117, opened 2019-03-06
Watch for file creation events in watch mode

shaunbennett:
Currently watch mode only watches for `DebouncedEvent::Chmod` and `DebouncedEvent::Write`. There are some edge cases where these events wont be called even when a file is modified.

For example, if the user is using vim with swap files enabled, it seems like only a `Remove` and then `Create` event is called for the rust file. Usually this is okay because there will also be a swp file being modified, but if the user has their swap files placed in a different directory, `rustlings watch` does not work.

This fixes the issue by also listening for the `Create` event, and only triggering for .rs files (so swp file changes are ignored).

I'm guessing that this may be related to #111, as I ran into the same issue and started looking into why watch wasn't working.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,6 +3,7 @@ use crate::verify::verify;
 use clap::{crate_version, App, Arg, SubCommand};
 use notify::DebouncedEvent;
 use notify::{RecommendedWatcher, RecursiveMode, Watcher};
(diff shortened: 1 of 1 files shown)


komaeda:
thanks for your contribution! :heart:

PR #120, opened 2019-03-13
Start verification at most recently modified file

abagshaw:
When in `watch` mode, it's a pain to have to wait for all previous exercises to compile before the one you just saved compiles...so now it skips compiling any files up and to the one most recently saved.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -84,7 +84,7 @@ fn watch() -> notify::Result<()> {
     let _ignored = verify(None);
 
     loop {
-        match rx.recv() {     
+        match rx.recv() {
             Ok(event) => match event {
                 DebouncedEvent::Create(b) | DebouncedEvent::Chmod(b) | DebouncedEvent::Write(b) => {
                     if b.extension() == Some(OsStr::new("rs")) {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -81,14 +81,16 @@ fn watch() -> notify::Result<()> {
...
 
     loop {
-        match rx.recv() {
+        match rx.recv() {     
             Ok(event) => match event {
                 DebouncedEvent::Create(b) | DebouncedEvent::Chmod(b) | DebouncedEvent::Write(b) => {
                     if b.extension() == Some(OsStr::new("rs")) {
-                        let _ignored = verify();
+                        println!("----------**********----------");
+                        println!();
```
vitkarpov: Is it a new line? Why not use `\n` in the macros above?

abagshaw: Fair enough, I was just following the precedent set here: https://github.com/rust-lang/rustlings/blob/6cd42bb8210b0e687942e388d66b07f286fe6bb6/src/main.rs#L39-L47

src/main.rs
```diff
@@ -81,14 +81,15 @@ fn watch() -> notify::Result<()> {
     let mut watcher: RecommendedWatcher = Watcher::new(tx, Duration::from_secs(2))?;
     watcher.watch("./exercises", RecursiveMode::Recursive)?;
 
-    let _ignored = verify();
+    let _ignored = verify(None);
 
     loop {
-        match rx.recv() {
+        match rx.recv() {
```
komaeda: could you trim the whitespace here? maybe running `cargo fmt` does this automatically, too

PR #123, opened 2019-03-17
Be nicer when rustlings isn't run from the right directory.

kisom:
Before, rustlings would panic if it wasn't in the right directory. It
took me a minute to figure out why, and this wasn't my first intro to
Rust. It would probably help new users if they saw a helpful message
instead of a stack trace.


komaeda:
thank you for this! i was thinking that it might be better to have this run on every `rustlings` execution, subcommand or not, since all of them require you to be in the correct directory. we could use something like `Path::new("info.toml").exists()` to check. how does that sound?

kisom:
That sounds great; I wasn't sure if they all required being in the right directory.

Issue #128, opened 2019-03-19
Error on install


jxxcarlson:
Oops, there is an error on Mac OS install rustlings:

```
 $ curl -L https://git.io/rustlings | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1350  100  1350    0     0   2318      0 --:--:-- --:--:-- --:--:--  2318
Let's get you set up with Rustlings!
Checking requirements...
SUCCESS: Git is installed
SUCCESS: Rust is installed
SUCCESS: Cargo is installed
Cloning Rustlings at rustlings/...
Checking out version 1.0.1...
Installing the 'rustlings' executable...

...

error: expected one of `(`, `,`, or `=`, found `::`
 --> /Users/carlson/.cargo/registry/src/github.com-1ecc6299db9ec823/onig_sys-69.1.0/src/constants.rs:1:16
  |
1 | #![allow(clippy::identity_op)]
  |                ^^ expected one of `(`, `,`, or `=` here

error: unexpected token: `identity_op`
 --> /Users/carlson/.cargo/registry/src/github.com-1ecc6299db9ec823/onig_sys-69.1.0/src/constants.rs:1:18
  |
1 | #![allow(clippy::identity_op)]
  |                  ^^^^^^^^^^^ unexpected token after this

error: aborting due to 2 previous errors

error: Could not compile `onig_sys`.
warning: build failed, waiting for other jobs to finish...
error: failed to compile `rustlings v1.0.1 (file:///Users/carlson/dev/rust/learn/rustlings)`, intermediate artifacts can be found at `/Users/carlson/dev/rust/learn/rustlings/target`

Caused by:
  build failed
````

komaeda:
looks like maybe you have an old version of rust installed? what rust version are you running (`rustc --version`), you need at least 1.31.0

ColinPitrat:
A check should be added in the script for the minimal version of Rust to provide a more user-friendly message when it's too old.

komaeda:
@ColinPitrat good idea!

ColinPitrat:
I'm having a look at how to do it :-)

komaeda:
closed in version 1.1.0! https://github.com/rust-lang/rustlings/releases/tag/1.1.0

PR #131, opened 2019-03-20
Verify that rust version is recent enough to install rustlings.

ColinPitrat:
This addresses issue #128.

I would have liked to write some tests for the vercomp function I
introduce, but there doesn't seem to be any CI setup yet?


komaeda:
@ColinPitrat yes, CI setup is something i'm looking at doing this evening actually! i'll let you know when

PR #134, opened 2019-03-27
fix watch command path execution

### LINKED ISSUE #126: "watch" is broken on windows
The issue seems to be that the call to verify(...) in main.rs:92 expects the path to be in Unix style (forward slashes instead of backslashes), so the later compare breaks.

I fixed this locally by just doing a string replace, but I guess a true Rustacean will fix this much more nicely so I didn't add a patch.

PS: It's also pretty inconvenient that "watch" doesn't continue with the next exercise once you fix the current one.

komaeda:
@hades32 @guttume could you test whether this works on windows by checking out the branch locally and running `cargo run watch`?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -88,7 +88,7 @@ fn watch() -> notify::Result<()> {
     let (tx, rx) = channel();
 
     let mut watcher: RecommendedWatcher = Watcher::new(tx, Duration::from_secs(2))?;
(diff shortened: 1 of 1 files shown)


guttume:
@komaeda I tried this branch also but it still sits on blank line when I make changes to my code. Including screenshot.
cmd
powershell

komaeda:
well, this still seems to fix _something_, so i'll merge it

@bors r+

PR #140, opened 2019-04-07
Fix intermittent test failure caused by race condition

cjpearce:
First public pull request 😬

There's an intermittent integration test failure when you use multiple test threads (at least for me on a mac). I narrowed it down to two tests each spawning a process using `Command` which then try to compile the same file at the same time. If the timing doesn't work out, they both try to compile, and then one process runs `clean` before the other can run the executable - causing a panic.

Screenshot 2019-04-07 at 19 54 55

You can prevent it from happening by running with a single thread (`cargo test -- --test-threads=1`), because the `Command` blocks. That's not a particularly good solution though because it's not something you can configure in `Cargo.toml`.

I considered making the affected tests just run serially, but it occurred to me that this could also happen if someone accidentally runs rustlings in watch mode in two terminals without realising it. I wound't consider this that unlikely given it's a tool for learning.

I fixed it by ensuring that the executables made from separate processes don't conflict by appending a process id to the output executable name. I also extracted the commands into a single file next to `clean` so that we don't have to repeat the generated file name everywhere and risk missing something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rust:
   - stable
   - beta
   - nightly
-script: cargo test --verbose -- --test-threads=1
+script: cargo test --verbose
 matrix:
   allow_failures:
     - rust: nightly
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,9 +1,8 @@
-use crate::util::clean;
+use crate::util;
 use crate::verify::test;
 use console::{style, Emoji};
 use indicatif::ProgressBar;
 use std::fs;
-use std::process::Command;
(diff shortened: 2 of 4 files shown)


komaeda:
Thanks for this! Do you want to edit the Travis build source (https://github.com/rust-lang/rustlings/blob/master/.travis.yml#L6) as well so that we can directly test if this works?

cjpearce:
Yeah of course! I didn't think of checking the CI script. Thanks

komaeda:
Thanks!

@bors r+

PR #143, opened 2019-04-11
Canonicalize paths to fix path matching

### LINKED ISSUE #126: "watch" is broken on windows

cjpearce:
This PR should fix #126. The main solution to the issue was using `canonicalize()` on the paths we create for the exercises from `info.toml` and any user-specified paths, so that path `ends_with` matching will work correctly.

As adding calls to the canonicalize function everywhere requires unwrapping, I also decided to extract a struct representing an exercise and use serde to deserialize the paths from the `info.toml` file up front. I also tried to move the path handling out into the `exercise.rs` file and down into `main.rs` so that it doesn't create as much clutter. There was already a lot of unwrapping and path handling in the other files and I felt like it was getting a bit too repetitive.

If the approach is going too far (too many changes etc.) I'm happy to try to produce a smaller PR that fixes the bug without any refactoring.


cjpearce:
After doing some more testing, I've discovered that I've created a panic when an incorrect filename is supplied. I'll tidy up the error handling to avoid this.

src/exercise.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Display for Exercise {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "{}", self.path.to_str().unwrap())
+    }
+}
+
+#[test]
+fn test_clean() {
+    std::fs::File::create(&temp_file()).unwrap();
```
komaeda: I think it's fine to `use` this

src/exercise.rs
```diff
@@ -0,0 +1,72 @@
...
+    let exercise = Exercise {
+        path: PathBuf::from("example.rs"),
+        mode: Mode::Test,
+    };
+    exercise.clean();
+    assert!(!std::path::Path::new(&temp_file()).exists());
```
komaeda: same as above

src/exercise.rs
cjpearce: I've tried to clean that up a bit by adding a use declaration in a test module. Putting the use at the top gives 'unused' warnings when doing a normal build. I couldn't spot a clear convention for this situation online but I'm open to suggestions if you have any ideas. I'm mainly trying to keep that test alive since it was in the original `utilities.rs` file.

PR #163, opened 2019-05-25
Adds a simple exercise for structures

briankung:
Thanks for rustlings! Here's a small contribution in return.

### DIFF
--- a/exercises/structs/README.md
+++ b/exercises/structs/README.md
@@ -1,0 +1,7 @@
+### Strings
+
+Rust has three struct types: a classic c struct, a tuple struct, and a unit struct.
+
+#### Book Sections
+
+- [Structures](https://doc.rust-lang.org/rust-by-example/custom_types/structs.html)
--- a/exercises/structs/structs1.rs
+++ b/exercises/structs/structs1.rs
@@ -1,0 +1,46 @@
+// structs1.rs
+// Address all the TODOs to make the tests pass!
+
+struct ColorClassicStruct {
+    // TODO: Something goes here
+}
+
+struct ColorTupleStruct(/* TODO: Something goes here */);
+
+struct ColorUnitStruct;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn classic_c_structs() {
+        // TODO: Instantiate a classic c struct!
+        // let green =
+
+        assert_eq!(green.name, "green");
+        assert_eq!(green.hex, "#00FF00");
+    }
+
+    #[test]
+    fn tuple_structs() {
+        // TODO: Instantiate a tuple struct!
+        // For more fun, use the field initialization shorthand.
+        // let green =
+
+        assert_eq!(green.0, "green");
+        assert_eq!(green.1, "#00FF00");
+    }
+
+    #[test]
+    fn unit_structs() {
+        // TODO: Instantiate a unit struct!
+        // let green =
+
+        if let ColorUnitStruct = green {
+            assert!(true);
+        } else {
+            assert!(false);
+        }
+    }
+}
--- a/info.toml
+++ b/info.toml
@@ -76,6 +76,12 @@ mode = "compile"
 path = "exercises/primitive_types/primitive_types6.rs"
 mode = "compile"
 
+# STRUCTS
+
+[[exercises]]
+path = "exercises/structs/structs1.rs"
+mode = "test"
+
 # TESTS
 
 [[exercises]]


Emiller88:
I'm having an issue with this. I'm not sure if correct answers should be discussed here or not. It may also be because of my lack of experience.
```rust
    #[test]
    fn unit_structs() {
        // TODO: Instantiate a unit struct!
        let green = ColorUnitStruct;

        if let ColorUnitStruct = green {
            assert!(true);
        } else {
            assert!(false);
        }
    }
```
Results in
```bash
error[E0162]: irrefutable if-let pattern
  --> exercises/structs/structs1.rs:41:16
   |
41 |         if let ColorUnitStruct = green {
   |                ^^^^^^^^^^^^^^^ irrefutable pattern
```

briankung:
@Emiller88 would you mind sharing that entire test?

Emiller88:
@briankung Sorry I should have mentioned it's in the `exercises/structs/structs1.rs` file
https://github.com/briankung/rustlings/blob/9b92aa08aead17d7737137f1c5c382548797b83e/exercises/structs/structs1.rs

briankung:
@Emiller88 I understand, sorry I must have had a brain fart (or there was an edit to your comment) but I didn't think I had the entirety of your failing test case. What version of Rust are you using?

Emiller88:
@briankung `rustc 1.32.0-nightly (14997d56a 2018-12-05)`

briankung:
Weird. Well, seeing as I'm a Rust newbie myself, I can't really say why this doesn't work on your machine. It does seem to make sense - after all once it knows the type of `green`, it will always match.

You can try throwing a `#[allow(irrefutable_let_patterns)]` at the top of the test function. Just to be clear, are you using `$ rustlings run exercises/structs/struct1.rs` to run this? And have you run the `rustlings` command on other exercises? For whatever reason, it seems to allow irrefutable let patterns by default for me (`rustc 1.37.0-nightly (02564de47 2019-06-10)`) as well as dead code in a bunch of exercises.

Emiller88:
I was using `rustlings watch` but it doesn't work with `rustlings run exercises/structs/struct1.rs` either.

`#[allow(irrefutable_let_patterns)]` doesn't help either.

briankung:
Odd. Unlikely to be related, but what OS are you using? I'm also confused why the allow attribute doesn't work since it [works in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3d58db8968e2e6b37e3e0bb7311ed5ec). And your test looks like the below?

```rust
    #[test]
    #[allow(irrefutable_let_patterns)]
    fn unit_structs() {
        // ...
    }
```

Emiller88:
Yes it did, the `#[allow(irrefutable_let_patterns)]` works in the playground however [without it the test still fails](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b94d896c7d1c19918dc58edceae3ff77) so I think it's an issue.

briankung:
I can't recreate the error so I can't reliably fix it, but how would you
like to proceed?

Emiller88:
Does this playground work for you? It's exactly what's in the `exercises/structs/structs1.rs` 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b94d896c7d1c19918dc58edceae3ff77

briankung:
The tests pass with a warning. I think the playground output should not
vary by client machine.

Emiller88:
Ah I didn't see it was a warning in the playground. When it runs with a test(At least in my version) the warning turns into an error when the test runs. If it's just unique to my version I think we're good here.

ksceriath:
I am using 
rustc 1.32.0 (9fda7c223 2019-01-16)

Facing the same issue as @Emiller88 above. This doesn't appear to be machine specific. I feel we should modify the test around this, else 'rustlings watch' will not function.

ksceriath:
This doesn't error after updating to rustc 1.35.0 (3c235d560 2019-05-20)

briankung:
It seems odd that it would error in one but not the other. I think the test should change, it's just especially odd that something like an irrefutable let pattern would change from version to version of Rust.

@ksceriath does the `#[allow(irrefutable_let_patterns)]` allow it to compile without errors? I think you can use this snippet to override the toolchain in a particular directory:

```
$ rustup override set 1.32.0
```

From: https://doc.rust-lang.org/stable/edition-guide/rust-2018/rustup-for-managing-rust-versions.html#managing-versions)

I'm trying it myself, as well.

briankung:
@Emiller88 and @ksceriath I've opened https://github.com/rust-lang/rustlings/pull/179, please take a look and see if those changes work for you. Sorry for the inconvenience!

ksceriath:
@briankung This works with 1.32 / 1.35. Thank you!

Emiller88:
@briankung Works on both for me as well thanks for fixing it!

Extrawurst:
going through the tests one by one, i am wondering why this is the first that has no help to scroll down to. even the link to the docs makes this exercise suddenly much harder than the previous

PR #170, opened 2019-06-07
fix(iterators): Rename iterator3.rs

### LINKED ISSUE #155: iterator example names inconsistent
The iterator tests in `standard_library_types` are named `iterator3.rs` (singular) and `iterators4.rs` (plural). They should probably match.

miller-time:
fixes #155

### DIFF
--- a/exercises/standard_library_types/iterators3.rs
+++ b/exercises/standard_library_types/iterators3.rs
@@ -1,4 +1,4 @@
-// iterator3.rs
+// iterators3.rs
 // This is a bigger exercise than most of the others! You can do it!
 // Here is your mission, should you choose to accept it:
 // 1. Complete the divide function to get the first four tests to pass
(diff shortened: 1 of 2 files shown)


komaeda:
Ah, one small nit for all of your pull requests: Could you format them according to the Angular commit message standards? (https://gist.github.com/stephenparish/9941e89d80e2bc58a153#format-of-the-commit-message)

So, for this, something like `fix(iterators): Rename iterator3.rs`

miller-time:
@komaeda done!

I see that clog was just added within the last couple days, and this commit format isn't mentioned in the README. Also, there is no CONTRIBUTING.md file so when you create a pull request there are no project-specific guidelines for formatting the PR, etc. 

I could take a stab at this, or else I would like to at least suggest that instructions be added somewhere. 🙂

komaeda:
@miller-time Yep, haven't had time to write up any guidelines yet, unfortunately. The `fixes: <issue #>` is a good idea. I think I have some time right now, so I'll write up a simple `CONTRIBUTING.md`.

PR #171, opened 2019-06-07
chore: Clean up some formatting in exercises

miller-time:
I noticed some formatting that isn't consistent with the `rustfmt` style and tried my best to run it on the files in the exercises directory (which does fail for files that can't compile!), which just caught some minor whitespace things here and there.

Note: also can't just apply `rustfmt` blindly because of the blank lines that lead to the hint comments


miller-time:
@komaeda what would the `<scope>` be for this pull request's title?

komaeda:
`chore: Clean up some formatting in exercises`

PR #172, opened 2019-06-07
chore: Fix a couple broken book links

miller-time:
While I was fixing these, I figured maybe it's good to just use a consistent URL for book links: https://doc.rust-lang.org/book/

Is there a plan for what to do with this old 1.4.0 example that doesn't exist in the current version of the book? Perhaps at least include a disclaimer when recommending that folks check it out?

https://github.com/rust-lang/rustlings/blob/34e31232dfddde284a341c9609b33cd27d9d5724/exercises/threads/README.md#L1


miller-time:
@komaeda what would the `<scope>` be for this pull request's title?

komaeda:
`chore: Fix a couple broken book links`

PR #177, opened 2019-06-14
Introduce fibonacci exercise

au-phiware:
Every good programming exercise book should include the fibonacci numbers!

One possible solution:

    (0..i).fold((0, 1), |(a, b), _| (b, a + b)).1

### DIFF
--- a/exercises/standard_library_types/iterators5.rs
+++ b/exercises/standard_library_types/iterators5.rs
@@ -1,0 +1,61 @@
+// iterators5.rs
+
+pub fn fibonacci(i: u64) -> u64 {
+    // Complete this function to return the ith fibonacci number
+    // Do not use:
+    // - return
+    // For extra fun don't use:
+    // - imperative style loops (for, while)
+    // - let
+    // For the most fun don't use:
+    // - recursion
+    // Scroll down for hints.
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn fibonacci_of_1() {
+        assert_eq!(1, fibonacci(1));
+    }
+    #[test]
+    fn fibonacci_of_2() {
+        assert_eq!(2, fibonacci(2));
+    }
+
+    #[test]
+    fn fibonacci_up_to_10() {
+        assert_eq!(
+            "[1, 1, 2, 3, 5, 8, 13, 21, 34, 55]",
+            format!("{:?}", (0..10).map(fibonacci).collect::<Vec<u64>>())
+        );
+    }
+}
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+// In an imperative language you might write a for loop to iterate through
+// add and keep track of the last two values into a mutable variables.
+// Or you might write code more functionally with recursion and a match clause.
+// But you can also use ranges and iterators to solve this in rust.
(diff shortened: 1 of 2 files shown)


exercises/standard_library_types/iterators5.rs
```diff
@@ -0,0 +1,61 @@
+// iterators5.rs
+
+pub fn fibonacci(i: u64) -> u64 {
+    // Complete this function to return the ith fibonacci number
+    // Do not use:
+    // - return
+    // For extra fun don't use:
```
komaeda: Let's say "challenge" instead of "fun" here?

au-phiware: Ha! This is a copy and paste job of the previous exercise (iterators4). I'm happy to change it; shall I change both?

Actually, I think I to rethink the wording altogether, it fails to recognise the similarity with the previous exercise and doesn't flow very well, leading to a bad user experience...

komaeda: That would be great!

darnuria:
Hi! Is it ok if I open a PR to complete this one of course if @au-phiware is ok ? I would fix the rebase issue, match hints to current system and do some rephrasing to keep hints and UX similar of iterator4. :)

exercises/standard_library_types/iterators5.rs
```diff
@@ -0,0 +1,61 @@
...
+
+
+
+
+
+
+
+
+
+
```
AnnikaCodes, line 46: Why is there so much whitespace here?

darnuria: It's due to the old format of hints (they were down below) I fixed it in my branch but before publishing a PR, I prefers to give some time to the original author.

dzmitry-lahoda:
@au-phiware could you please update pr to  mainstream?

darnuria:
@dzmitry-lahoda still insterested in rebasing and updating content of the PR to today or to redevelop it if licensing is a concern.

dzmitry-lahoda:
fibonacci can be implemented:
1. recursive (recursion)
2. for loop with tuple to show swap elements via tuple patters and procedural loop
3. your solutios
4. recursive but parallel via threads
5. via channels (is rust std still has some?)
6. via matrices (2d array in rust)
7. approximate with floats and golden ration

so each task could ask for specific impl to teach area of rust

Zerotask:
@au-phiware I like your idea. Do you plan to update your PR?

au-phiware:
Apologies for my late reply, in truth I had completely forgotten that I had opened this PR and have only just seen this thread!  I am happy for others to continue this work, if it needs reading then just do it 😁

au-phiware:
> fibonacci can be implemented:
> 1. recursive (recursion)
> 2. for loop with tuple to show swap elements via tuple patters and procedural loop
> 3. your solutios
> 4. recursive but parallel via threads
> 5. via channels (is rust std still has some?)
> 6. via matrices (2d array in rust)
> 7. approximate with floats and golden ration
> 8. introduce 128 bit number and 8 bit number (input 8 bit and output of 128 bit) to show limits of calculation, show how rust handles integer overflow
> 
> so each task could ask for specific impl to teach area of rust 
> 

The spirit of my original PR was just a simple function with an iterator, the first two suggestions are okay but I think for the others it is better to have something closer to a real world example.  Also, I don't think parallelism will work since each successive value requires the previous value to be computed first.

PR #186, opened 2019-07-02
feat(iterators2): adds iterators2 exercise including config

Jesse-Cameron:
Hi there!

I really enjoyed doing the rustlings exercises so I thought that I would try to add an exercise!

This one just covers a couple of basic iterator operations. Getting people used to the `map` and `collect` functions.

However, it does feel kinda similar to the next exercise. So I may also revisit some of the tests in iterators3 if we think that is necessary.

### DIFF
--- a/exercises/standard_library_types/iterators2.rs
+++ b/exercises/standard_library_types/iterators2.rs
@@ -1,0 +1,145 @@
+// iterators2.rs
+// In this module, you'll learn some of unique advantages that iterators can offer
+// Step 1. Complete the `capitalize_first` function to pass the first two cases
+// Step 2. Apply the `capitalize_first` function to a vector of strings, ensuring that it
+// Step 3. Apply the `capitalize_first` function again to a list, but try and ensure it returns a single string
+// As always, there are hints below!
+
+pub fn capitalize_first(input: &str) -> String {
+    let mut c = input.chars();
+    match c.next() {
+        None => String::new(),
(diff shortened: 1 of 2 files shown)


exercises/standard_library_types/iterators2.rs
```diff
@@ -0,0 +1,145 @@
+// iterators2.rs
+// In this module, you'll learn some of unique advantages that iterators offers
```
komaeda: `that iterators offer`

exercises/standard_library_types/iterators2.rs
```diff
@@ -0,0 +1,145 @@
...
+
+
+
+
+
+
+// Step 3.
+// This is very similar to the previous test. The only real change is that you need to
+// alter the type that collect is coerced into. For a bonus you could try doing this with a
+// turbofish
```
komaeda: could you add a newline at the end here?

exercises/standard_library_types/iterators2.rs
```diff
@@ -0,0 +1,145 @@
+// iterators2.rs
+// In this module, you'll learn some of unique advantages that iterators offers
+// Step 1. Complete the `capitalize_first` function to pass the first two cases
+// Step 2. Apply this function to a vector of strings, make sure that it returns a list
```
komaeda: I would explicitly mention `capitalize_first` in both step 2 and step 3 here

PR #192, opened 2019-07-13
fix(test1): Swap assertion parameter order

petemcfarlane:
`Expected` should come before `actual`, other wise it leads to confusing compiler messages, e.g.
```
note: expected type `()`
         found type `{integer}`
```
There may be other tests that need updating, but this is as far as I am through the Rustlings course right now :)


petemcfarlane:
Screenshot 2019-07-13 at 13 25 56

Issue #160, opened 2019-05-18
rustlings verify accepts option1.rs prematurely


marienz:
For `rustlings verify`, mode "compile" really only means "compile": it does not run the resulting program. This means option1.rs (which has a main() that panics) passes before I've completed the exercise. I almost missed the entire exercise, as I was working through them based on the errors from `rustlings verify`.

If this behavior from `rustlings verify` is intentional, `option1.rs` needs to be changed to a "test" exercise (with either main() itself as test function or a trivial test function that calls main()).

Alternatively, `rustlings verify` could start running the code too. Changing it to use `compile_and_run` works but makes the output a little noisy (did confirm I didn't miss any other tests before option1.rs), so it might need a new or refactored function that only prints the output on failure.

Alternatively alternatively, rustlings could grow an "exec" mode for exercises like this one.

cjpearce:
I totally agree that there needs to be a better way of going through these sorts of exercises. It feels like a lot of the exercises are meant for exploring how the language works, and it even appears that the intention is for them to be run multiple times.

As an example `primitive_types3.rs` gives the user feedback using print statements, which users will never see if they use verify like the intro message suggests.

```
    let a = ???

    if a.len() >= 100 {
        println!("Wow, that's a big array!");
    } else {
        println!("Meh, I eat arrays like that for breakfast.");
}
```

It seems like it might be quite difficult to get verify working in a way that runs commands for the user without becoming really messy like you mentioned, but I could imagine the `rustlings watch` command printing the output last exercise to compile and then waiting for user input before moving to the next exercise.

It might make sense at that point to point people towards the `rustlings watch` command in the intro message, and maybe even give it a different name to make it feel slightly more accessable.

I'd happily put something like that together if it sounded like a reasonable solution. Other people have been confused by the way the exercises are used as well (#127).

marienz:
I hadn't tried `rustlings watch` because I obsessively save my files all the time and figured triggering `watch` on each save would be distracting (and demotivating because I'd get lots of errors for partially completed examples), and also because it wasn't clear to me `rustlings watch` actually goes through the exercises in order. I thought `rustlings watch` wouldn't do anything until I edited something after starting it, so `verify` seemed more convenient.

So yeah, maybe the only problem here is that the homepage recommends `verify` (and mentions `watch` as an alternative "If you don't want to rerun verify every time you change a file").

[...]

Issue #127, opened 2019-03-18
Why not use tests for some exercises


Hades32:
There are several exercises of the style

```
if you_did_it_right {
  print("great");
} else {
  print("try again");
}
```

Why aren't these exercises done as tests?

MForever78:
Echo on this. some exercises can now pass by just not having compiling errors.

mo8it:
Closing but feel free to open new issues pointing to specific exercises. You are also welcome to submit pull requests :D

PR #209, opened 2019-08-18
fix(primitive_types4): Fail on a slice covering the wrong area

Dylnuge:
I noticed this issue and it seems like a similar one was raised/fixed in #160 this way. This is my first contribution to this repo (or any Rust project) so let me know if I messed up or need to fix anything!

---
This commit converts primitive_types4 to a test and asserts that the
slice given is equal to the expected slice.

The intent of the primitive_types4 exercise appears to be to ensure the
user understands inclusive and exclusive bounds as well as slice syntax.
`rustlings` commands using `compile` do not verify that a specific
println is reached and, in the case of `watch` and `verify` (but not
`run`), they do not output the `println`s at all.

This fix is semantically similar to #198. It does not take a stance on
the correct way to handle this for all exercises; see #127. There are
likely other exercises whose intent are masked by this issue.

### DIFF
--- a/exercises/primitive_types/README.md
+++ b/exercises/primitive_types/README.md
@@ -6,3 +6,4 @@ compiler. In this section, we'll go through the most important ones.
 #### Book Sections
 
 - [Data Types](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html)
+- [The Slice Type](https://doc.rust-lang.org/stable/book/ch04-03-slices.html)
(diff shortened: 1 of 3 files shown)


Dylnuge:
@komaeda I'm happy to take a pass at moving all the other exercises from `compile` to `test` if that seems like the right way to go about it. I wasn't sure if there was another option that might be preferable (e.g. panicking when hitting a failed condition)

ngolin:
This is misleading. Now how to get the code compiled? Slice can't compare array, type mismatching.

PR #217, opened 2019-09-07
Tense in README

timthelion:
Don't see why run should be past tense.  At least in my native General American dialect it shouldn't be.


fmoko:
thanks!

@bors: r+

PR #219, opened 2019-10-21
Added exercise for struct update syntax

vyaslav:
Added one exercise for struct update syntax `struct2.rs`


vyaslav:
@fmoko please have a look

PR #220, opened 2019-10-23
Add windows install script

gdoenlen:
Adds a powershell install script for windows users that are at least
running powershell 5. It is almost a direct port of install.sh.
This would be used to automatically download the script much like
curl | bash, but with:

`Invoke-WebRequest https://urltoscript.com | Select-Object
-ExpandProperty Content | Out-File $env:TMP/install_rustlings.ps1;
Unblock-File $env:TMP/install_rustlings.ps1; Invoke-Expression
$env:TMP/install_rustlings.ps1`

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,0 +1,92 @@
+#!/usr/bin/env pwsh
+
+#Requires -Version 5
+param($path = "$pwd/rustlings")
+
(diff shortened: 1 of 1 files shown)


fmoko:
Thanks!

@bors: r+

PR #224, opened 2019-10-28
feat: Add enums exercises

nyxtom:
Creates an exhaustive list of enum exercises. This goes through the basics of different ways to derive enums with mixed data type variants, as well as the use of the all important `match` operator.


nyxtom:
I’ll go ahead and update the exercises. Though the book itself uses ipaddr and ipkind as a point of interest (this is mainly what I followed here)

nyxtom:
@fmoko I ended up just simplifying the implementations to just stick with the `Message` enumeration example since it's the most consistent use case (i.e. commands).

nyxtom:
Does @bors need to rerun to remove the label `waiting-on-author`?

fmoko:
Merging this now.

@bors: r+

PR #230, opened 2019-11-11
Changes the execution mode for `watch`, asking for user input

jrvidal:
We've [observed](https://hackmd.io/-cK6aPhnTwiCiI7u6k0xug?both) that learners can get confused when they do get everything right, but they _still_ get errors... which come from the next exercise, no the one they just edited.

This PR changes it so they have to confirm they want to move forward by removing the `I AM NOT DONE` comment.

Screenshot at 2019-11-11 15:13:39

* [ ] The particular string is of course subject to bikeshed.

### Alternatives/doubts
* The coolest solution I could imagine would involve a proc-macro attribute `#![ready(false)]` that they could edit once they're done, but it's a bit complicated to set up.
* For now I've put `I AM NOT DONE` everywhere, I think it's what make more sense.

### DIFF
--- a/exercises/variables/variables1.rs
+++ b/exercises/variables/variables1.rs
@@ -1,6 +1,13 @@
 // variables1.rs
 // Make me compile! Scroll down for hints :)
 
+// About this `I AM NOT DONE` thing:
+// We sometimes encourage you to keep trying things on a given exercise,
+// even after you already figured it out. If you got everything working and
+// feel ready for the next exercise, you the `I AM NOT DONE` comment below.
+
+// I AM NOT DONE
+
 fn main() {
     x = 5;
     println!("x has the value {}", x);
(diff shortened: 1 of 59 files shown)


exercises/variables/variables1.rs
```diff
@@ -1,6 +1,13 @@
 // variables1.rs
 // Make me compile! Scroll down for hints :)
 
+// About this `I AM NOT DONE` thing:
+// We sometimes encourage you to keep trying things on a given exercise,
+// even after you already figured it out. If you got everything working and
+// feel ready for the next exercise, you can just delete this comment!
```
fmoko: This comment, and the following one

tests/integration_tests.rs
```diff
@@ -105,3 +108,20 @@ fn run_single_test_no_exercise() {
         .assert()
         .code(1);
 }
+
+#[test]
+fn all_exercises_require_confirmation() {
```
jrvidal: Checks that all exercises have the "annotation".

Not really checking the supported format exhaustively, it should at least prevent the common case where somebody completely forgets to add it in a new exercise.

PR #241, opened 2019-11-18
feat(watch): show hint while watching

vyaslav:
`rustlings hint ...` command is not convenient when doing exercises with `rustlings watch`.  
This PR makes it possible for user to type `hint` while running `watch`  and get hint text for exercise which is currently failing.
e.g.
```rust
...
  --> exercises/variables/variables1.rs:13:36
   |
13 |     println!("x has the value {}", x);
   |                                    ^ not found in this scope

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0425`.

type 'hint' to get help:
hint
Hint: The declaration on line 12 is missing a keyword that is needed in Rust
to create a new variable binding.
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,9 +6,12 @@ use notify::DebouncedEvent;
 use notify::{RecommendedWatcher, RecursiveMode, Watcher};
 use std::ffi::OsStr;
 use std::fs;
+use std::io;
 use std::path::Path;
 use std::process::{Command, Stdio};
 use std::sync::mpsc::channel;
+use std::sync::{Arc, Mutex};
+use std::thread;
 use std::time::Duration;
 
 mod exercise;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -107,6 +110,25 @@ fn main() {
         println!("{}", text);
     }
 }
+fn spawn_watch_shell(failed_exercise_hint: &Arc<Mutex<Option<String>>>) {
+    let failed_exercise_hint = Arc::clone(failed_exercise_hint);
+    println!("type 'hint' to get help");
```
fmoko: Please use proper capitalization here, e.g. `Type 'hint' to get help`

src/main.rs
```diff
@@ -107,6 +110,25 @@ fn main() {
         println!("{}", text);
     }
 }
+fn spawn_watch_shell(failed_exercise_hint: &Arc<Mutex<Option<String>>>) {
```
fmoko: An extra line before this would be nice!

src/main.rs
vyaslav: good catch, I have fixed this and other issue.

PR #249, opened 2019-12-16
feat: Add type conversion and parsing exercises

AbdouSeck:
This pull request adds exercises for converting values into specific types. The exercises uses string to struct type conversions, but most of the traits in the exercises can handle more than just string parsing and conversions.

The following traits are covered:
1. `From` and `Into`
2. `TryFrom` and `TryInto`
3. `AsRef`
4. `FromStr`

The `as` operator is also covered.

### DIFF
--- a/exercises/conversions/README.md
+++ b/exercises/conversions/README.md
@@ -1,0 +1,20 @@
+### Type conversions
+
+
+Rust offers a multitude of ways to convert a value of a given type into another type.
+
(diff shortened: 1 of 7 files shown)


fmoko:
Thanks!

@bors: r+

fmoko:
@AbdouSeck Looks like you're missing the "I AM NOT DONE" comment in your exercises!

AbdouSeck:
@fmoko I just added the lines to the exercise files. Please let me know if anything else looks amiss.

Thanks,

Abdou

PR #252, opened 2019-12-22
fix(docs): Added a necessary step to Windows installation process

Socratides:
Added an essential step to Windows installation process (Setting ExecutionPolicy to RemoteSigned because it wouldn't install otherwise).


fmoko:
Thanks!

PR #253, opened 2019-12-22
feat: Show a completion message when watching

### LINKED ISSUE #251: rustlings watch does nothing when all exercises are complete
I just completed everything in 2.1.0 and `rustlings watch` showed that my threads1 solution compiled, but no other output was shown. I assumed my solution was running but hit a deadlock or something, but 15 minutes later I realized `watch` does nothing when you're done

It'd be nice to display a congratulatory message so we can feel accomplished and know when we're done! 😸

codehearts:
The completion message is shown only once all exercises succeed and are
not annotated with "I AM NOT DONE." The watch command will also exit

closes #251

Let me know if there are any tests I could add or if the completion message should be tweaked!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,6 +2,7 @@ use crate::exercise::{Exercise, ExerciseList};
 use crate::run::run;
 use crate::verify::verify;
 use clap::{crate_version, App, Arg, SubCommand};
+use console::Emoji;
 use notify::DebouncedEvent;
 use notify::{RecommendedWatcher, RecursiveMode, Watcher};
 use std::ffi::OsStr;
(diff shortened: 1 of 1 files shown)


AbdouSeck:
Hi @codehearts,

Thanks for the PR. I noticed that the `src/verifiy.rs` module was changed to add `println!` lines in the `verify` function, and also `std::process::exit(0)` is used inside the `watch` function in `main.rs`. While that's great, both the printing and process exiting could have been done inside the `main` function. Since both the `verify` and `watch` functions have clear return types and values, I thought the following would be easier:

- Have no printing in the module `verify` since it can be used for something that may not require any printing
- Have the `watch` return `Ok(())` whenever `verify_result.is_ok()` instead of `std::process::exit(0)` (since the function can be used for something else that may not require process exiting)

Doing the above makes it so that you can do the following inside the `main` function:

```
if matches.subcommand_matches("watch").is_some() {
        if let Ok(()) = watch(&exercises) {
            println!("Done with all exercises.");
            // More printing
            std::process::exit(0);
        }
    }
```
Please keep in mind that these are just suggestions. You're welcome to keep the PR intact. I just thought it might help to move process exiting steps into the `main` function, so that it is clear that `main` is the driver of the binary.

Thank you,

Abdou

codehearts:
I agree with your suggestions, I’ll rebase with that implementation as soon as I can

codehearts:
I've rebased your suggestions into the original commit, since I think your approach was better. Let me know if there's anything else you see that might need some tweaking!

AbdouSeck:
Hi @codehearts looks great to me!

fmoko:
thank you so much!

@bors: r+

PR #264, opened 2020-01-14
Add variables5 to introduce shadowing

### LINKED ISSUE #260: Add hint to variable shadowing for variables3.rs
I think a nice addition to variables3.rs would be to also mention variable shadowing as it also solves the exercise and also gets explained in chapter 3 of the book (https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html)

Any opinions on that one?

TorbenJ:
closes #260


fmoko:
:+1:

PR #269, opened 2020-02-14
feat: Add clippy lints

Tarnadas:
This is a feature PR which adds the possiblity to create clippy exercises.

Clippy has many awesome linting rules, which can give a deeper understanding about the Rust programming language, therefor I made this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ target/
 **/*.rs.bk
 .DS_Store
 *.pdb
+exercises/clippy/Cargo.toml
+exercises/clippy/Cargo.lock
--- a/src/exercise.rs
+++ b/src/exercise.rs
@@ -1,7 +1,7 @@
 use regex::Regex;
 use serde::Deserialize;
 use std::fmt::{self, Display, Formatter};
-use std::fs::{remove_file, File};
+use std::fs::{self, remove_file, File};
 use std::io::Read;
 use std::path::PathBuf;
 use std::process::{self, Command};
@@ -9,6 +9,7 @@ use std::process::{self, Command};
 const RUSTC_COLOR_ARGS: &[&str] = &["--color", "always"];
 const I_AM_DONE_REGEX: &str = r"(?m)^\s*///?\s*I\s+AM\s+NOT\s+DONE";
 const CONTEXT: usize = 2;
+const CLIPPY_CARGO_TOML_PATH: &str = "./exercises/clippy/Cargo.toml";
 
 fn temp_file() -> String {
     format!("./temp_{}", process::id())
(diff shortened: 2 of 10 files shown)


src/exercise.rs
```diff
@@ -58,6 +60,24 @@ impl Exercise {
...
+edition = "2018"
+[[bin]]
+name = "{}"
+path = "{}.rs""#,
+                    self.name, self.name, self.name
+                );
+                fs::write(CLIPPY_CARGO_TOML_PATH, cargo_toml)
+                    .expect("Failed to write 📎 Clippy 📎 Cargo.toml file.");
+                Command::new("cargo")
+                    .args(&["clippy", "--manifest-path", CLIPPY_CARGO_TOML_PATH])
```
fmoko: This would imply that you have to have Clippy installed, right? As far as I know, Clippy isn't added as a default Rustup component, so you'd have to mention somewhere that you need to explicitly do `rustup component add clippy`

.gitignore
```diff
@@ -2,3 +2,5 @@
 target/
 **/*.rs.bk
 .DS_Store
+exercises/clippy/Cargo.toml
```
fmoko: I'm assuming the separate Cargo.toml is so that Clippy doesn't lint the entire project?

Tarnadas: Yes, I was hoping that there is an option to invoke Clippy without Cargo and without a Cargo.toml and just pass it the bin with an option, but it seems that this is not possible.
This is why I'm generating a custom Cargo.toml just to invoke Clippy with it

src/exercise.rs
Tarnadas: You're right.
I added the Clippy installation to the install scripts. I assume this should be good then?

Tarnadas:
> This is a good idea! I think that we should have a bit more information about how to run these kinds of exercises, probably in the comments of `clippy1.rs`. Something explaining what exactly Clippy is would help too.

I added a section to the first Clippy example. I'm not sure if this is enough or if a README should be added.

fmoko:
>I'm not sure if this is enough or if a README should be added.

I would say just to be sure, add a README file that explains that you need Clippy installed for these exercises (but that, if you'd used the installation script, it would have already installed it for you). Also, a link to more information about Clippy would be good to have in that file.

Other than that, looks good to me!

Tarnadas:
I fixed the conflicts and added a README

liby:
@Tarnadas 
Sorry to interrupt. I have a question about this PR. I want to modify this [line of code](https://github.com/rust-lang/rustlings/blob/1cd9328a5872f1d6a95b02824e1af33a423c047a/src/exercise.rs#L120), in order to achieve the purpose of modifying the `Cargo.toml` file generated under the `exercises/clippy` folder. 
But I found that no matter how I modify it, I will not change the contents of the `exercises/clippy/Cargo.toml` file. May I ask why? If there is an obvious answer to the question I ask, I'm sorry I can't see it.

Issue #270, opened 2020-02-19
`output`-based exercise


jrvidal:
I've been wondering whether it'd be useful to have a 3rd kind of exercise, `"output"`/`Mode::Output`, that compares the stdout of an exercise to an expected value before moving on to the next one. This would imply a minor re-invention the `#[test]` framework, since we'd need to produce our own `Expected`/`Actual` output and stuff. 

However, I feel that it would be a good fit for some `println`-based exercises, where the user can simply manipulate the program until it generates the desired output in a way that feels more natural to me than using `#[test]`. For instance, a hypothetical 0-exercise with an empty `main()` that expects you to print "Hello world". Or some of the `if`-based exercises, where the user can test branches until the correct one is executed. In fact, I suspect it'd be a good fit for most of the exercises for which we get complaints about them passing too soon.

@fmoko What do you think? If that sounds OK I'd be glad to send a patch.

fmoko:
@jrvidal That sounds good! It's probably a good replacement for `#[test]`-based exercises, except for the ones that teach you about `#[test]`.

PR #271, opened 2020-02-20
refactor: exercise evaluation

jrvidal:
After working a bit on #270, I realized that it'd be useful to first perform a minor refactor of exercise evaluation.

* Now we have standard methods to compile + execute that return `Result`s.
* Success/failure messages are standardized.

### DIFF
--- a/src/exercise.rs
+++ b/src/exercise.rs
@@ -4,7 +4,7 @@ use std::fmt::{self, Display, Formatter};
 use std::fs::{remove_file, File};
 use std::io::Read;
 use std::path::PathBuf;
-use std::process::{self, Command, Output};
+use std::process::{self, Command};
 
 const RUSTC_COLOR_ARGS: &[&str] = &["--color", "always"];
 const I_AM_DONE_REGEX: &str = r"(?m)^\s*///?\s*I\s+AM\s+NOT\s+DONE";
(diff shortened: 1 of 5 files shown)


src/exercise.rs
```diff
@@ -47,9 +47,34 @@ pub struct ContextLine {
...
+
+struct FileHandle;
+
+#[derive(Debug)]
+pub struct ExerciseOutput {
+    pub stdout: String,
+    pub stderr: String,
+}
+
+impl Drop for FileHandle {
```
jrvidal: This is a silly trick to avoid calling `clean()` all over the place. I think it's not racier than what we were doing before.

src/verify.rs
```diff
@@ -15,78 +15,78 @@ pub fn verify<'a>(start_at: impl IntoIterator<Item = &'a Exercise>) -> Result<()
...
+            warn!(
+                "Testing of {} failed! Please try again. Here's the output:",
                 exercise
             );
-            println!("{}", style(formatstr).red());
-            println!("{}", String::from_utf8_lossy(&runcmd.stdout));
-            exercise.clean();
-            Err(())
+            println!("{}", output.stdout);
+            return Err(());
```
jrvidal: Hm, `return` not needed.

src/exercise.rs
fmoko: :+1:

jrvidal:
Awesome. I do have the `mode="output"` code ready, I'll submit after this one.

PR #272, opened 2020-02-21
fix(installation): make fatal errors more obvious

QuintenJohnson:
I initially ran the installation script without rust installed. The fact that the error message was labeled with WARNING made me unsure whether installation was successful or I needed to re-run after installing rust. There's an error code returned on fatal errors, but this change will make things clearer.


fmoko:
Thanks!

@bors: r+

PR #276, opened 2020-02-25
Enable test for exercise test4

Stigjb:
Enable test mode for exercise test4.rs, and add a new test case. This exercise was easy to "pass" by e.g. writing a macro which returned a constant string literal, unless you ran `main` and looked at the output.


fmoko:
Thanks!

PR #278, opened 2020-02-26
feature: adds "output" mode, resolves #270

### LINKED ISSUE #270: `output`-based exercise

jrvidal:
* New: `mode = "output`.
* If `output = "foobar"` is specified for an `"output"` exercise, it runs a "test" against the expected output.
* An `"output"` exercise prints its output in the `🎉 Move on when you're ready` screen. 

<details>

<summary> Screenshots </summary>

Screenshot at 2020-02-26 21:52:49
Screenshot at 2020-02-26 21:52:23

</details>

After playing a while with this, I'm not entirely sold on the idea :sweat_smile: I think having expected output _could_ be useful in a very exploratory exercise like "Get `main` to print `Hello world!` to the screen". But, in general, I feel like `"test"` exercises guide you better toward the answer. Take `structs1.rs` for instance and its hypothetical alternative:

<details>

<summary> <code>structs1.rs</code> </summary>

```rust
struct ColorClassicStruct {
    // TODO: Something goes here
}

struct ColorTupleStruct(/* TODO: Something goes here */);

#[derive(Debug)]
struct UnitStruct;

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn classic_c_structs() {
        // TODO: Instantiate a classic c struct!
        // let green =

        assert_eq!(green.name, "green");
        assert_eq!(green.hex, "#00FF00");
    }

    #[test]
    fn tuple_structs() {
        // TODO: Instantiate a tuple struct!
        // let green =

        assert_eq!(green.0, "green");
        assert_eq!(green.1, "#00FF00");
    }

    #[test]
    fn unit_structs() {
        // TODO: Instantiate a unit struct!
        // let unit_struct =
        let message = format!("{:?}s are fun!", unit_struct);

        assert_eq!(message, "UnitStructs are fun!");
    }
}
```

</details>

<details>

<summary> <code>structs1.rs</code> with <code>"output"</code> </summary>

```rust
struct ColorClassicStruct {
    // TODO: Something goes here
}

struct ColorTupleStruct(/* TODO: Something goes here */);

#[derive(Debug)]
struct UnitStruct;

fn main() {
    // TODO: Instantiate a classic c struct!
    // let green =
    println!("green.name = {}", green.name);
    println!("green.hex = {}", green.hex);

    // etc.
}
```

</details>

I don't think it works quite as well, since the learner has no clue what the value of `green.name` should be until they get the program to compile and see the Expected/Actual error message. And it'll be way harder to reach that point without the clue on what the fields of `ColorClassicStruct` should be. I suspect we could blame the exercise itself. Others, like `if1.rs`, might work better.

What I do think it works great is non-expected output exercises. Most of the `"compile"` exercises felt a bit anti-climatic. Once you get the program to compile, you can still play with it thanks to `I AM NOT DONE`, but you don't _see_ the results of your tweaks. Even if we don't adopt `mode="output"`, I think the default behavior of `"compile"` should be this.

So, summary of options:
1. Scrub this, make `"compile"` always print output.
2. Keep this PR as it is. It might be useful in future exercises, or maybe you have a strong opinion that some `mode = "test"` exercises should be converted.
    * If we keep it, we could still make `"compile"` always print as in (1) and
    reserve `mode = "output"` for mandatory expected output (i.e. change `Mode::Output(Option<String>)` to `Mode::Output(String)`).

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -1,4 +1,4 @@
-use crate::exercise::{Exercise, Mode, State};
+use crate::exercise::{CompiledExercise, Exercise, Mode, State};
 use console::style;
 use indicatif::ProgressBar;
 
@@ -6,7 +6,7 @@ pub fn verify<'a>(start_at: impl IntoIterator<Item = &'a Exercise>) -> Result<()
     for exercise in start_at {
         let compile_result = match exercise.mode {
             Mode::Test => compile_and_test(&exercise, RunMode::Interactive),
-            Mode::Compile => compile_only(&exercise),
+            Mode::Compile => compile_and_run_interactively(&exercise),
             Mode::Clippy => compile_only(&exercise),
         };
         if !compile_result.unwrap_or(false) {
@@ -30,23 +30,37 @@ fn compile_only(exercise: &Exercise) -> Result<bool, ()> {
     let progress_bar = ProgressBar::new_spinner();
     progress_bar.set_message(format!("Compiling {}...", exercise).as_str());
     progress_bar.enable_steady_tick(100);
-    let compilation_result = exercise.compile();
+
+    let _ = compile(&exercise, &progress_bar)?;
     progress_bar.finish_and_clear();
 
-    match compilation_result {
-        Ok(_) => {
-            success!("Successfully compiled {}!", exercise);
-            Ok(prompt_for_completion(&exercise))
-        }
+    success!("Successfully compiled {}!", exercise);
+    Ok(prompt_for_completion(&exercise, None))
+}
+
+fn compile_and_run_interactively(exercise: &Exercise) -> Result<bool, ()> {
+    let progress_bar = ProgressBar::new_spinner();
+    progress_bar.set_message(format!("Compiling {}...", exercise).as_str());
+    progress_bar.enable_steady_tick(100);
+
+    let compilation = compile(&exercise, &progress_bar)?;
+
+    progress_bar.set_message(format!("Running {}...", exercise).as_str());
+    let result = compilation.run();
+    progress_bar.finish_and_clear();
+
+    let output = match result {
+        Ok(output) => output,
         Err(output) => {
-            warn!(
-                "Compilation of {} failed! Compiler error message:\n",
-                exercise
-            );
-            println!("{}", output.stderr);
-            Err(())
+            warn!("Ran {} with errors", exercise);
+            println!("{}", output.stdout);
+            return Err(());
         }
-    }
+    };
+
+    success!("Successfully ran {}!", exercise);
+
+    Ok(prompt_for_completion(&exercise, Some(output.stdout)))
 }
 
 fn compile_and_test(exercise: &Exercise, run_mode: RunMode) -> Result<bool, ()> {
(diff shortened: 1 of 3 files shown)


src/verify.rs
```diff
@@ -101,11 +137,21 @@ fn prompt_for_completion(exercise: &Exercise) -> bool {
         Mode::Compile => "The code is compiling!",
         Mode::Test => "The code is compiling, and the tests pass!",
         Mode::Clippy => "The code is compiling, and 📎 Clippy 📎 is happy!",
+        Mode::Output(..) => "The code is compiling and the output is correct!",
```
jrvidal: This message doesn't make sense for `Mode::Output(None)` :woman_facepalming:

tests/integration_tests.rs
```diff
@@ -70,6 +70,26 @@ fn run_single_test_success() {
         .success();
 }
 
+#[test]
```
jrvidal: ### Why aren't there tests for failure cases?

The behavior of `rustlings run` for a `mode="output"` exercise with expected output is to simply run the program without running the "test harness". So I can't actually test the test harness itself :disappointed: I suppose this could be an argument in favor of running the comparison on `rustling run`

tests/integration_tests.rs
```diff
@@ -70,6 +70,26 @@ fn run_single_test_success() {
         .success();
```
jrvidal: @fmoko One little detail that I noticed: the `verify_all_failure` test case misled me a bit, since it only tests that _some_ exercise fails to verify. Should we change the name?

src/verify.rs
```diff
@@ -72,26 +73,62 @@ fn compile_and_test(exercise: &Exercise, run_mode: RunMode) -> Result<bool, ()>
...
+                Ok(())
+            } else {
+                Err((expected, &output.stdout))
+            }
+        }
+    };
+
+    match comparison_result {
+        Ok(_) => {
+            success!("Successfully tested {}!", exercise);
```
jrvidal: Re-added this line, a regression introduced in https://github.com/rust-lang/rustlings/pull/271 :see_no_evil:

fmoko:
I might be a month late, but I think the first option sounds good for now. How do you wanna continue with this, make the changes in this PR or open a new one?

jrvidal:
I can make the changes in this PR :+1:

src/verify.rs
jrvidal: Still re-added in the last version.

fmoko:
Looks good now! I'm trying to figure out how we wanna release this... it's definitely a breaking change, and there's some SemVer minor things laying around on master, so we could either do a minor release and then a major one a bit after or just coagulate all of it into a major release. What do you think?

jrvidal:
Hm, usually I'd be inclined to push a minor and _then_ a major, so people can avoid breakage if they want. However, given the way I assume most people consume rustlings (they clone it, work on it for a while, and probably never update the repo), I think it'd be fine to just release a major.

PR #280, opened 2020-02-28
feat: added generics exercises

sjmann:
I've added a few exercises on generics since I was looking for some while I was going through rustlings alongside the book. I hope they can be of use (and that generics wasn't a deliberate omission 😄)!

### DIFF
--- a/exercises/generics/generics3.rs
+++ b/exercises/generics/generics3.rs
@@ -6,6 +6,7 @@
 // Make the necessary code changes to support alphabetical report cards, thereby making the second
 // test pass.
 
+// I AM NOT DONE
 pub struct ReportCard {
     pub grade: f32,
     pub student_name: String,
--- a/exercises/generics/generics3.rs
+++ b/exercises/generics/generics3.rs
@@ -1,0 +1,45 @@
+// An imaginary magical school has a new report card generation system written in rust!
+// Currently the system only supports creating report cards where the student's grade 
+// is represented numerically (e.g. 1.0 -> 5.5). However, the school also issues alphabetical grades
+// (A+ -> F-) and needs to be able to print both types of report card!
+
+// Make the necessary code changes to support alphabetical report cards, thereby making the second
+// test pass.
+
+pub struct ReportCard {
+    pub grade: f32,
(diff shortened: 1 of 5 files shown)


info.toml
```diff
@@ -624,6 +624,38 @@ Try mutating the incoming string vector.
...
+Think carefully about what we need to do here. Currently we are wrapping only values of
+type 'u32'. Maybe we need to update the explicit references to this data type somehow?
+"""
+
+[[exercises]]
+name = "generics3"
+path = "exercises/generics/generics3.rs"
+mode = "test"
+hint = """
+To find the best solution to this challenge you're going to need to think back to your
```
sjmann: I'm not 100% sure about this exercise. It does require knowledge of traits (which is the section before this one), it seems kind of logical to me as the areas are quite related, but I don't know how insulated you like the exercises to be. Maybe it could be a candidate to be extended slightly and made into a new `test`?

fmoko: I'd say use it for now and readjust it if it turns out to not work that well!

fmoko:
First of all, I'm so sorry for being so late on this. 

Second of all, this looks really good! I'm going to merge it, thanks so much!

exercises/generics/generics3.rs
```diff
@@ -0,0 +1,46 @@
+// An imaginary magical school has a new report card generation system written in rust!
+// Currently the system only supports creating report cards where the student's grade 
+// is represented numerically (e.g. 1.0 -> 5.5). However, the school also issues alphabetical grades
+// (A+ -> F-) and needs to be able to print both types of report card!
+
+// Make the necessary code changes to support alphabetical report cards, thereby making the second
+// test pass.
```
brettclanton001, line 7: Most of the other exercises explicitly mention the command to get the hint.

Example: https://github.com/rust-lang/rustlings/blob/master/exercises/functions/functions1.rs#L2

Not mentioning it makes it seem like there is no hint. :)

Karuma303: Rustling here and I totally agree. I did the exercise a couple of minutes ago and was confused that the command for the hint missing ;)

PR #284, opened 2020-03-10
Adds additional test to meet exercise rules

sl4m:
This PR adds an additional test to meet this exercise [rule](https://github.com/rust-lang/rustlings/blob/master/exercises/conversions/from_into.rs#L33).

### DIFF
--- a/exercises/conversions/from_into.rs
+++ b/exercises/conversions/from_into.rs
@@ -70,4 +70,11 @@ mod tests {
         assert_eq!(p.name, "Mark");
         assert_eq!(p.age, 20);
     }
+    #[test]
+    fn test_bad_age() {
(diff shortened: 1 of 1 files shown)


exercises/conversions/from_into.rs
```diff
@@ -70,4 +70,10 @@ mod tests {
         assert_eq!(p.name, "Mark");
         assert_eq!(p.age, 20);
     }
+    #[test]
+    fn test_bad_age() {
+        let p = Person::from("Mark,twenty");
```
fmoko: A comment like in the tests above would be nice here!

sl4m: Done

PR #290, opened 2020-04-06
feat: Add Option2 exercise

sanjaykdragon:
I added another exercise for Option<T>, as promised.
Let me know if there are any problems with this.
This one is specifically about if let / while let

### DIFF
--- a/exercises/option/option2.rs
+++ b/exercises/option/option2.rs
@@ -5,7 +5,7 @@
 
 fn main() {
     let optional_value = Some(String::from("rustlings"));
-    //make this an if let statement - value is "Some" type
+    // Make this an if let statement whose value is "Some" type
     value = optional_value {
         println!("the value of optional value is: {}", value);
     } else {
--- a/exercises/option/option2.rs
+++ b/exercises/option/option2.rs
@@ -18,7 +18,7 @@ fn main() {
     }
 
     // make this a while let statement - remember that vector.pop also adds another layer of Option<T>
-    // you can stack Option<T>'s into while let and if let
+    // You can stack `Option<T>`'s into while let and if let
     value = optional_values_vec.pop() {
         println!("current value: {}", value);
     }
--- a/exercises/option/option2.rs
+++ b/exercises/option/option2.rs
@@ -1,0 +1,21 @@
+// option2.rs
+// Make me compile! Execute `rustlings hint option2` for hints
+
+// I AM NOT DONE
+
+fn main() {
+    let optional_value = String::from("rustlings");
+    if let Some(value) = optional_value {
+        println!("the value of optional value is: {}", value);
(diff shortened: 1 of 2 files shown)


sanjaykdragon:
@fmoko

sanjaykdragon:
> 
> 
> Hmm, I think this exercise could be structured in a different way. As I see it, you just need to wrap the variables in `Some` to pass right now, whereas I think a better approach would be to let the user write their own `if let` and `while let` statements. Also, you should add a couple of comments about which lines the user should try to edit to nudge them in the right direction.

Alright, this is a better idea. I was thinking of doing that, but I felt that leaving the while let / if let sections blank would result in people just changing the entire code. Will change the style now

exercises/option/option2.rs
```diff
@@ -0,0 +1,25 @@
+// option2.rs
+// Make me compile! Execute `rustlings hint option2` for hints
+
+// I AM NOT DONE
+
+fn main() {
+    let optional_value = Some(String::from("rustlings"));
+    //make this an if let statement - value is "Some" type
```
fmoko, line 8:
```suggestion
    // Make this an if let statement whose value is "Some" type
```

exercises/option/option2.rs
```diff
@@ -0,0 +1,25 @@
...
+    value = optional_value {
+        println!("the value of optional value is: {}", value);
```
fmoko, line 10:
```suggestion
        println!("The optional value contains: {}", value);
```

exercises/option/option2.rs
```diff
@@ -0,0 +1,25 @@
...
+    } else {
+        println!("optional value does not have a value!");
```
fmoko, line 12:
```suggestion
        println!("The optional value doesn't contain anything!");
```

exercises/option/option2.rs
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    let mut optional_values_vec: Vec<Option<i8>> = Vec::new();
+    for x in 1..10 {
+        optional_values_vec.push(Some(x));
+    }
+
+    // make this a while let statement - remember that vector.pop also adds another layer of Option<T>
```
fmoko, line 20:
```suggestion
    // Make this a while let statement - remember that `pop` also adds another layer of `Option<T>`
```

exercises/option/option2.rs
```diff
@@ -0,0 +1,25 @@
...
+    // you can stack Option<T>'s into while let and if let
```
fmoko, line 21:
```suggestion
    // You can stack `Option<T>`'s into while let and if let
```

sanjaykdragon:
added your nits

fmoko:
Thanks!

Issue #281, opened 2020-03-04
Follow `rustlings watch` with editor by opening source files in the same order


agentofuser:
First of all, thank you so much for rustlings, this is an amazing project! A well thought-out practice test suite is worth a thousand tutorials.

I got lazy from having to close nvim and open it on the next file to keep up with `rustlings watch`, so I came up with a little kludge:

```bash
cd rustlings # local rustlings repo
rg '^path = "' info.toml | rg '"(.*)"' --only-matching --replace='$1' --null | xargs -J {} nvim '{}'
```

This is the same as typing:

```bash
nvim file1 file2 file3 ...
```

When you do this, nvim opens one file at a time, and you can navigate between them with `:previous` and `:next`.

So after passing the tests for one exercise, I delete the `// I AM NOT DONE` line, then type `:wn` on nvim, which saves the file (`w` is for "write") and opens the next one (`n`).

That's enough keystroke savings for me, but if this sounds like an interesting idea, maybe it could be turned into a `rustlings watch editor` command that calls `$EDITOR` in succession with the file that's currently failing `rustlings watch`, so both can be kept running side by side in sync.

nealmcb:
Great idea!  And thanks for the script.  But I guess my version of rg is out-of-date, since I got
`Unknown flag: '--only-matching'`

Here's my version, which works with emacs and should work any time `$EDITOR` is defined:

```
files=$(sed -n 's/^path = "\([^"]*\)"/\1/p' info.toml)
# Open each file in sequence if it seems not done, wait for it to be closed, then continue
for f in $files; do if grep -q "// I AM NOT DONE" $f; then $EDITOR $f; fi; done
```

Next, I'd like to see a solution that, when appropriate, opens the latest `README.md`

Issue #259, opened 2020-01-02
as_ref_mut.rs has two I AM NOT DONE


steelwil:

AbdouSeck:
Hi @steelwil, thanks for reporting this. The idea behind the two **I AM NOT DONE** lines is to indicate that there are two separate tasks in the file to finish. If it seems too cumbersome, the second one can definitely disappear. Let me know what you think.

Thank you,

Abdou

steelwil:
I think keeping things consistent is important, writing a different exercise may be a better idea.

AbdouSeck:
Hi @steelwil, I don't think it's necessary to write a different exercise. I would be fine with removing the second "I AM NOT DONE" occurrence.

Tuxified:
:wave: I think this issue can be closed as it has been changed by #292 ?

PR #292, opened 2020-04-07
Remove duplicate not done comment

Tuxified:
As indicated in #259 , I found it confusing to have 2 comments as the code wouldn't compile unless I solved both issues (I used the script from #281 to remove a comment and use `:wn` to go to next exercise, hence this tripped me).


sanjaykdragon:
IMO the comment should be moved into the block above

Tuxified:
I choose the bottom one 'cause it was closer to the code I was editing ( I used vim, hence easier to spot ), but I'll switch to the comment on top if that's the convention

fmoko:
@Tuxified that sounds good!

Tuxified:
Ok, I made the switch to remove the bottom one. Any other remarks/questions ? Do you agree with this change ?

PR #295, opened 2020-04-08
fix(run) Compile clippy exercise files 

crodjer:
Additionally to running clippy, also compile the exercise file so that
`rustling run clippy1` works after a successful completion of the
exercise.

### DIFF
--- a/src/exercise.rs
+++ b/src/exercise.rs
@@ -98,6 +98,15 @@ path = "{}.rs""#,
                 );
                 fs::write(CLIPPY_CARGO_TOML_PATH, cargo_toml)
                     .expect("Failed to write 📎 Clippy 📎 Cargo.toml file.");
+                // To support the ability to run the clipy exercises, build
+                // an executable, in addition to running clippy. With a
+                // compilation failure, this would silently fail. But we expect
+                // clippy to reflect the same failure while compiling later.
+                Command::new("rustc")
+                    .args(&[self.path.to_str().unwrap(), "-o", &temp_file()])
+                    .args(RUSTC_COLOR_ARGS)
+                    .output()
+                    .expect("Failed to compile!");
                 // Due to an issue with Clippy, a cargo clean is required to catch all lints.
                 // See https://github.com/rust-lang/rust-clippy/issues/2604
                 // This is already fixed on master branch. See this issue to track merging into Cargo:


src/exercise.rs
```diff
@@ -102,6 +102,11 @@ path = "{}.rs""#,
                 // See https://github.com/rust-lang/rust-clippy/issues/2604
                 // This is already fixed on master branch. See this issue to track merging into Cargo:
                 // https://github.com/rust-lang/rust-clippy/issues/3837
+                Command::new("rustc")
+                    .args(&[self.path.to_str().unwrap(), "-o", &temp_file()])
+                    .args(RUSTC_COLOR_ARGS)
+                    .output()
+                    .expect("Failed to compile!");
```
jrvidal, line 109: Hm, the problem with this approach is that an exercise not compiling is sort of an expected state. A user could then edit an exercise and get `rustlings` to panic! (We do panic in two instances in this branch for things that would be way more unexpected)

So I propose the following: you could run try to compile the program to `temp_file()` and then:
* If it fails, you return that as the `cmd` of this branch, and
* If if succeeds, you proceed to run clippy and return _that_ as the `cmd`.

crodjer: Got it. Added a fixup to change behavior based on the compile status. If `compile_output.status.success()` is a success, we proceed with running clippy, otherwise it'll directly evaluate to compile output.

I tried modelling this with `.and_then`, but to avoid excessive nesting, I went ahead with inline extracting of the compile command result.

crodjer:
> ### Contributors
> 
> @crodjer

Had added a fixup commit, so it doesn't confirm to the conventions. I was planning to rebase and squash once the review is complete.

src/exercise.rs
jrvidal: :man_facepalming: I realize now that I misunderstood your first attempt and it was perfectly fine! :woman_facepalming: 

<details>

I mistook the `.expect()` call you added for something that would panic on a failed compilation. But in fact it would only panic if there was a problem with the `rustc` executable, which is something we already panic on :see_no_evil:

</details>

Sorry for the noise, I'll leave it up to you whether to revert to your initial solution or not.

crodjer: Yes @jrvidal  it would have continued through with a non-zero exit code. But, the benefit of the new approach was that we won't even execute clippy with compilation errors and respond with `rustc`'s output, which I'd like to think in a higher entity.

Although, with the older approach too, clippy fails with the same error as rustc would have, so does reflect the correct problem to the user.
<details>
<summary>Example compilation failure with the earlier approach.</summary>

```rust
    Checking clippy1 v0.0.1 (/home/rohan/projects/rustlings/exercises/clippy)
error[E0308]: mismatched types
  --> clippy1.rs:12:13
   |
12 |     if y != "foo" {
   |             ^^^^^ expected `f64`, found `&str`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `clippy1`
```
</details>

So, because it results in a simpler code flow, reverted to the old approach with a comment. Also, squashing into a single commit so that commit-lint is satisfied.

crodjer:
@jrvidal Could you please re-review the latest changes, as per the discussion? I also tried multiple modifications in the commit message, but couldn't get the commit lint to pass.

jrvidal:
@fmoko Not sure if commit-lint is working as intended or it's still a WIP

fmoko:
@jrvidal This PR was before I properly set it up, so I think it's a bit confused. Should be fine for merging

crodjer:
> @jrvidal This PR was before I properly set it up, so I think it's a bit confused. Should be fine for merging

@jrvidal, @fmoko 
Okay. But, I don't have an option to merge. It is restricted to collaborators I am guessing?

crodjer:
Thanks!

PR #301, opened 2020-04-11
chore: add result1.rs to exercise list

YKG:
[result1.rs](https://github.com/fmoko/rustlings/blob/master/exercises/error_handling/result1.rs) is missing in `info.toml`


fmoko:
`result1` is right [here](https://github.com/fmoko/rustlings/blob/master/info.toml#L552), it just got moved down a bit. But now that I look at it, `option1` is there twice for some reason.

PR #351, opened 2020-04-12
chore: update variables5.rs book link

saurav-2104:
chore: update variables5.rs book link

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -52,7 +52,8 @@ because we want to assign a different typed value to an existing variable. Somet
 you may also like to reuse existing variable names because you are just converting
 values to different types like in this exercise.
 Fortunately Rust has a powerful solution to this problem: 'Shadowing'!
-You can read more about 'Shadowing' in the book's section 'Variables and Mutability'.
+You can read more about 'Shadowing' in the book's section 'Variables and Mutability': 
+https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html#shadowing
 Try to solve this exercise afterwards using this technique."""
 
 # IF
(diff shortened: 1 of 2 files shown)


saurav-2104:
As a first time user, I was not able to figure out which book the exercise was talking about.

info.toml
```diff
@@ -52,7 +52,8 @@ because we want to assign a different typed value to an existing variable. Somet
 you may also like to reuse existing variable names because you are just converting
 values to different types like in this exercise.
 Fortunately Rust has a powerful solution to this problem: 'Shadowing'!
-You can read more about 'Shadowing' in the book's section 'Variables and Mutability'.
+You can read more about 'Shadowing' in the book's section 'Variables and Mutability': 
+https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html?highlight=variables#shadowing.
```
jrvidal, line 56: Hm, do we want the `highlight` query param?

saurav-2104: Oh, my bad. I think I copied my searched results from browser. Removed it.

info.toml
```diff
@@ -52,7 +52,8 @@ because we want to assign a different typed value to an existing variable. Somet
 you may also like to reuse existing variable names because you are just converting
 values to different types like in this exercise.
 Fortunately Rust has a powerful solution to this problem: 'Shadowing'!
-You can read more about 'Shadowing' in the book's section 'Variables and Mutability'.
+You can read more about 'Shadowing' in the book's section 'Variables and Mutability': 
+https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html#shadowing.
```
jrvidal, line 56: One last minor nit: in my terminal, the final dot messes with the link, so it ends up not jumping to the "Shadowing" section.

saurav-2104: Removed "."
Also, `cargo` updated the Cargo.lock as well. At this point, I am not sure if it is unwanted.

PR #356, opened 2020-04-17
fix confusing comment in conversions/try_from_into.rs

lebedevsergey:
There is a confusing comment in conversions/try_from_into.rs due to the copy/paste from from_into.rs task. Fixed it.


jrvidal:
Thanks!

fmoko:
@all-contributors add @lebedevsergey for docs

Issue #300, opened 2020-04-09
rustlings run resets Cargo.toml (clippy1)


jhscheer:
I solved `clippy1.rs` by using `float_cmp::approx_eq`, however I cannot finish this exercise because during `rustlings run clippy1` the `Cargo.toml` resets and deletes the new dependency I added.
```
[dependencies]
float-cmp = "0.6.0"
```

sjmann:
I think that you may have arrived at a correct, but unintended answer to this exercise. Have you checked the hint from clippy which appears against the code before you modify it?

I think this exercise could be probably be improved by adding an extra line to the header comment explaining what to do (maybe something like: `fix this code by following clippy's advice`)

jhscheer:
> I think that you may have arrived at a correct, but unintended answer to this exercise. Have you checked the hint from clippy which appears against the code before you modify it?
> 
Now I did, thanks.
> I think this exercise could be probably be improved by adding an extra line to the header comment explaining what to do (maybe something like: `fix this code by following clippy's advice`)
>
Probably yes.

PR #357, opened 2020-04-17
chore: add explanatory comment to clippy1

sjmann:
I noticed while investigating [#300 ](https://github.com/rust-lang/rustlings/issues/300) that it's a little jarring when transitioning to the `Mode::Clippy` exercises, as it's not immediately clear that the code fails to compile due to clippy warnings.

I have added comments to clippy1 to clarify the compilation failures are from clippy.


jrvidal:
Thanks!

Issue #358, opened 2020-04-19
possiable improved tests for exercise/conversions/from_into.rs


emiddleton:
I am working through the questions and I noticed that what seems to be the desired answer has some issues with various malformed string types.

The error cases I was thinking of were 
* missing age eg. "Mark,"
* missing comma and age eg. "Mark"
* missing name and age eg. ","
* missing name eg. . ",1"
* missing name and invalid age eg. ",one".

To handle these without panicking I ended up with
 
```rust
impl From<&str> for Person {
    fn from(s: &str) -> Person {
        if s.len() == 0 {
            Person::default()
        } else {
            let mut person_itr = s.split(',');
            match person_itr.next().map(|name| {
                (
                    if name.is_empty() {
                        Err("Missing Name")
                    } else {
                        Ok(name.to_string())
                    },
                    person_itr.collect::<String>().parse::<usize>(),
                )
            }) {
                Some((Ok(name), Ok(age))) => Person { name, age },
                _ => Person::default(),
            }
        }
    }
}
```

Without the above test cases I would have assumed the various elements were present in the string and thrown an exception if missing.

AbdouSeck:
Hi @emiddleton,

Thanks for bringing this up. I am assuming that the panicking is due to unclear instructions.
For the error cases you're thinking of, if we use the instructions, you should get `Person {name: "John", age: 30}`. The only exception may be `",1"`. With this one, you should also get `Person {name: "John", age: 30}`, but the instructions are not quite clear about what to do when only the name part of the string is empty.

**Next steps**:
1. Add the clarification that a name cannot be an empty string, so `Person::default()` should be returned.
2. Add more test cases using the error cases you've provided here.


If you'd like, please feel free to add those changes.

Please let me know what you think.

Thank you!

Abdou

AbdouSeck:
@emiddleton I believe this was addressed in #368. Please feel free to reach out if something looks amiss. Otherwise, @fmoko can go ahead and close this issue.

Thank you!

Abdou

PR #368, opened 2020-04-22
test: Add missing test case for from_str exercise

apatniv:
Added the missing test case for "from_str" exercise.

### DIFF
--- a/exercises/conversions/from_into.rs
+++ b/exercises/conversions/from_into.rs
@@ -29,7 +29,8 @@ impl Default for Person {
 // 1. If the length of the provided string is 0, then return the default of Person
 // 2. Split the given string on the commas present in it
 // 3. Extract the first element from the split operation and use it as the name
-// 4. Extract the other element from the split operation and parse it into a `usize` as the age
+// 4. If the name is empty, then return the default of Person
+// 5. Extract the other element from the split operation and parse it into a `usize` as the age
 // If while parsing the age, something goes wrong, then return the default of Person
 // Otherwise, then return an instantiated Person onject with the results
 impl From<&str> for Person {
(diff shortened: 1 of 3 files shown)


AbdouSeck:
Hi @apatniv,

Thanks for offering to improve the test cases for the `conversions` exercises. I believe the issue #358 brought this up. Could you please take a look at my response there and add the recommended changes to this pull request?

I believe incorporating those changes would go hand in hand with this pull request.

Please feel free to reach out if you have any additional questions.

Thank you,

Abdou

apatniv:
Abdou

Thanks for your response. I will take a look at #358 and add the necessary checks.

apatniv:
Would you recommend putting all malformed strings into one test case like this

```
    fn test_with_malformed_strings() {
        let bad_inputs = [
            "Mark",
            ",",
            ",one",
        ];
        
        for val in &bad_inputs {
            let result = Person::try_from(*val);
            assert!(result.is_err());
        }
        
    }
```

or create a new test function for each of the cases like 

```test_missing_comma_and_age```
```test_missing_name```

Former is easier to write but may be bad for reading/debugging purpose. Later, is easier to debug at the expense of verbosity as we need to add similar test cases in `try_from_into.rs`, `from_str.rs` and `from_into.rs`. 

IMPO, later approach may be easier for beginners.

AbdouSeck:
Hi @apatniv,

I am much more drawn to the second option for two reasons:

1. It aligns with the existing test functions and very easily tells the reader which exact combination of name and age is being tested.
2. If you use an array of strings, whenever one of them fails the test, the remaining are not tested. Furthermore, in a setup like this, the failure message may not be as clear.

I understand that it's going to be quite verbose, so feel free to do whatever sounds better to you and does not consume too much of your precious time.
Anyway, I am happy to keep the exchange going if you have any questions.

Thank you,

Abdou

apatniv:
@AbdouSeck Thanks for the feedback. I add all the possible test cases and tried to follow the conventions in individual test cases.

AbdouSeck:
Hi @apatniv, this looks fantastic! Thank you so much for the work!
@fmoko this should be ready.

Thank you all!

Abdou

fmoko:
Thanks!

fmoko:
@all-contributors add @apatniv for code and test

PR #373, opened 2020-04-28
docs: Update `Invoke-WebRequest` to `Start-BitsTransfer`

tijb:
To run `Invoke-WebRequest` you have to have the Internet Explorer Feature enabled, and you must have opened Internet Explorer at least once, which some people may not have (me for example).
If you don't you will receive an error and the install will fail. 
This also simplifies the download command (imo).
`Start-BitsTransfer` is a drop in replacement which doesn't need you to have internet explorer enabled.

_apologies if the commit naming is wrong, i wasn't sure what commit type this would fall under._

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Set-ExecutionPolicy RemoteSigned
 Then, you can run:
 
 ```ps
-Start-BitsTransfer https://git.io/rustlings-win | Select-Object -ExpandProperty Content | Out-File $env:TMP/install_rustlings.ps1; Unblock-File $env:TMP/install_rustlings.ps1; Invoke-Expression $env:TMP/install_rustlings.ps1
+Start-BitsTransfer -Source https://git.io/rustlings-win -Destination $env:TMP/install_rustlings.ps1; Unblock-File $env:TMP/install_rustlings.ps1; Invoke-Expression $env:TMP/install_rustlings.ps1
 ```
 
 To install Rustlings. Same as on MacOS/Linux, you will have access to the `rustlings` command after it.
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Set-ExecutionPolicy RemoteSigned
 Then, you can run:
 
 ```ps
-Invoke-WebRequest https://git.io/rustlings-win | Select-Object -ExpandProperty Content | Out-File $env:TMP/install_rustlings.ps1; Unblock-File $env:TMP/install_rustlings.ps1; Invoke-Expression $env:TMP/install_rustlings.ps1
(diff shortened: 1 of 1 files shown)


kennyek:
An alternative is to add the `UseBasicParsing` parameter in the first command to get it to work.

```sh
Invoke-WebRequest https://git.io/rustlings-win -UseBasicParsing | Select-Object -ExpandProperty Content | Out-File $env:TMP/install_rustlings.ps1; Unblock-File $env:TMP/install_rustlings.ps1; Invoke-Expression $env:TMP/install_rustlings.ps1
```

Solves the following error when running the command:

```sh
Invoke-WebRequest : The response content cannot be parsed because the Internet Explorer engine is not available, or
Internet Explorer's first-launch configuration is not complete. Specify the UseBasicParsing parameter and try again.
At line:1 char:1
+ Invoke-WebRequest https://git.io/rustlings-win | Select-Object -Expan ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotImplemented: (:) [Invoke-WebRequest], NotSupportedException
    + FullyQualifiedErrorId : WebCmdletIEDomNotSupportedException,Microsoft.PowerShell.Commands.InvokeWebRequestComman
   d
```

Brief explanation of the problem:
https://stackoverflow.com/a/38054505

tijb:
@kennyek didn't know that. 
Personally, I prefer the `Start-BitsTransfer` because of how it simplifies/clarifies the download command.

danwilhelm:
I looked into `Invoke-WebRequest`, `Start-BitsTransfer`, and a third method `System.Net.WebClient`. Based on its performance and broad compatibility, I believe `System.Net.WebClient` might be the best option. (A good resource for comparing them is: https://blog.jourdant.me/post/3-ways-to-download-files-with-powershell) 

- `Start-BitsTransfer` is apparently not always guaranteed to work, since some organizations may disable BITS.
- `Invoke-WebRequest` (with `UseBasicParsing`) supposedly does not work in some cases according to the post above. This method also seems to be significantly slower than the others by default (unless `$ProgressPreference = 'SilentlyContinue'` is also set).
- `System.Net.WebClient` was introduced in PowerShell 2, whereas `Invoke-WebRequest` requires PowerShell 3+. Similarly to `Invoke-WebRequest`, it seems to be universally compatible. It is the recommended method in the post above.

Here is the same command using `WebClient`:

`(New-Object System.Net.WebClient).DownloadFile("https://git.io/rustlings-win", "$env:TMP/install_rustlings.ps1"); Unblock-File $env:TMP/install_rustlings.ps1; Invoke-Expression $env:TMP/install_rustlings.ps1`

Also:
1. Perhaps the README should mention to run these commands in PowerShell, since they will not work via `cmd`.
2. In the documentation, Microsoft now recommends using `HttpClient` (.NET Framework 4.5+) instead of `WebClient` (.NET Framework 1.1+). For compatibility reasons, it seems `WebClient` is the better choice.

fmoko:
> Perhaps the README should mention to run these commands in PowerShell, since they will not work via cmd.

Good idea! Do you mind adding that?

Issue #387, opened 2020-05-06
Try harder to confine the user, and instruct them not to create new Errors.


EvanCarroll:
I learned a lot messing with `exercises/error_handling/errorsn.rs` but it wasn't an especially good experience. Without more information, you can go down a pretty horrible rabbit hole here. I think we should do a better job to

* confine the user into just modifying that function
* inform the user that they needn't create new errors.

I believe both of these are failures of the instructions and not hints. Find more information here,

https://stackoverflow.com/q/61626705/124486

PR #388, opened 2020-05-06
fix(errorsn): Try harder to confine the user.

EvanCarroll:
I went way off into the weeds on this. Trying harder to stop others.

#387


fmoko:
@all-contributors add @EvanCarroll for content

PR #389, opened 2020-05-09
fix(options1): Add hint about Array Initialization

jmahmood:
Array initialization / default variables is unrelated to Options and not overtly mentioned beforehand.  
As the question is about Options, it is easy for someone new to Rust to misunderstand and look elsewhere for a solution.


fmoko:
@all-contributors add @jmahmood for content

PR #393, opened 2020-05-12
feat: Rewrite try_from_into

### LINKED ISSUE #392: Solutions for `try_from_into` and `from_str` are same
I'm just copy-paste my code from fisrt exercise to second and it's ok.
May be should combine them into one?

IkaR49:
English is not my native language, so I may have made some mistakes in the text. I would be grateful if someone could check it out.

closes #392

### DIFF
--- a/exercises/conversions/from_str.rs
+++ b/exercises/conversions/from_str.rs
@@ -17,6 +17,7 @@ struct Person {
 // 3. Extract the first element from the split operation and use it as the name
 // 4. If the name is empty, then return an error
 // 5. Extract the other element from the split operation and parse it into a `usize` as the age
+//    with something like `"4".parse::<usize>()`.
 // If while parsing the age, something goes wrong, then return an error
 // Otherwise, then return a Result of a Person object
 impl FromStr for Person {
--- a/exercises/conversions/try_from_into.rs
+++ b/exercises/conversions/try_from_into.rs
@@ -5,98 +5,100 @@
 use std::convert::{TryInto, TryFrom};
 
 #[derive(Debug)]
-struct Person {
-    name: String,
-    age: usize,
+struct Color {
+    red: u8,
+    green: u8,
+    blue: u8,
 }
 
 // I AM NOT DONE
-// Your task is to complete this implementation
-// in order for the line `let p = Person::try_from("Mark,20")` to compile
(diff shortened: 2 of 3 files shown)


AbdouSeck:
Hi @IkaR49,

This looks really good. Thank you for the great work. I have one suggestion here. [The hint provided in the `info.toml` for the `from_str` exercise](https://github.com/rust-lang/rustlings/blob/d6c0a688e6a96f93ad60d540d4b326f342fc0d45/info.toml#L801) tells people to do an almost copy-paste operation from the `try_from_into` exercise to answer the question. However, that will not be the case once this pull request gets merged.
Can you please edit the hint for `from_str` to say something else? You can say something like:
```
The implementation of FromStr should return an Ok with a Person object, or an Err with a string if the string is not valid.
```

Please feel free to reach out if you have any questions.

Thank you so much!

Abdou

IkaR49:
Hello @AbdouSeck.

Are you suggesting replacing i16 with isize, if I understand correctly?

AbdouSeck:
@IkaR49,

Not really. I think it would be fine to just add a test cases to handle when values above 255 are passed. I am thinking of the following tests:

```
#[test]
fn test_out_of_range_slice() {
    let v = vec![10000, 1000, 1000];
    let c = Color::try_from(&v[..]);
    assert!(c.iserr());
}
```
and
```
#[test]
fn test_out_of_range_array() {
    let v = [10000, 1000, 1000];
    let c = Color::try_from(v);
    assert!(c.iserr());
}
```
and also
```
#[test]
fn test_out_of_range_tuple() {
    let v = (10000, 1000, 1000);
    let c = Color::try_from(v);
    assert!(c.iserr());
}
```

These will make sure that the implementations remember to return an error if the values provided are greater than 255.

Please let me know if you have any questions.

Thank you,

Abdou

IkaR49:
@AbdouSeck,

I fixed it. What you think about it?

ekaats:
I've just been going through the rustlings exercises and found this one a little confusing. I could not get the `type Error = Box<dyn error::Error>;`  working. After changing this back to `type Error = String;` and just returning an `Err("Something went wrong")`-esque error I was able to finish the exercise without issue. Maybe make a separate exercise for that error?

PR #394, opened 2020-05-13
fix(test2): name of type String and &str

GaurangTandon:
I know this is quite boring, but while we are putting backticks, it's good to be more accurate.


fmoko:
@all-contributors add @GaurangTandon for content

PR #400, opened 2020-05-16
fix(option2): Add TODO to comments

dev-cyprium:
I think a simple add of "TODO" will make this exercise look more in line with the other exercises.
Other exercises included a "TODO" before when the programmer was expected to enter a solution.


fmoko:
@all-contributors add @dev-cyprium for content

Issue #244, opened 2019-12-04
Transition from `tests3.rs` to `test3.rs` is confusing


cincodenada:
So I was going through rustlings (which has been great, thanks!) and as it happens, the next unit after `tests3.rs` is `test3.rs` in the root. It took me some hard staring to figure out why the code didn't look like I expected it to before I noticed the one-character difference in the filenames.

I'm not sure what the best solution here is - perhaps `test*.rs` in the root directory could be renamed to `quiz*.rs`, since that also describes what they are, and that doesn't collide with a concept being taught?

jrvidal:
I too have thought about this. Maybe `tests<X>.rs` could be renamed to `testing<X>.rs`?

fmoko:
Maybe something like `test-yourself<X>`?

ScarboroughCoral:
Is this problem solved?

ScarboroughCoral:
Oh, I am confused too.I think it's necessary to rename it.

sanjaykdragon:
> 
> 
> Maybe something like `test-yourself<X>`?

Yeah I think this is a better name

jrial:
My $0.02: renaming to "quiz" is the best suggestion I read in this thread so far. If the idea is to avoid confusion with the previous "test" file, it makes little sense to keep "test" in the name.

tonyito:
Does anyone know of the status of this issue? Still says test 1-4, when 'quiz' would be the perfect alternative as @jrial said.

fmoko:
I think it's a good idea, and if someone doesn't snipe it from me, I'll do it as soon as I can.

sanjaykdragon:
sniped @fmoko 
https://github.com/rust-lang/rustlings/pull/402

PR #402, opened 2020-05-16
changed test to quiz: fixes issues in #244

sanjaykdragon:
title, refer to 
https://github.com/rust-lang/rustlings/issues/244


sanjaykdragon:
> 
> 
> Wouldn't only renaming the files mess with rustlings watch that is looking for the actual file names to compile?

you are right, this should fix it

fmoko:
You would also need to change the reference to the filenames in https://github.com/rust-lang/rustlings/blob/master/README.md#L102.

sanjaykdragon:
> 
> 
> You would also need to change the reference to the filenames in [/README.md@`master`#L102](https://github.com/rust-lang/rustlings/blob/master/README.md#L102) [](/rust-lang/rustlings/blob/HEAD@{2020-05-17T10:54:54Z}/README.md#L102).

Didn't notice that, fixed.

> 
> 
> Could you change the title of each exercise? `// test1.rs` to `// quiz1.rs` and so forth?

Forgot about this, done and also changed up the wording from "test about X" to "quiz about X". let me know if I missed anything

PR #406, opened 2020-05-20
remove struct2 hint comment

jademcgough:
There was a hint added for struct2 by https://github.com/rust-lang/rustlings/pull/355, so this comment is misleading.


jrvidal:
Thanks!

Issue #403, opened 2020-05-17
Smart Pointer Exercises


vikulikov:
There is an exercise to use `Arc<T>` with threads. But I haven't found any exercises to practice in using `Rc<T>`, `RefCell<T>`, `Weak<T>` etc separately or together.

danwilhelm:
Agreed -- it would be nice to cover `Box` and `Rc` before the threadsafe `Arc`.

AlexandruGG:
I just wrote an exercise for `Box` here: https://github.com/rust-lang/rustlings/pull/409.

Let me know what you think 😁

fmoko:
Closing for now, feel free to let me know if you're interested in more smart pointer-related exercises.

Pacheco95:
@diannasoreil I would love it! I also created an issue asking for more smart pointers exercises https://github.com/rust-lang/rustlings/issues/1020

PR #409, opened 2020-05-26
feat: Add box1.rs exercise

AlexandruGG:
Having completed all of the existing exercises, I noticed there wasn't one with the `Box` type.

Although this exercise alone might not completely solve https://github.com/rust-lang/rustlings/issues/403, I think it's a start 🦀 .

### DIFF
--- a/exercises/standard_library_types/box1.rs
+++ b/exercises/standard_library_types/box1.rs
@@ -12,24 +12,42 @@
 // Step 1: use a `Box` in the enum definition to make the code compile
 // Step 2: create both empty and non-empty cons lists of by replacing `unimplemented!()`
 //
+// Note: the tests should not be changed
+//
 // Execute `rustlings hint box1` for hints :)
 
 // I AM NOT DONE
 
 #[derive(PartialEq, Debug)]
-enum List {
+pub enum List {
     Cons(i32, List),
     Nil,
 }
 
 fn main() {
-    let empty_list = unimplemented!();
-    println!("This is an empty cons list: {:?}", empty_list);
+    println!("This is an empty cons list: {:?}", create_empty_list());
+    println!("This is a non-empty cons list: {:?}", create_non_empty_list());
+}
 
-    let non_empty_list = unimplemented!();
-    println!("This is a non-empty cons list: {:?}", non_empty_list);
+pub fn create_empty_list() -> List {
+    unimplemented!()
+}
 
-    // Do not change these
-    assert_eq!(List::Nil, empty_list);
-    assert_ne!(empty_list, non_empty_list);
+pub fn create_non_empty_list() -> List {
+    unimplemented!()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_create_empty_list() {
+        assert_eq!(List::Nil, create_empty_list())
+    }
+
+    #[test]
+    fn test_create_non_empty_list() {
+        assert_ne!(create_empty_list(), create_non_empty_list())
+    }
 }
--- a/exercises/standard_library_types/box1.rs
+++ b/exercises/standard_library_types/box1.rs
@@ -1,0 +1,35 @@
+// box1.rs
+//
+// At compile time, Rust needs to know how much space a type takes up. This becomes problematic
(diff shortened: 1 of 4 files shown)


info.toml
```diff
@@ -614,6 +614,24 @@ hint = """
...
+hint = """
+Step 1
+The compiler's message should help: since we cannot store the value of the actual type
+when working with recursive types, we need to store a reference (pointer) to its value.
+We should therefore place our `List` inside a `Box`. More details in the book here:
+https://doc.rust-lang.org/book/ch15-01-box.html#enabling-recursive-types-with-boxes
+
+Step 2
+Creating an empty list should be fairly straightforward (hint: peek at the assertions).
+For a non-empty list keep in mind that wee want to use our Cons "list builder".
```
danwilhelm, line 630: wee -> we

exercises/standard_library_types/box1.rs
```diff
@@ -0,0 +1,35 @@
...
+fn main() {
+    let empty_list = unimplemented!();
+    println!("This is an empty cons list: {:?}", empty_list);
+
+    let non_empty_list = unimplemented!();
+    println!("This is a non-empty cons list: {:?}", non_empty_list);
+
+    // Do not change these
+    assert_eq!(List::Nil, empty_list);
+    assert_ne!(empty_list, non_empty_list);
```
danwilhelm, line 34: This is a clever way of requiring a non-Nil list variant without giving away the solution!

danwilhelm:
This is a good entry-level exercise with good accompanying explanations.

Based on prior Rustlings feedback, the main complaint might be that its solution is inside the Rust book. Personally, I believe this is fine since it gives people an opportunity to play with the ideas in the book.

+ I suggest packaging the asserts as tests. Notice in this screenshot that when the asserts fail, no error is displayed. This may be confusing to the user:
image

info.toml
AlexandruGG: good spot, macbook butterfly keyboard 😅 . I'll correct this and put the hints through Grammarly

AlexandruGG:
> This is a good entry-level exercise with good accompanying explanations.
> 
> Based on prior Rustlings feedback, the main complaint might be that its solution is inside the Rust book. Personally, I believe this is fine since it gives people an opportunity to play with the ideas in the book.
> 
> * Notice in this screenshot that when the asserts fail, no error is displayed with `rustlings watch`. This may be confusing to the user, so I suggest placing the asserts into tests.
>   image

Thank you for the feedback!

On the first point: I know the book readily provides the solution, but I thought this could serve as an easy intro to Boxes, and you still get the satisfaction of solving it by reading the book in my view :).

On the second point: I hadn't noticed that before because I tested with `rustlings run`, where the full panic message is displayed. However, I realised I can move my exercise to the top of `info.toml` to test with `rustlings watch` 😄 . So I've refactored it now with tests.

AlexandruGG:
unrelated question: would it be ok to add `.idea.` to the `.gitignore`? I think it would be useful for those using IntelliJ with Rust, as the IDE generates this folder

Screenshot 2020-05-27 at 10 17 14

danwilhelm:
I tried it out, and it works!

Perhaps someone more involved with the project could weigh in on whether it is okay for the solution to be in the book? In my opinion, adding this as-is is an improvement to what already exists. `Box` is often seen in Rust code, yet it was not previously covered by Rustlings! (A possible variant would be using a linked list rather than a cons list, e.g. at the bottom of https://rust-unofficial.github.io/too-many-lists/first-layout.html.)

PS - @fmoko gave you the thumbs up on .idea, so I'm assuming that's the official go-ahead :D

fmoko:
> PS - @fmoko gave you the thumbs up on .idea, so I'm assuming that's the official go-ahead :D

Yep, go ahead!

AlexandruGG:
> > PS - @fmoko gave you the thumbs up on .idea, so I'm assuming that's the official go-ahead :D
> 
> Yep, go ahead!

Thanks, added now!

What's the next step for this?

jrvidal:
Will look at this ASAP

AlexandruGG:
thanks everyone, I had a good time writing this and using **rustlings** in general. This was my first every contribution to an open-source project 😁

PR #410, opened 2020-05-27
fix(option1): Don't add only zeros to the numbers array

tsauvajon:
If you decide to print the elements in `numbers`, it will display only zeros, which can be confusing.

This change makes the numbers a little bit bigger.

```
printing: 0
printing: 0
printing: 0
printing: 0
printing: 0

vs

printing: 0
printing: 19
printing: 38
printing: 57
printing: 77
```


sanjaykdragon:
Looks good

PR #411, opened 2020-05-28
chore: Update variables6.rs book link

allansduarte:
Just was updated the book link to add an anchor to have direct access to the "Differences Between Variables and Constants" section that is related to the hint exercise.


jrvidal:
Thanks!

PR #422, opened 2020-06-04
Show tests prints

### LINKED ISSUE #262: View a specific exercise's stdout
Hi there,

I'm working through rustlings on a Windows machine. It's a wonderful resource - thank-you so much for putting the effort into making it.

One thing I have yet to figure out is how to view the output of the current exercise. Currently I just see if an exercise compiles (and the compile errors if not). However, if an exercise uses `println!` it would be lovely to be able to see that output somewhere.

Am I doing something wrong or is this something that's yet to be implemented?

AbdouSeck:
This new feature can be accessed by invoking rustlings with `--nocapture`.

closes #262

**_MINOR CHANGES_**:
    * Documentation added to source code under [`src`](https://github.com/rust-lang/rustlings/tree/master/src)
    * Both unit and integration tests for `--nocapture` have been added.

**_BREAKING CHANGES_**:
The following function take a new boolean argument:
    * `run`
    * `verify`
    * `test`
    * `compile_and_test`

`compile_and_test` is the function that uses this new boolean argument to determine whether or not to display the output from the test harnesses.

Please feel free to reach out with any questions.

Thank you,

Abdou

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -124,16 +124,16 @@ fn prompt_for_completion(exercise: &Exercise, prompt_output: Option<String>) -> 
         Mode::Clippy => "The code is compiling, and 📎 Clippy 📎 is happy!",
     };
 
-    println!("");
+    println!();
     println!("🎉 🎉  {} 🎉 🎉", success_msg);
-    println!("");
+    println!();
 
     if let Some(output) = prompt_output {
         println!("Output:");
         println!("{}", separator());
         println!("{}", output);
         println!("{}", separator());
-        println!("");
+        println!();
     }
 
     println!("You can keep working on this exercise,");
@@ -141,12 +141,12 @@ fn prompt_for_completion(exercise: &Exercise, prompt_output: Option<String>) -> 
         "or jump into the next one by removing the {} comment:",
         style("`I AM NOT DONE`").bold()
     );
-    println!("");
+    println!();
     for context_line in context {
         let formatted_line = if context_line.important {
             format!("{}", style(context_line.line).bold())
         } else {
-            format!("{}", context_line.line)
+            context_line.line.to_string()
         };
 
         println!(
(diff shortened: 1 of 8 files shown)


jrvidal:
ack, I'll look into this ASAP

src/verify.rs
```diff
@@ -73,7 +88,10 @@ fn compile_and_test(exercise: &Exercise, run_mode: RunMode) -> Result<bool, ()>
     progress_bar.finish_and_clear();
 
     match result {
-        Ok(_) => {
+        Ok(output) => {
+            if verbose {
+                println!("{}", output.stdout);
```
jrvidal, line 93: I'm starting to think that we should be a bit more sophisticated while handling the output of exercises, and try to have both stdout and stderr merged as they would appear in the terminal screen.

Anyway, out of scope, just thinking out loud.

AbdouSeck: @jrvidal thanks for the review! I agree with providing more information to the end users as long as there is a way to control that; like with command line options to make output `--quiet` or `--verbose`.

Normally, an exercise's standard error should be empty unless there is a complication error or a runtime one. Unless someone explicitly uses `eprint!` or `eprintln!` (or the many other ways of writing to targeted file handles) in their code to solve an exercise, there should be nothing in the standard error. So, keeping standard output and error separate may actually come in handy when trying to figure out if a user actually wrote to the standard error.

I am open to discussing this further, though.

Thanks!

Abdou

AbdouSeck:
I decided to a commit which hopefully closes #423. I just added a couple of steps to check for string emptiness and use `master` as the backup for installing the program when the latest tag can't be obtained.

Thanks,

Abdou

AbdouSeck:
@fmoko please let me know if this needs additional work to get it merged.

Thanks,

Abdou

fmoko:
@AbdouSeck All good! Just went under my radar.

src/verify.rs
mo8it: > I'm starting to think that we should be a bit more sophisticated while handling the output of exercises, and try to have both stdout and stderr merged as they would appear in the terminal screen.
> 
> Anyway, out of scope, just thinking out loud.

Done in the `v6` branch with `os_pipe` :D

PR #429, opened 2020-06-08
fix(arc1): Passively introduce attributes

alexxroche:
Ensure that std::sync::Arc is actually used, as this exercise can be compiled using things already learnt in previous exercises.


AbdouSeck:
@fmoko this seems like a really good idea to me. Unless someone else has any reservation to it, I think it should be good to merge.

Thanks!

Abdou

Issue #432, opened 2020-06-11
exercices/structs/structs3.rs Use of float for a cost in a course, really ?


SebastienCaunes:
Line 28 there is this horrible error: 
`
    fn get_fees(&self, cost_per_kg: f32) -> f32 {
`

**Floats must not be used for a price, never, ever. And even more in a lesson.**

This is much more clever, and just correct:
`
    fn get_fees(&self, cent_per_kg: i32) -> i32 {
`

Sorry but I see this error too often to stay silent, this point has to be understood and not taken lightly: **Floats are not decimal numbers.** Please do not contribute in growing a population of rust coders that don't understand floats.

Some might think, "it's not the purpose of the lesson, that's too advanced to be explained", so why not using integers and cents then? Why should we show a student one of the worst thing he could ever do in production?

fmoko:
Thanks for reporting this! I would advise you to soften your language a bit in the future, as some maintainers may not take too kindly to the very pejorative tone you're putting up here. That being said, I agree that this is a bad practice. Would you be up for creating a PR to fix this?

danwilhelm:
@SebastienCaunes If you are able to make a PR, note there is a similar problem with `assert_eq!` - it is not ideal to perform exact comparisons on floating point values. Also it might be nice to change `from` -> `sender_city` and `to` -> `recipient_city`.

SebastienCaunes:
@fmoko : houps ! sorry for the tone, I'm not native english and also I had a lot of hard fights on this subject. I even got fired for "incompatibility with the lead developer" trying to explain him that it's a bad idea to use a float as a primary key in database to handle more than 2 bilions records.

Good Idea to do a PR I'll try to do it cleanly for getting forgiven for my bad tone. :)

@danwilhelm : yes I saw that too, and I think I will also convert the weight to grams so we only have i32.

orlovedev:
I can't get why I need to multiply the price per **kilo** by the weight in **grams** to make the test pass. Is it something wrong with me or the naming needs to be amended since the previous update?

Also, in the hint, the method is referred to as "calculate_transport_fees" but in the impl it is called "get_fees".

SebastienCaunes:
> I can't get why I need to multiply the price per **kilo** by the weight in **grams** to make the test pass. Is it something wrong with me or the naming needs to be amended since the previous update?
> 
> Also, in the hint, the method is referred to as "calculate_transport_fees" but in the impl it is called "get_fees".

[...]

PR #433, opened 2020-06-11
fix : Use of integer for prices, therefore also for weight

SebastienCaunes:
rename confusing "from" and "to" to sender_country and recipient_country as suggested

### DIFF
--- a/exercises/structs/structs3.rs
+++ b/exercises/structs/structs3.rs
@@ -7,17 +7,17 @@
 
 #[derive(Debug)]
 struct Package {
-    from: String,
-    to: String,
-    weight: f32
+    sender_country: String,
+    recipient_country: String,
+    weight_in_grams: i32,
 }
 
 impl Package {
-    fn new(from: String, to: String, weight: f32) -> Package {
-        if weight <= 0.0 {
+    fn new(sender_country: String, recipient_country: String, weight_in_grams: i32) -> Package {
+        if weight_in_grams <= 0 {
             // Something goes here...
         } else {
-            return Package {from, to, weight};
+            return Package {sender_country, recipient_country, weight_in_grams};
         }
     }
 
@@ -25,8 +25,8 @@ impl Package {
         // Something goes here...
     }
 
-    fn get_fees(&self, cost_per_kg: f32) -> ??? {
-        // Something goes here...
+    fn get_fees(&self, cents_per_kg: i32) -> ??? {
+        // Something goes here... (beware of grams to kg conversion)
     }
 }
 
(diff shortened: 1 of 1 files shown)


SebastienCaunes:
Issue 432 https://github.com/rust-lang/rustlings/issues/432

exercises/structs/structs3.rs
```diff
@@ -7,26 +7,26 @@
...
         }
     }
 
     fn is_international(&self) -> ??? {
         // Something goes here...
     }
 
-    fn get_fees(&self, cost_per_kg: f32) -> ??? {
-        // Something goes here...
+    fn get_fees(&self, cents_per_kg: i32) -> ??? {
```
danwilhelm, line 28: Perhaps this should also be `cents_per_gram`. Otherwise, the output may have to be a float, which we're trying to avoid. For example, a package weighing 150 grams -- with a cost of 10 cents per kg -- would have a cost of 1.5 cents.

SebastienCaunes: 1.5 cent will be rounded and that's what we want because we can't pay 0.5 cents anyway.
Cents per gram is not precise enough as a price is usually less than one cent per gram.
Yes it require to divide by 1000. I think that is not hard to grasp. I also think it would be nice to add a link to a good explanation why float are bad for money.

danwilhelm: Currently, it asks the user to supply the return type. Many users likely will reason that it should be a float, since otherwise information is lost (and as-is it is unclear whether they should round or truncate). 

So, to avoid another issue, it might be good to either:
1. Force the return type to be an integer, and add a comment asking the user to truncate the result. -or-
2. Use `cents_per_gram`, where we will not have to deal with fractions.

SebastienCaunes: There is no need to truncate or round, integers operations do it by themselves.
Relevant information is lost with floats, not with integer.
I think your option 1 is the good one, no need to ask for any truncate or rounding. It is done naturally by integers operation.
Option 2 makes no sense as even the smallest integer price for 1 gram is out of scale ($1000 for 1kg).

danwilhelm: Thinking through this as a student, I would notice that `1254 / 1000 == 1`. I would be hesitant to lose the extra 254 grams via an integer division.

Since Rustlings did not specify how to handle excess grams, I would likely decide the safest approach is to pro-rate `cents_per_kg`, maybe like this:

> let weight_in_kgs = (weight_in_grams as f32) / 1000.;
> return weight_in_kgs * (cents_per_kg as f32) as i32; 

Yet even now, I am unsure whether I wrote the function as Rustlings intended -- should I have rounded or truncated the result? (Money is often rounded.)

So, this is why I recommend including a comment explaining exactly how the fees should be computed. I am just trying to avoid another Issue later on 😀

SebastienCaunes: You complicate your life too much ! 😀 You must multiply first then divide.
`
fn get_fees(&self, cents_per_kg: i32) -> i32 {
    self.weight_in_grams * cents_per_kg / 1000
}
`

No need for cast or anything.

SebastienCaunes: We could also simply store weight in kg and integer, ignoring grams for the example it makes everything more simple and is more correct than using floats for money.

danwilhelm: 😀 Yes, that is a much better method and yields the same result! So perhaps a specification would be something like:

``` rust
// Fees are pro-rated. Disregard any final fractions of a cent.
// Try to compute the fees without using floats!
```

SebastienCaunes: As you want as long there's no float used for money, I'm happy.

I just did the clippy1 exercice with the beautifull advice :
```
  if y != x {
   |        ^^^^^^ help: consider comparing them within some error: `(y - x).abs() > error`
```
That's how floats must be handled !

danwilhelm: > We could also simply store weight in kg and integer, ignoring grams for the example it makes everything more simple and is more correct than using floats for money. Should I do another pull request this way ?

It might be easiest just to make them all the same units. Other than this, I think it looks great!

jrvidal: I'm a bit unsure about the units puzzle: it risks losing focus, but a bit of non-Rust-related challenge makes the exercises more amenable. No strong opinion anyway.

exercises/structs/structs3.rs
```diff
@@ -37,31 +37,31 @@ mod tests {
...
 
     #[test]
     fn calculate_transport_fees() {
-        let country_from = String::from("Spain");
-        let country_to = String::from("Spain");
+        let sender_country = String::from("Spain");
+        let recipient_country = String::from("Spain");
 
-        let country_fee = ???;
+        let cents_per_kg = ???;
```
jrvidal, line 61: Random thought, totally out of scope for this PR: it feels weird to me that we have learners guess backwards the appropriate input to obtain a predetermined output.

Also, I'm starting to think that forcing miscompilations through invalid code like `???` is something we should dial back a bit.

PR #439, opened 2020-06-15
Update links

ShalokShalom:
All links previously directing to second edition of the book, got changed to direct to the current version.


jrvidal:
Thanks!

PR #448, opened 2020-06-21
feat: add fake manifest for analyzers/IDEs (fixes #443)

### LINKED ISSUE #443: no error squiggles in VS Code with rls-vscode extension
VS Code: 1.45.1
OS: 4.19.118-1-MANJARO
rls-vscode: 0.7.8

Creating a new crate/workspace like shown [here](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) the error squiggles work fine. But opening the root directory of rustlings (directory where Cargo.toml is located):
- no error squiggles
- no problems in the PROBLEMS window, despite of code errors

Is this a problem of the rustlings project directory structure? Opening src/main.rs and editing it, the error squiggles appear. Sorry if this topic isn't related to rustlings itself, it's my first time in the Rust world.

jrvidal:
This enables analyzers to work correctly as if the `exercises` folder were a true Cargo project.

* It is an "orphan" Cargo.toml, completely unrelated to the root `Cargo.toml`. This has pros and cons.
   * It could be potentially confusing to learners, since they have to open the `exercises` folder for the analyzer to work.
   * OTOH it does not "infect" the main project build with tons of targets, many of them in a non-compilable state.
* The folder structure is unchanged! :tada:


exercises/Cargo.toml
```diff
@@ -0,0 +1,286 @@
+# This Cargo.toml is AUTOGENERATED, you shouldn't need to edit it.
```
jrvidal, line 1: We could generate this on the fly, but it does seem a bit more error-prone

src/exercise.rs
```diff
@@ -167,9 +167,10 @@ path = "{}.rs""#,
     fn run(&self) -> Result<ExerciseOutput, ExerciseOutput> {
         let arg = match self.mode {
             Mode::Test => "--show-output",
-            _ => ""
+            _ => "",
```
jrvidal, line 170: Some `cargo fmt` noise here and there.

tooling/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
jrvidal, line 1: The maintenance commands are hidden behind a lib and a feature so it does not interfere with the normal build. It is also exposed as an alias in `.cargo/config` to help prospective contributors.

exercises/Cargo.toml
```diff
@@ -0,0 +1,286 @@
...
+# is to help editors analyze your code.
+# To regenerate it, run `cargo generate-manifest`.
+[package]
+name = 'exercises'
+version = '0.1.0'
+edition = '2018'
+authors = ['The Rustlings Maintainers']
+publish = false
+
+[[bin]]
```
jrvidal, line 12: All targets are configured as `[[bin]]`. I'm not sure if this could be problematic for exercises without `main` (maybe some analyzers will complain that is missing?).

jsejcksn:
I tried this:

- Copied [`./exercies/Cargo.toml`](https://github.com/rust-lang/rustlings/blob/90c148df088e18a3039fee5ea33ad4b3f204df72/exercises/Cargo.toml) to my own clone of the rustlings repo (checked out to `4.0.0`)
- Opened `./examples` in VS Code
- Opened an exercise file

The Rust extension began working for **quite a while**, processing all the binaries. Eventually, it resulted in in-editor intellisense linting for each of the exercise files.

However, all of the problems from all files appear in the Problems panel at once. This might be all that is possible (I'm not sure how the Rust extension handles problem reporting). In comparison, the TypeScript language server only reports problems for files which are currently open in editor tabs.

Is there a way to get problem reporting without actually running all the debug incremental builds? It would speed things up tremendously.

---

Aside: Interestingly, this was among the reported problems:

> `~/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/src/libcore/macros/mod.rs`
> mod.rs[70, 17]: expected struct `std::vec::Vec`, found `()`
```
mismatched types

expected struct `std::vec::Vec`, found `()`

note: expected struct `std::vec::Vec<std::string::String>`
      found unit type `()`rustc(E0308)
```

exercises/Cargo.toml
jsejcksn: Errors are produced for each target without a `main` fn.

jrvidal:
@jsejcksn Thanks for your testing. Out of curiosity, what analyzer were you using with VSCode? I don't get any errors about missing `main()` functions.

In fact, a plain old `cargo check` does _not_ show errors for all files, which makes me think that this manifest will not work as intended :disappointed: It doesn't even seem deterministic :upside_down_face:

jsejcksn:
@jrvidal Whatever the Rust extension uses by default

jrvidal:
I've been looking into this a bit more. The problem with an all-encompassing manifest is that `cargo check` chokes with some many targets, and bails out as soon as it finds 3 failing exercises or so.

I'll update the review soon trying a different approach: making the manifest point to the "current" exercise, per `rustlings watch`.

jsejcksn:
> The problem with an all-encompassing manifest is that `cargo check` chokes with so many targets, and bails out as soon as it finds 3 failing exercises or so.

@jrvidal The strategy I shared in https://github.com/rust-lang/rustlings/issues/443#issuecomment-657150623 works well to address that issue.

jrmoulton:
Is there a reason that this was never merged?

Zerotask:
@jrvidal I like your idea. Do you plan to update your PR?

jrvidal:
It is an open question whether we want IDE integration or not: figuring out things "manually" might be a core part of the experience we want to provide. I wouldn't mind updating the PR, but I'd like to hear @fmoko's thoughts first.

guyellis:
> It is an open question whether we want IDE integration or not: figuring out things "manually" might be a core part of the experience we want to provide.

@jrvidal I think that the option to enable or disable IDE integration should up to the user. Each of us learns in a different way. For some they will have a better/faster learning experience not having IDE integration. Others will do better with it. I think that the best solution would be to provide the learner with the option to enable/disable and some docs on why they would want one or the other.

PR #456, opened 2020-06-27
Introduce a Pointers section

TheLe0:
This is a very important topic to have, because great programmers need to understand pointers, because they help a lot
to understand how low level code works.

### DIFF
--- a/exercises/pointers/README.md
+++ b/exercises/pointers/README.md
@@ -1,0 +1,10 @@
+### Pointers
+
+Rust is a low level low-level programming language and focus on high level code performance, for this
+is necessary a good memory management and this infers on the use of pointers.
+Pointers is a primitive type, that allows to access the memory address of a variable, or the value 
+alocated on an specific address.
+
+#### Book Sections
+
+- [Pointers](https://doc.rust-lang.org/std/primitive.pointer.html)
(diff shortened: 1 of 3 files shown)


fmoko:
Thanks for your pull request! Unfortunately, I have to say I'm opposed to the inclusion of pointers in Rustlings, as it's not a very common use case to directly interface with them, and you can accomplish 99% of all things in Rust without ever touching raw pointers (in fact, this is one of the strong suits of Rust's memory management model - being able to have granular memory management without having to use pointers directly). The only thing that might be useful to know in relation to pointers is dereferencing, but I don't see it worth adding that in scope of this PR.

fmoko:
(Also, I wouldn't agree that "great programmers need to know how pointers work", you don't need to know all low level concepts to be a great programmer. If you're into low level systems programming, sure, it's important to know how pointers work, but I still believe that many people would find this topic's inclusion into Rustlings more confusing than helpful)

jrvidal:
(Side note: "pointer" in Rust usually refers to "raw pointer", while this exercise is about references. Though their semantics are quite similar, the current terminology is potentially confusing).

TheLe0:
I presume this was a rejection, so i'm going to close this.
Thanks for your time!
cheers!

PR #464, opened 2020-07-09
Update README.md

Nuc1eoN:
It should either be `keeps it` or `keep them`, but `keep it` is grammatically wrong :)


fmoko:
Thanks!

PR #474, opened 2020-07-21
feat(cli): Added 'cls' command to 'watch' mode

DeepSpace2:
feat(cli): Added 'cls' command to 'watch' mode


fmoko:
Personally, I'd prefer `clear` instead of `cls`. Any reason why you've called it that?

DeepSpace2:
@fmoko Not really, we can change it to what the consensus will be. May be even both :)

fmoko:
@DeepSpace2 I think just 'clear' would be fine :) Thanks!

DeepSpace2:
@fmoko fixed, changed to 'clear' :)

PR #483, opened 2020-07-30
feat : Read File via File and BufReader.

UsairimIsani:


AbdouSeck:
@UsairimIsani is it possible to combine this PR with #521? They seem to tackle IO operations, with your PR doing the reading while @calvinbrown085 is doing some writing operations.

Thanks,
Abdou

PR #488, opened 2020-08-04
chore(installation): Refactor installation script

kruschk:
- Run "install.sh" through `shellcheck` and implement the suggested changes, bar one regarding the use of `ls`.
- Ensure the `git clone ...` operation is successful before continuing.
- Add more descriptive comments.
- Refactor and simplify code.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,184 +1,133 @@
 #!/usr/bin/env bash
-set -euo pipefail
 
-echo "Let's get you set up with Rustlings!"
-
-echo "Checking requirements..."
-if [ -x "$(command -v git)" ]
-then
-    echo "SUCCESS: Git is installed"
-else
-    echo "ERROR: Git does not seem to be installed."
-    echo "Please download Git using your package manager or over https://git-scm.com/!"
-    exit 1
-fi
-
-if [ -x "$(command -v cc)" ]
-then
-    echo "SUCCESS: cc is installed"
-else
-    echo "ERROR: cc does not seem to be installed."
-    echo "Please download (g)cc using your package manager."
-    echo "OSX: xcode-select --install"
-    echo "Deb: sudo apt install gcc"
-    echo "Yum: sudo yum -y install gcc"
-    exit 1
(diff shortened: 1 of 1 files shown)


diannasoreil:
@kruschk Any interest in updating this to fit the current state? It's ok if not :)

kruschk:
@diannasoreil I'd be happy to! :slightly_smiling_face: It looks like the installation script hasn't changed much since I submitted my PR: there's a new dependency check for GCC and the minimum rust version has been bumped to 1.39. It should be fairly straightforward for me to update.

Just out of curiosity: is the GCC dependency necessary? Isn't `rustc` sufficient?

kruschk:
@diannasoreil Is it okay to use a generic error message, that includes the name of the offending command, when a dependency check fails? The current approach, with custom messages for the various dependencies, makes it awkward to factor out that logic.

install.sh
```diff
@@ -1,184 +1,133 @@
...
-    echo "Deb: sudo apt install gcc"
-    echo "Yum: sudo yum -y install gcc"
-    exit 1
-fi
+# Compare two version strings, v1 and v2, passed as arguments (e.g 1.31 and
+# 1.33.0). Returns 1 if v1 > v2, 0 if v1 == v2, 2 if v1 < v2.
+compare_versions() {
+    [ "$1" == "$2" ] && return 0
+    declare -a v1=() v2=()
+    mapfile -d . -t v1 < <(echo "$1")
```
diannasoreil, line 8: Hm, this fails on my M1 Pro Mac on the default Apple bash shell because it seems `mapfile` is not provided.

kruschk: Ahh, sorry, I was tidying up while preparing to install a new distro and forgot that you would get a notification when I pushed to the branch. I didn't verify that the script still worked; I just pushed what I had. I wouldn't have caught your particular issue, though, so thanks for bringing it up!

If you like, I can revise this properly once more so that we can (hopefully) get it merged!

PR #498, opened 2020-08-17
fix(arc1): index mod should equal thread count

seeplusplus:
fixes a small math error where some indices will get double counted

### DIFF
--- a/exercises/standard_library_types/arc1.rs
+++ b/exercises/standard_library_types/arc1.rs
@@ -21,7 +21,7 @@ fn main() {
             let mut sum = 0;
             while i < child_numbers.len() {
                 sum += child_numbers[i];
-                i += 5;
+                i += 8;
             }
             println!("Sum of offset {} is {}", offset, sum);
         }));


liby:
>  where some indices will get double counted

I don't quite understand here. Could you tell me more about it? If there’s an obvious answer to a question I’m asking, I’m sorry I didn’t get it Have a nice day.

seeplusplus:
> 
> 
> > where some indices will get double counted
> 
> I don't quite understand here. Could you tell me more about it? If there’s an obvious answer to a question I’m asking, I’m sorry I didn’t get it Have a nice day.

So the for `for loop` on L18 is going to spawn 8 different threads. The `i = offset` bit and the `i += 8` bit basically say "for the thread I'm on, start at my 'offset' and skip every other number." Effectively this is intended to partition the number space. However, if the number of threads you spawn is DIFFERENT than the amount you skip each iteration (`i += 8`), you will hit some numbers twice. For instance, if we had 3 threads and instead the line read `i += 2` (where correctly it should read `i += 3`) then the threads would count the numbers like so:

+ thread 0: 0, 2, 4, 6, ...
+ thread 1: 1, 3, 5, 7, ...
+ thread 2: 2, 4, 6, 8, ...

So in this example there's overlap between thread 2 and 0. In this PR the thread count was 8 and the incorrect modulo was 5 so we had,

+ thread 0: 0, 5, 10, ...
+ thread 1: 1, 6, 11, ...
+ thread 2: 2, 7, 12, ...
...
+ thread 5: 5, 10, 15, ...
+ thread 6: 6, 11, 16, ...
+ thread, 7: 7, 12, 17, ...

liby:
@seeplusplus Thank you for giving me such a detailed explanation, I fully understand, thank you so much!

PR #502, opened 2020-08-24
Make comments on example clearer

CaribouJohn:
I actually struggled because I overlooked the fact that I needed to change the 2.1 to "A+". 

So I wanted to make things clearer for future rustlings.


fmoko (approved):
Looks good to me! Thanks!

PR #504, opened 2020-08-25
fix(exercises): adding question mark to quiz2

Ant0wan:
Adding question marks for consistency.
Would it be easier to understand quiz2 ?

Closes#496


fmoko (approved):
I agree, it would be more easily understandable. Thoughts @jrvidal @AbdouSeck?

jrvidal:
:+1:

PR #509, opened 2020-08-27
fix(structs3): Small adjustment of variable name

JSquar:
A small follow-up to PR #432, where the price of kg was changed to price per gram to avoid usage of floats. The variable identifier should therefore also be changed, because `cents_per_kg` would indicate 3/1000 and not 3.

### DIFF
--- a/exercises/structs/structs3.rs
+++ b/exercises/structs/structs3.rs
@@ -29,7 +29,7 @@ impl Package {
         // Something goes here...
     }
 
-    fn get_fees(&self, cents_per_kg: i32) -> ??? {
+    fn get_fees(&self, cents_per_gram: i32) -> ??? {
         // Something goes here... (beware of grams to kg conversion)
     }
 }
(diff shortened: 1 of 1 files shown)


fvsch:
The comment should probably be changed too, then?
The "(beware of grams to kg conversion)" was added in https://github.com/rust-lang/rustlings/pull/433 and doesn't make sense anymore if the exercise expects grams all around.

With the current exercise (on 4.0.0) my solution looks like:

```rust
    fn get_fees(&self, cents_per_kg: i32) -> i32 {
        // Something goes here... (beware of grams to kg conversion)
        let cents_per_gram = cents_per_kg as f32 / 1000_f32;
        let fees = cents_per_gram * self.weight_in_grams as f32;
        fees.floor() as i32
    }
```

(And `let cents_per_kg = 3000;` to pass the test. Note that I'm a total beginner so the conversion code might not be great.)

With this PR, I guess it would be greatly simplified to:

```rust
    fn get_fees(&self, cents_per_gram: i32) -> i32 {
        // Something goes here...
        cents_per_gram * self.weight_in_grams
    }
```

With `let cents_per_gram = 3;` as input.

exercises/structs/structs3.rs
```diff
@@ -29,7 +29,7 @@ impl Package {
         // Something goes here...
     }
 
-    fn get_fees(&self, cents_per_kg: i32) -> ??? {
+    fn get_fees(&self, cents_per_gram: i32) -> ??? {
         // Something goes here... (beware of grams to kg conversion)
```
fmoko, line 33: This needs to be changed!

fmoko (approved):

exercises/structs/structs3.rs
JSquar: You mean the comment? Yes, I changed it

PR #512, opened 2020-08-31
fix(using_as): Add test so that proper type is returned.

ryanpcmcquen:
It was pretty easy (at least for me), to get this to compile but return an incorrect value. I've added a test to ensure the proper type is used.

Using a `usize` cast:
Screen Shot 2020-08-31 at 12 54 02 PM

Using `f64`:
Screen Shot 2020-08-31 at 12 53 41 PM

Seems like the test is helpful to avoid that.

### DIFF
--- a/exercises/conversions/using_as.rs
+++ b/exercises/conversions/using_as.rs
@@ -3,6 +3,7 @@
 // It also helps with renaming imports.
 //
 // The goal is to make sure that the division does not fail to compile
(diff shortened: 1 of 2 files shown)


AbdouSeck:
Hi @ryanpcmcquen thanks for this PR. This is very much needed for the exercise!

@fmoko, should be good to go.

Best,
Abdou

PR #516, opened 2020-09-10
features: init, preposition

Horki:


Horki:
#517 let me know if this is a good idea,

TODO:
* update exercise.rs with Mode::Feature
* update `run.rs` with compilation for feature mode, and looping all features under [info.toml](https://github.com/Horki/rustlings/blob/cfg-features/info.toml#L429)

diannasoreil:
I think this is a bit too specific to Cargo to be in Rustlings, we want to focus more on Rust's syntax, and less on Cargo-specific attributes/features.

Horki:
@diannasoreil thanks for reply 👍

PR #519, opened 2020-09-14
fix: measurement mismatch in structs3.rs

orlovedev:
Since the previous change where floats were replaced with integers, the `get_fees` function started to feel weird. The tests imply that we multiply **cents_per_kg** by **weight_in_grams** which is not logical.

This change just renames variables. Now, there is the price in cents per **gram** and the weight in **grams**, thus not misleading those who do the exercise.

Discussed in #432

### DIFF
--- a/exercises/structs/structs3.rs
+++ b/exercises/structs/structs3.rs
@@ -29,8 +29,8 @@ impl Package {
         // Something goes here...
     }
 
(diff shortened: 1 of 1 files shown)


orlovedev:
Closing as a duplicate to #509

PR #520, opened 2020-09-21
Exercise order: string before struct

ephetic:
The struct exercises rely on knowledge learned in the string exercises, so should follow rather than precede them.

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -52,7 +52,7 @@ because we want to assign a different typed value to an existing variable. Somet
 you may also like to reuse existing variable names because you are just converting
 values to different types like in this exercise.
 Fortunately Rust has a powerful solution to this problem: 'Shadowing'!
-You can read more about 'Shadowing' in the book's section 'Variables and Mutability': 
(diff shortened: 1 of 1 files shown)


ephetic:
Just discovered `result1.rs` comes long after `errorsn.rs` which clearly extends it.  This PR maybe needs to be closed in favor of a broader review of exercise order.

ZFLloyd:
I am currently in the process of going through exercises, and I had the same reflection. Following the compiler I instinctively ended up using `&str` for the c and tuple struct, which then led me to introducing lifetimes. And using borrowed time for a struct like this doesn't really seems like a good pattern to me.
After some research I've eventually used `String` before even getting to the strings exercises that would have helped a lot (including the quizz 2).

AbdouSeck:
Hi all, I am closing this PR in favor of #541. `Struct` before `String` is a bit hard to wrap one's head around. But remember that the `String` type in `rust` is a fat pointer to some buffer of data. You can use `String` typed data without having to know anything about how they behave. So, it would make sense to use them in structs before really understanding how they work.

Issue #482, opened 2020-07-29
Stdio, Read, Write and Files


UsairimIsani:
Add Exercises on Stdio, Read, Write and Files.

calvinbrown085:
I can try and take files if that would be okay!

seeplusplus:
I'd like to help close this.  What kinds of exercises would we like to see?  With @calvinbrown085 's work we have reading from a file (files1). I could see adding something like: 

+ Files: files2, writing/amending
+ Stdio: stdio1, capturing user input from stdin and writing it out to stdout

I'd be happy to commit to creating these exercises (as well as any others you folks suggest).

calvinbrown085:
IMO, writing a new file would be a good exercise, I also agree that your stdio ones would be good too!

seeplusplus:
Great! I've started to work on this! Hopefully, this weekend I'll get some solid time to just write up a quick and dirty first draft of these challenges. 

The branch I'm working under is:
https://github.com/seeplusplus/rustlings/tree/add-stdio-exercises

The gist of it is I'll be creating exercises for:

1. reading a file
2. reading from stdin
3. writing a file from stdin

In that order. If anyone has any recommendations for other exercises that would be nice to have, or any ways to implement these changes, I'd be glad to take input.

samusz:
May be also write to stderr ?

mo8it:
I think that we can add a whole IO section with at least 2 exercises :)

Open to pull requests :D

frroossst:
I can try and work on a new io section if this is still wanted

mo8it:
@frroossst Go ahead!

I think that we can have 3 small exercises about dealing with files:
1. Opening and reading a file. Maybe just count the number of lines? The solution should show both manually opening and reading the file and using [`fs::read_to_string`](https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html).
2. Creating a file and writing to it. The file should be truncated if it already exists. The solution should show both manually opening and writing the file and using [`fs::write`](https://doc.rust-lang.org/stable/std/fs/fn.write.html).
3. Appending to an existing (log) file. It should be created if it doesn't already exist. No truncation.

In all of these exercises, the user should complete the body of a function which takes a path as argument. Input files should be created in tests and the path is then passed to the user function.

[...]

PR #521, opened 2020-09-21
feat: Adding files exercise

calvinbrown085:
#482 

This adds an example for reading in a file from the filesystem! Please let me know if I can do anything else here!

### DIFF
--- a/exercises/files/README.md
+++ b/exercises/files/README.md
@@ -1,0 +1,8 @@
+### Files
+The rust standard library supports reading files and writing them directly from the language!
+
+`std::fs::File` is just `A reference to an open file on the filesystem.`
+
+#### Book Sections
+
+- [Reading a file](https://doc.rust-lang.org/stable/book/ch12-02-reading-a-file.html)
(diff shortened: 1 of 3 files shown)


AbdouSeck:
Hi @calvinbrown085, thanks for proposing this feature. I am trying to figure out what the learner is supposed to do here. Are you trying to have the learner write the unit test case? If so, then the exercise won't work as written. The learner **_does not_** write and update unit tests. They update functions on top of the file, and your unit tests try to check the correctness of said functions.

For instance, in the file [`file1.rs`](https://github.com/rust-lang/rustlings/blob/8d9fe0580a4fdddeae6bea3ed99486f477d5abf7/exercises/files/files1.rs), what probably want to have is an incomplete function definition at the top. So, this means leaving `fn create_file() -> std::io::Result<()> {}` incomplete. And then, you want to write unit tests that `assert!` that the file has been created.

Please feel free to reach out if you wish to discuss this further.

Thanks,
Abdou

AbdouSeck (changes requested):
Please my comments

Issue #434, opened 2020-06-11
tests3.rs does not check for both tests before continuing


peterkos:
It appears that `tests3.rs` will succeed even if the only defined test is `is_true_when_even`, despite the file comment saying to add a second one to test (presumably) `is_false_when_not_even`:
```
// ... Then write a second test that tests whether we get the result
// we expect to get when we call `is_even(5)`.
```

I feel like it should test for both cases to match the instructions. 
Debug info:
```
❯ rustc --version
rustc 1.44.0 (49cae5576 2020-06-01)
❯ rustlings --version
rustlings 3.0.0
```

sautoirs:
I think `is_false_when_even()` should be renamed `is_false_when_odd()` or `is_false_when_not_even()` as proposed by @peterkos.

sautoirs:
My comment is fixed by #553. Thx

PR #522, opened 2020-09-21
Fixing test3 to have enough tests to make sure we test all cases

calvinbrown085:
#434

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -65,6 +65,7 @@ fn compile_and_run_interactively(exercise: &Exercise) -> Result<bool, ()> {
         Err(output) => {
             warn!("Ran {} with errors", exercise);
             println!("{}", output.stdout);
+            println!("{}", output.stderr);
             return Err(());
         }
     };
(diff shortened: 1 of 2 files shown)


src/verify.rs
```diff
@@ -65,6 +65,7 @@ fn compile_and_run_interactively(exercise: &Exercise) -> Result<bool, ()> {
         Err(output) => {
             warn!("Ran {} with errors", exercise);
             println!("{}", output.stdout);
+            println!("{}", output.stderr);
```
jrvidal, line 68: I think this is a spurious change, isn't it?

jrvidal (approved):
LGTM :rocket:

src/verify.rs
fmoko: yeah, i think this should go in a separate PR if anything

jrvidal: I think it's part of https://github.com/rust-lang/rustlings/pull/522/commits/9f61db5dbe38538cf06571fcdd5f806e7901e83a, which is already on `main`.

calvinbrown085: @jrvidal @fmoko That's my bad, I think I based my change off of master, instead of 4.0.0 which had another commit in it.

PR #537, opened 2020-09-24
Remove type when passing in a struct to an enum

yudhiesh:
I am fairly new to Rust and I was doing this exercise and noticed that the default test had an issue where it was defining the type being passed to the enum as a struct. Test was failing and the compiler advised to remove the type of Point. When it was removed test ran successfully.


BryanQuigley:
+1.  I got stuck on this too. 

#560 also addresses this so this seems pretty common.

xehpuk:
The test shows how to implement the enum types.

This should work:

```rust
enum Message {
    Move(Point),
    // ...
}
```

PR #545, opened 2020-10-01
fix(variables5) : make shadowing solution more prominent

### LINKED ISSUE #375: variables5: shadowing solution should be more prominent
For people coming from different programming languages that does not allow shadowing in the same scope, the natural solution for them is to change the assignments to the same type in order to compile.

YJDoc2:
fix(variables5) : make shadowing solution more prominent
closes #375


fmoko (approved):
good idea, thanks!

PR #560, opened 2020-10-13
Update enums3.rs

fpolica91:
This will cause the function to keep failing, In order to pass I needed to remove a pair of quotes, not sure if this was intentional and my solution was wrong, but this was the only solution I was able to come up with.

### DIFF
--- a/exercises/enums/enums3.rs
+++ b/exercises/enums/enums3.rs
@@ -51,7 +51,7 @@ mod tests {
             position: Point { x: 0, y: 0 },
             color: (0, 0, 0),
         };
-        state.process(Message::ChangeColor((255, 0, 255)));
+        state.process(Message::ChangeColor(255, 0, 255));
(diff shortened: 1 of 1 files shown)


darnuria:
@fpolica91 hi! Thanks for your submission!

In these exercise I think the point is to write out an `enum Message` with changing as little as possible where the code is already written.

Did you consider using a `(u8, u8, u8)` for `Message::Color`?

Right now in your file I wonders you wrote in `Message` `Color(u8, u8, u8)`. It's not the same as `Color((u8, u8, u8))`,
These change might need to change your implementation in `fn process`.

But I also agree that it's not really clear and it's subtle in a beginner journey to see the difference to from the two types. So I think there is matter to discuss! :D

darnuria:
@fmoko wanted to say #560 I think. :D

fmoko:
you're right haha

diannasoreil:
Closing along with #537.

PR #566, opened 2020-10-21
Remove a redundant character from the range

wh5a:


fmoko:
Technically this is correct, as `x..y` ranges exclude `y` and `x..=y` ranges include `y `.

Issue #549, opened 2020-10-04
Pretty much exact same exercise: Conversions - from_into and from_str


NicolasResier:
Dear,

Thank you very much for all these wonderful exercises. They were loads of fun and very interesting.
I would like to report however that there are 2 exercises at the end that are pretty much the same:
- Conversions > from_into.rs
- Conversions > from_str.rs

The only difference is that in case of a problem, in one you send a default while in the other you send an error.

Hope this helps.
Greetings,
Nicolas

mariusfrinken:
To add to that comment:

Why does the first test for an empty input string expects an error, while all other tests with bad input simply expect the function to panic?

If you want to teach people that returning a well formulated **Error** is the way to go in such circumstances and not just hacking something together with `panic!(),` please make that unambiguously clear.

AbdouSeck:
Hi @NicolasResier, thanks for raising this concern. I wrote those two exercises with the goal of highlighting the similarities between the `FromStr` and `From` traits. While these have somewhat different API's when used, their implementation are very similar. Again, the goal here is to make it obvious that `From` and `FromStr` look _**very very**_ similar.

@mariusfrinken, I am not sure what you're getting at. But the reason behind having `#[should_panic]` test cases is to handle situations where a user `unwraps` without actually checking that the returned value is `ok`. Also, in `from_str.rs`, your goal is not to write unit tests; the tests are there to check that you've implemented the code above it properly.

With that said, if you have any suggestions for what the exercises should be, please share them either via this issue or via a pull request. I am more than happy to discuss this further. And, sorry about the delayed response.

Thanks,
Abdou

mariusfrinken:
@AbdouSeck 
To clarify:
The naive approach I took to this exercise:
0. copy everything from the other similar exercise, make some slight changes
1. Oh I should throw an error, but what kind of error?
2. Let's look into the test cases to see the expected behaviour
3. Oh you expect that thing to panic
4. Let's try calling `panic!()` instead of messing with the Error notation 
(honestly my personal problem here was, that I just looked at Chapter 9. Error Handling, and there I found no detailed explanation on how to create your own Errors etc)
5. It worked for all test cases except the first one
6. Confusion
7. ...
8. I realize that it is indeed bad practice to use `panic!()` here, but I feel like the other testcases should reflect that, too.

My recommendation:
[...]

PR #571, opened 2020-10-26
feat(try_from_into): Add tests

fiplox:
Hello everyone :smile:. I was trying to resolve the `try_from_into` exercise and found out that I passed all the tests if I sum up for example all elements in array and test if it is <= 765 or >= 0. But if we sum up `(255 + 255 + (-1))` it gives us a good result but the third value is negative, so we should not pass it.

I added 3 tests for each type to prevent users pass this type of errors.

Example for array:
```rust
// Array implementation
impl TryFrom<[i16; 3]> for Color {
    type Error = String;
    fn try_from(arr: [i16; 3]) -> Result<Self, Self::Error> {
        if arr[0] + arr[1] + arr[2] <= 765 && arr[0] + arr[1] + arr[2] >= 0 {
            let c = [arr[0] as u8, arr[1] as u8, arr[2] as u8];
            Ok(Color {
                red: c[0],
                green: c[1],
                blue: c[2],
            })
        } else {
            Err("NOP".to_string())
        }
    }
}
```
This code passes all the tests, but if we add test with [-1, 255, 255] it does not pass :+1:.
I'm open for discussion if it is not merge-able to correct ! :smiley:

### DIFF
--- a/exercises/conversions/try_from_into.rs
+++ b/exercises/conversions/try_from_into.rs
@@ -25,22 +25,19 @@ struct Color {
 // Tuple implementation
 impl TryFrom<(i16, i16, i16)> for Color {
     type Error = String;
-    fn try_from(tuple: (i16, i16, i16)) -> Result<Self, Self::Error> {
-    }
+    fn try_from(tuple: (i16, i16, i16)) -> Result<Self, Self::Error> {}
 }
 
 // Array implementation
 impl TryFrom<[i16; 3]> for Color {
     type Error = String;
-    fn try_from(arr: [i16; 3]) -> Result<Self, Self::Error> {
-    }
+    fn try_from(arr: [i16; 3]) -> Result<Self, Self::Error> {}
 }
 
 // Slice implementation
 impl TryFrom<&[i16]> for Color {
     type Error = String;
-    fn try_from(slice: &[i16]) -> Result<Self, Self::Error> {
-    }
+    fn try_from(slice: &[i16]) -> Result<Self, Self::Error> {}
 }
 
 fn main() {
@@ -76,6 +73,11 @@ mod tests {
         let _ = Color::try_from((-1, -10, -256)).unwrap();
     }
     #[test]
+    #[should_panic]
+    fn test_tuple_sum() {
+        let _ = Color::try_from((-1, 255, 255)).unwrap();
+    }
+    #[test]
     fn test_tuple_correct() {
         let c: Color = (183, 65, 14).try_into().unwrap();
         assert_eq!(c.red, 183);
@@ -92,6 +94,11 @@ mod tests {
     #[should_panic]
     fn test_array_out_of_range_negative() {
         let _: Color = [-10, -256, -1].try_into().unwrap();
+    }
+    #[test]
+    #[should_panic]
+    fn test_array_sum() {
+        let _: Color = [-1, 255, 255].try_into().unwrap();
     }
     #[test]
     fn test_array_correct() {
@@ -111,6 +118,12 @@ mod tests {
     #[should_panic]
     fn test_slice_out_of_range_negative() {
         let arr = [-256, -1, -10];
+        let _ = Color::try_from(&arr[..]).unwrap();
+    }
+    #[test]
+    #[should_panic]
+    fn test_slice_sum() {
+        let arr = [-1, 255, 255];
         let _ = Color::try_from(&arr[..]).unwrap();
     }
     #[test]
--- a/exercises/conversions/try_from_into.rs
+++ b/exercises/conversions/try_from_into.rs
@@ -4,7 +4,7 @@
 // You can read more about it at https://doc.rust-lang.org/std/convert/trait.TryFrom.html
 use std::convert::{TryFrom, TryInto};
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Color {
     red: u8,
     green: u8,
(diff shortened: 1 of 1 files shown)


exercises/conversions/try_from_into.rs
```diff
@@ -76,6 +73,11 @@ mod tests {
         let _ = Color::try_from((-1, -10, -256)).unwrap();
     }
     #[test]
+    #[should_panic]
+    fn test_tuple_sum() {
+        let _ = Color::try_from((-1, 255, 255)).unwrap();
```
darnuria, line 78: Disclamer: I am not maintainer of Rustlings just a modest suggestion:

Would have done something 
`assert!(Color::try_from((-1, 255, 255)).is_err())`
[playground link](like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=use%20std%3A%3Aconvert%3A%3ATryFrom%3B%0A%0Astruct%20Color%3B%0Aimpl%20TryFrom%3C(i16%2C%20i16%2C%20i16)%3E%20for%20Color%20%7B%0Atype%20Error%20%3D%20String%3B%0Afn%20try_from(tuple%3A%20(i16%2C%20i16%2C%20i16))%20-%3E%20Result%3CSelf%2C%20Self%3A%3AError%3E%20%7B%0A%20%20%20%20Err(String%3A%3Afrom(%22No%22))%0A%7D%0A%0A%7D%0Afn%20main()%20%7B%0A%20%20%20%20assert!(Color%3A%3Atry_from((-1%2C%20255%2C%20255)).is_err())%0A%7D)

 I try to avoid unwrap in test because I hate the stack-trace it generate would prefers a `expect` or if the error is precise even an `assert_eq!` or a `assert!(expr.is_ok)` or `assert!(expr.is_err())` if it's hard to match value.

darnuria (changes requested):
Otherwise I love the idea, because it help to avoid sneaking arround the intent of the exercice. When I "cheese" an exercise I always personally feel bad. ;P

exercises/conversions/try_from_into.rs
fiplox: Then we would need to rewrite every test with `is_err()` because every test uses `unwrap()`

AbdouSeck: Hi y'all, this is a great addition. The desire to move from `.unwrap()` has been brought up in #549. If you're up for it, please feel free to rewrite the tests by substituting `.unwrap()` with `.is_err()` and removing the `#should_panic` directives.

AbdouSeck:
Just going to link this PR with issue #549. If the unit tests can rewritten to replace `.unwrap()` with `.is_err()`, then this PR could close the linked issue.

Thanks,
Abdou

AbdouSeck:
@fiplox It should be fine to do that, as long as the `#[should_panic]` directive is removed from the target tests.

Thanks,
Abdou

fiplox:
@AbdouSeck Would it be ok replace this test :
```rust 
#[test]
    fn test_slice_correct() {
        let v = vec![183, 65, 14];
        let c = Color::try_from(&v[..]).unwrap();
        assert_eq!(c.red, 183);
        assert_eq!(c.green, 65);
        assert_eq!(c.blue, 14);
    }
```
with this :
```rust
    #[test]
    fn test_slice_correct() {
        let v = vec![183, 65, 14];
        let c = match Color::try_from(&v[..]) {
            Ok(color) => color,
            Err(_) => Color {
                red: 0,
                green: 0,
                blue: 0,
            },
        };
        assert_eq!(c.red, 183);
        assert_eq!(c.green, 65);
        assert_eq!(c.blue, 14);
    }
```
and other correct tests too the same way?

AbdouSeck:
@fiplox That would be fine with me.

exercises/conversions/try_from_into.rs
```diff
@@ -66,71 +63,98 @@ mod tests {
...
     fn test_array_correct() {
-        let c: Color = [183, 65, 14].try_into().unwrap();
+        let c: Color = match [183, 65, 14].try_into() {
+            Ok(color) => color,
+            Err(_) => Color {
+                red: 0,
+                green: 0,
+                blue: 0,
+            },
+        };
```
darnuria, line 115: I know bikesheding is tiresome but would definitely prefers:

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=%23%5Bderive(Debug%2C%20PartialEq)%5D%0Astruct%20Color%20%7B%0A%20%20%20%20red%3A%20u8%2C%0A%20%20%20%20green%3A%20u8%2C%0A%20%20%20%20blue%3A%20u8%2C%0A%7D%0A%0A%23%5Bcfg(test)%5D%0Amod%20tests%20%7B%0A%20%20%20%20use%20super%3A%3A*%3B%0A%20%20%20%20%23%5Btest%5D%0A%20%20%20%20fn%20test_coulour()%20%7B%0A%20%20%20%20%20%20%20%20let%20c%20%3A%20std%3A%3Aresult%3A%3AResult%3CColor%2C%20String%3E%20%3D%20Ok(Color%20%7Bred%3A%20182%2C%20green%3A%2065%2C%20blue%3A%2014%7D)%3B%0A%20%20%20%20%20%20%20%20assert_eq!(c%2C%20Ok(Color%20%7B%20red%20%3A%20183%2C%20green%3A%2065%2C%20blue%3A%2014%20%7D))%3B%0A%0A%20%20%20%20%2F%2F%20or%0A%20%20%20%20%7D%0A%7D%0A)
```rust
let c : std::result::Result<Color, String> = Ok(Color {red: 182, green: 65, blue: 14});
assert_eq!(c, Ok(Color { red : 183, green: 65, blue: 14 }));
```
Failing message will be better than zeroish value.
Rustlings is used as a basis for new rustacean, it's important to try to stick with nice and idiomatic way of using rust :D

exercises/conversions/try_from_into.rs
```diff
@@ -66,71 +63,98 @@ mod tests {
...
         let v = vec![183, 65, 14];
-        let c = Color::try_from(&v[..]).unwrap();
+        let c = match Color::try_from(&v[..]) {
+            Ok(color) => color,
+            Err(_) => Color {
+                red: 0,
+                green: 0,
+                blue: 0,
+            },
+        };
```
darnuria, line 145: Same as upper checking against with `assert_eq!` an expected value wrapped in a `std::Result` will be more idiomatic :heart: .

exercises/conversions/try_from_into.rs
```diff
@@ -66,71 +63,98 @@ mod tests {
...
     fn test_tuple_correct() {
-        let c: Color = (183, 65, 14).try_into().unwrap();
+        let c: Color = match (183, 65, 14).try_into() {
+            Ok(color) => color,
+            Err(_) => Color {
+                red: 0,
+                green: 0,
+                blue: 0,
+            },
+        };
```
darnuria, line 86: Same as below. :)

exercises/conversions/try_from_into.rs
```diff
@@ -66,71 +63,98 @@ mod tests {
     use super::*;
 
     #[test]
-    #[should_panic]
     fn test_tuple_out_of_range_positive() {
-        let _ = Color::try_from((256, 1000, 10000)).unwrap();
+        assert!(Color::try_from((256, 1000, 10000)).is_err());
```
darnuria, line 67: great! ^_^

darnuria (changes requested):
Sorrryyyy i know i look like the bikesheding strange guy who comes up after the discussion settled.

But I think it's better to rethink the way of checking error, beginners look in rustlings tests to find out a solution or to learn idiomatic testing. It's like correction of homework, giving the best of ourself help beginners to tackle Rust difficulties. :D Just IMHO.

exercises/conversions/try_from_into.rs
fiplox: Agreed. Changed in the latest commit :heart:

AbdouSeck (approved):

AbdouSeck (approved):
I am not a huge of the additional `PartialEq` derive, but I hope it does not add any confusion to beginners.

fiplox:
It should not add any confusion as it is one of the last exercises. :wink:

fiplox:
Did force push to leave only useful information.

fmoko (approved):

PR #573, opened 2020-10-28
fix(structs1): Adjust wording

13r0ck:
The hint for structs1 improperly stated that there are 2 types of structs. Updated to clarify there are 3.

I changed the link to the page related to the exercise directly, though the one previous link is not necessarily wrong as it is one page before the link I prefer.


13r0ck:
If I am wrong then https://doc.rust-lang.org/rust-by-example/custom_types/structs.html needs to be updated, as It states there are 3.

TianyiShi2001 (approved):

fmoko (commented):
Looks good, a couple of small nits

PR #581, opened 2020-11-05
feat: Add "rustlings list" feature

### LINKED ISSUE #576: Ability to list the names of all available exercises for searching/debugging purposes
Currently, there doesn’t seem to any option to get the list of all exercises names. In my personal opinion, it would to useful to provide this feature for searching and debugging purposes.   This could help in finding out all exercises for a particular topic and their exact names. Also, users can debug if they end up accidentally running a wrong exercise name.

```bash
 > rustlings run if
No exercise found for your given name!
```

One way to solve the problem is to add a new subcommand say “list”, that will list all exercise names. This will help users to do the following things: 

- Find out all possible exercises

- Filter based on topics 
```bash 
rustlings list | grep “variable” 
# or
 rustings list | rg “if” 
```
 
Alternative, is that user should know about **`info.toml`** structure and look into name property under exercises. 

Let me if you think this is valuable addition and I will more than glad to make a pull request.

Sorry if this has already been asked, but I couldn't find it.

seeplusplus:
closes #576

I wasn't sure if I wanted to open a PR for this, since I wasn't sure if this was a feature we'd want to add to the project. Opening it up to gain visibility and get more feedback.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,6 +54,11 @@ fn main() {
                 .about("Returns a hint for the current exercise")
                 .arg(Arg::with_name("name").required(true).index(1)),
         )
+        .subcommand(
+            SubCommand::with_name("list")
+                .alias("l")
+                .about("Lists the exercises available in rustlings")
+        )
         .get_matches();
 
     if matches.subcommand_name().is_none() {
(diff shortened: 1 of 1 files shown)


fmoko (approved):
This looks good to me!

jrvidal (approved):
LGTM :rocket:

PR #587, opened 2020-11-12
fix(try_from_into): type error

wh5a:


AbdouSeck (approved):
This was overlooked. Thank you for catching it.

AbdouSeck:
@fmoko this should be good to go. Thanks!

Issue #405, opened 2020-05-19
rustlings hint primitive_types4 suggest wrong documentations.


wontheone1:
```
$ rustlings hint primitive_types4

...
If you're curious why the first argument of `assert_eq!` does not
have an ampersand for a reference since the second argument is a
reference, take a look at the Deref coercions section of the book:
https://doc.rust-lang.org/book/ch15-02-deref.html
```

The coercion from arrays to slices doesn't happen via dereference. 
It happens via `unsizing`, a completely different mechanism.
The coercion [T; N] -> [T] happens through the same means as Type -> dyn Trait.
Refer https://doc.rust-lang.org/std/marker/trait.Unsize.html

ayushmishra2005:
@wontheone1 I would like to address this. I agree with your suggestion. Could you please suggest me the text, which you want here?

wontheone1:
@ayushmishra2005 , I don't really have a particular suggestion, sorry

diannasoreil:
Hopefully this rewrite (https://github.com/rust-lang/rustlings/commit/6020ec1fe21aa1763821940dcc7029b99fb9d00a#diff-6bb9edccc7e8b95865f9969548084abe509295653d1423a4c0bb7b01fccb2bf3R217) makes it make a bit more sense?

PR #588, opened 2020-11-15
Fix help docs for primitive types 4

calvinbrown085:
Fixes https://github.com/rust-lang/rustlings/issues/405

Please let me know if there are other comments wanted here.


jrvidal:
I would personally prefer if we drop the tip altogether. `primitive_types4` is the 10th exercise or so... the learner is probably not at a stage where we can point them to something as involved as unsizing.

calvinbrown085:
Sounds good, I can drop the tip!

Issue #287, opened 2020-03-22
Threads1.rs hint references "sample answers"; where are they? 


andrew-scott-fischer:
The Threads1.rs hint references "sample answers".  I looked through the project and have not found them.  Are sample answers something that existed in a previous version and no longer exist?

woss:
i think they meant from the book. i solved it when i read the chapter on Threads in Rust by example. i'll give you links, i know you can figure it out. :) 

https://stevedonovan.github.io/rust-gentle-intro/7-shared-and-networking.html#shared-state
https://doc.rust-lang.org/book/ch16-03-shared-state.html#atomic-reference-counting-with-arct


> As you might suspect, Mutex<T> is a smart pointer. More accurately, the call to lock returns a smart pointer called MutexGuard, wrapped in a LockResult that we handled with the call to unwrap. The MutexGuard smart pointer implements Deref to point at our inner data; the smart pointer also has a Drop implementation that releases the lock automatically when a MutexGuard goes out of scope, which happens at the end of the inner scope in Listing 16-12. As a result, we don’t risk forgetting to release the lock and blocking the mutex from being used by other threads because the lock release happens automatically.

sapir:
I also thought it meant full exercise solutions, like #293. Maybe change it to say "sample answers from the book"?

alexxroche:
threads1.rs is the first exercise where I got frustrated.
<details> <summary>SPOILERS</summary>

```rust
use std::sync::{Arc, Mutex};
use std::thread;
use std::time::Duration;

struct JobStatus {
    jobs_completed: u32,
}

fn main() {
    let status = Arc::new(Mutex::new(JobStatus { jobs_completed: 0 }));
    thread::spawn(move || {
        for _ in 0..10 {
            thread::sleep(Duration::from_millis(250));
            status.clone().lock().unwrap().jobs_completed += 1;
        }   
    });
  /* 
    // I haven't been able to get this part to work
    while status.lock().unwrap().jobs_completed < 10 {
    // so I've resorted to this hack in frustration
  */
  let mut counter = 0;
  loop {
      if counter >= 6 {break;}; counter += 1;
      println!("waiting... {} ", counter);
            thread::sleep(Duration::from_millis(500));
    }

}
```
</details>
[...]

PR #591, opened 2020-11-28
The magic number 6 is given a bit more context

pcn:
I also ran into difficulty with this exercise. While looking for a hint, I saw this thread issue comment: https://github.com/rust-lang/rustlings/issues/287#issuecomment-644067533

Since `6` is just the appropriate number and has nothing to do with the solution, and since when we're frustrated we have a tendency to grasp at straws, I think it would help a tiny bit to eliminate that possible source of uncertainty.


fmoko (approved):

newblaque (approved):

jrvidal:
@all-contributors please add @pcn for maintenance

PR #592, opened 2020-11-29
fix(vec1): Have test compare every element in a and v

tinkhauser:
The previous test would stop comparing elements in array a and vec v upon reaching the last element of either. This resulted in the test passing even if v did not contain all the elements in a. This change to the test fixes that bug and should only pass if all the elements in a and v are present and equal.

### DIFF
--- a/exercises/collections/vec1.rs
+++ b/exercises/collections/vec1.rs
@@ -20,6 +20,6 @@ mod tests {
     #[test]
     fn test_array_and_vec_similarity() {
         let (a, v) = array_and_vec();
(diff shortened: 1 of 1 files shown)


mfrlin:
Just got to this problem. My solution was `let v = vec![];` and the test passed.

This PR fixes it.

AbdouSeck (approved):
Using `cmp::Eq` rather than `.zip` + `.all` is more correct. @fmoko this looks good.

@tinkhauser thanks for the PR!

Abdou

PR #541, opened 2020-09-27
feat: match exercise order to book chapters

JuliaCao:
Exercises are not organized in order for ```rustlings watch``` and it was hard to follow the exercises because we always had to jump back and forth between chapters.

1. Updated exercise order to match that of book chapters
2. Added exercise to book chapter mapping table to exercise README.md

PR #598, opened 2020-12-12
fix: added missing exercises to info.toml

JuliaCao:
Fix to this PR #541


AbdouSeck (approved):
@JuliaCao thanks so much for making the changes. We're now back to 74 exercises and the collections set and `iterator1` are now included.
I appreciate the quick turnaround. @fmoko this should be good now.

Best,
Abdou

faith (approved):

fmoko (approved):

PR #599, opened 2020-12-13
feat(cli): Improve the list command with options, and then some

### LINKED ISSUE #281: Follow `rustlings watch` with editor by opening source files in the same order

### LINKED ISSUE #465: Unexpected completion
POC：
1. Finish exercises in order while `rustlings watch`.
2. When the current working exercise is `move_semantics2.rs`, make a copy of it.
3. The new file `move_semantics2 copy.rs` is generated, then rustlings outputs `★ All exercises completed! ★` and exits.

I have not verified the file name in the above process is critical or not.
In fact, I am not sure that is an issue ... Maybe I should not do that?

Env:
- rustc 1.44.1 (c7087fe00 2020-06-17)
- cargo 1.44.1 (88ba85757 2020-06-11)
- rustlings 4.0.0

### LINKED ISSUE #585: `as_ref_mut` not ran
Hey,
First of all, thank you very much for this instructive course, which helped me to get back into rust very fast.
I noticed that, when I finished the second-to-last exercise, `from_str`, that `rustlings watch` immediately told me that I was finished, although I did not touch `as_ref_mut`. Do you know why that might be?
I am running rustlings 4.1.0 and rustc 1.47.0
Best regards
Anton

AbdouSeck:
The PR is two-fold:
1. `rustlings list` should now display more than just the exercise names. Information such as file paths and exercises statuses should be displayed. The `--paths` option limits the displayed fields to only the path names; while the `--names` option limits the displayed fields to only exercise names. You can also control which exercises are displayed, by using the `--filter` option, or the `--solved` or `--unsolved` flags.

**Sample invocations**:
```sh
# Display exercises that deal with traits, or enums, or structs.
rustlings list --filter "trait,enum,struct"

# Sample output:
# Name             	Path                                          	Status
# structs1         	exercises/structs/structs1.rs                 	Pending
# structs2         	exercises/structs/structs2.rs                 	Pending
# structs3         	exercises/structs/structs3.rs                 	Pending
# enums1           	exercises/enums/enums1.rs                     	Pending
# enums2           	exercises/enums/enums2.rs                     	Pending
# enums3           	exercises/enums/enums3.rs                     	Pending
# traits1          	exercises/traits/traits1.rs                   	Pending
# traits2          	exercises/traits/traits2.rs                   	Pending

# Display only exercises pending to be solved
rustlings list --unsolved

# Sample output:
# Name             	Path                                          	Status
# structs2         	exercises/structs/structs2.rs                 	Pending
# structs3         	exercises/structs/structs3.rs                 	Pending
# enums1           	exercises/enums/enums1.rs                     	Pending
# enums2           	exercises/enums/enums2.rs                     	Pending
# enums3           	exercises/enums/enums3.rs                     	Pending
# modules1         	exercises/modules/modules1.rs                 	Pending
# modules2         	exercises/modules/modules2.rs                 	Pending
# collections1     	exercises/collections/vec1.rs                 	Pending
# collections2     	exercises/collections/vec2.rs                 	Pending
# collections3     	exercises/collections/hashmap1.rs             	Pending
# collections4     	exercises/collections/hashmap2.rs             	Pending
```
This portion should also close #281.

**Some shell utility use cases**:
- Fetching pending _exercise file paths_ with the keyword "conversion" to pass to my editor:
```sh
vim $(rustlings list --filter "conversion" --paths --unsolved)
```

- Fetching _exercise names_ with keyword "conversion" to pass to `rustlings run`:
```sh
for exercise in $(rustlings list --filter "conversion" --names)
do
    rustlings run ${exercise}
done
```

2. The PR should also fix #465, and will likely fix #585, as well. The bug mentioned in those issues has to do with the way the `watch` command handler fetches the pending exercises.
Going forward, the least recently updated exercises along with all the other exercises in a pending state are fetched.

@fmoko @jrvidal if you've got some time, I'd love an eye on this. If there is anyone else with some experience on (or interest in) the CLI component, please feel free to let me know what you think.

Thanks,
Abdou

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,7 +7,7 @@ use notify::DebouncedEvent;
 use notify::{RecommendedWatcher, RecursiveMode, Watcher};
 use std::ffi::OsStr;
 use std::fs;
-use std::io;
+use std::io::{self, prelude::*};
 use std::path::Path;
 use std::process::{Command, Stdio};
 use std::sync::mpsc::channel;
@@ -58,6 +58,45 @@ fn main() {
             SubCommand::with_name("list")
                 .alias("l")
                 .about("Lists the exercises available in rustlings")
+                .arg(
+                    Arg::with_name("paths")
+                        .long("paths")
+                        .short("p")
+                        .conflicts_with("names")
+                        .help("Show only the paths of the exercises")
+                )
+                .arg(
+                    Arg::with_name("names")
+                        .long("names")
+                        .short("n")
+                        .conflicts_with("paths")
+                        .help("Show only the names of the exercises")
+                )
+                .arg(
+                    Arg::with_name("filter")
+                        .long("filter")
+                        .short("f")
+                        .takes_value(true)
+                        .empty_values(false)
+                        .help(
+                            "Provide a string to match the exercise names.\
+                            \nComma separated patterns are acceptable."
+                        )
+                )
+                .arg(
+                    Arg::with_name("unsolved")
+                        .long("unsolved")
+                        .short("u")
+                        .conflicts_with("solved")
+                        .help("Display only exercises not yet solved")
+                )
+                .arg(
+                    Arg::with_name("solved")
+                        .long("solved")
+                        .short("s")
+                        .conflicts_with("unsolved")
+                        .help("Display only exercises that have been solved")
+                )
         )
         .get_matches();
 
@@ -93,9 +132,51 @@ fn main() {
     let exercises = toml::from_str::<ExerciseList>(toml_str).unwrap().exercises;
     let verbose = matches.is_present("nocapture");
 
-    if matches.subcommand_matches("list").is_some() {
-        exercises.iter().for_each(|e| println!("{}", e.name));
+    // Handle the list command
+    if let Some(list_m) = matches.subcommand_matches("list") {
+        if ["paths", "names"].iter().all(|k| !list_m.is_present(k)) {
+            println!("{:<17}\t{:<46}\t{:<7}", "Name", "Path", "Status");
+        }
+        let filters = list_m.value_of("filter").unwrap_or_default().to_lowercase();
+        exercises.iter().for_each(|e| {
+            let fname = format!("{}", e.path.display());
+            let filter_cond = filters
+                .split(',')
+                .filter(|f| f.trim().len() > 0)
+                .any(|f| e.name.contains(&f) || fname.contains(&f));
+            let status = if e.looks_done() { "Done" } else { "Pending" };
+            let solve_cond = {
+                (e.looks_done() && list_m.is_present("solved"))
+                    || (!e.looks_done() && list_m.is_present("unsolved"))
+                    || (!list_m.is_present("solved") && !list_m.is_present("unsolved"))
+            };
+            if solve_cond && (filter_cond || !list_m.is_present("filter")) {
+                let line = if list_m.is_present("paths") {
+                    format!("{}\n", fname)
+                } else if list_m.is_present("names") {
+                    format!("{}\n", e.name)
+                } else {
+                    format!("{:<17}\t{:<46}\t{:<7}\n", e.name, fname, status)
+                };
+                // Somehow using println! leads to the binary panicking
+                // when its output is piped.
+                // So, we're handling a Broken Pipe error and exiting with 0 anyway
+                let stdout = std::io::stdout();
+                {
+                    let mut handle = stdout.lock();
+                    handle.write_all(line.as_bytes()).unwrap_or_else(|e| {
+                        match e.kind() {
+                            std::io::ErrorKind::BrokenPipe => std::process::exit(0),
+                            _ => std::process::exit(1),
+                        };
+                    });
+                }
+            }
+        });
+        std::process::exit(0);
     }
+
+    // Handle the run command
     if let Some(ref matches) = matches.subcommand_matches("run") {
         let name = matches.value_of("name").unwrap();
 
@@ -123,13 +204,18 @@ fn main() {
         println!("{}", exercise.hint);
     }
 
+    // Handle the verify command
     if matches.subcommand_matches("verify").is_some() {
         verify(&exercises, verbose).unwrap_or_else(|_| std::process::exit(1));
     }
 
+    // Handle the watch command
     if matches.subcommand_matches("watch").is_some() {
         if let Err(e) = watch(&exercises, verbose) {
-            println!("Error: Could not watch your progess. Error message was {:?}.", e);
+            println!(
+                "Error: Could not watch your progess. Error message was {:?}.",
+                e
+            );
             println!("Most likely you've run out of disk space or your 'inotify limit' has been reached.");
             std::process::exit(1);
         }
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -223,7 +298,14 @@ fn watch(exercises: &[Exercise], verbose: bool) -> notify::Result<()> {
                         let filepath = b.as_path().canonicalize().unwrap();
                         let pending_exercises = exercises
                             .iter()
-                            .skip_while(|e| !filepath.ends_with(&e.path));
+                            .filter(|e| filepath.ends_with(&e.path))
```
jrvidal, line 301: I'm inclined to keep `skip_while` here, so in the case of a jump, say, from exercise7 to exercise14, the user can still follow the logical thread of exercises from exercise14. In this way the `chain`ed iterator would work as a second pass just in case you skipped something.

I like that, for a "normal" flow, this does solve #465. But note that if the user skips some exercises, creating an extraneous file will again show some weird behaviour (we will take them back to the first skipped exercise). This is tricky to solve in a stateless fashion. If we do want to solve this case, maybe we could refactor and repurpose `failed_exercise_hint`.

src/main.rs
```diff
@@ -93,9 +132,43 @@ fn main() {
...
+            let filter_cond = filters.split('|').filter(|f| f.len() > 0).any(|f| {
+                e.name.contains(&f) || fname.contains(&f)
+            });
+            let status = if e.looks_done() {
+                "Done"
+            } else {
+                "Pending"
+            };
+            let solve_cond = {
+                (status == "Done" && list_m.is_present("solved"))
```
jrvidal, line 152: I would use the `looks_done` output in these conditions instead of comparing strings.

src/main.rs
```diff
@@ -93,9 +132,43 @@ fn main() {
...
-        exercises.iter().for_each(|e| println!("{}", e.name));
+    // Handle the list command
+    if let Some(list_m) = matches.subcommand_matches("list") {
+        if ["paths", "names"].iter().all(|k| !list_m.is_present(k)) {
+            println!("{:<17}\t{:<46}\t{:<7}", "Name", "Path", "Status");
+        }
+        let filters = list_m.value_of("filter").unwrap_or_default().to_lowercase();
+        exercises.iter().for_each(|e| {
+            let fname = format!("{}", e.path.display());
+            let filter_cond = filters.split('|').filter(|f| f.len() > 0).any(|f| {
```
jrvidal, line 143: I think this needs a `cargo fmt`.

src/main.rs
```diff
@@ -58,6 +58,45 @@ fn main() {
...
+                )
+                .arg(
+                    Arg::with_name("filter")
+                        .long("filter")
+                        .short("f")
+                        .takes_value(true)
+                        .empty_values(false)
+                        .help(
+                            "Provide a string to match the exercise names.\
+                             Pipe separated patterns are acceptable."
```
jrvidal, line 83: Not that it is a big deal, but comma-separated patterns will work as well (not usually in file names) and they don't need to be quoted.

src/main.rs
```diff
@@ -58,6 +58,45 @@ fn main() {
             SubCommand::with_name("list")
                 .alias("l")
                 .about("Lists the exercises available in rustlings")
+                .arg(
+                    Arg::with_name("paths")
```
jrvidal, line 62: Personal opinion: I don't see much value in adding format-controlling options. If a user wants to extract a particular column, they could probably figure out some shell hackery to do so (e.g. `cut`).

The same almost applies for the filtering options (they could use `grep`), except perhaps `filter` if we wanted to save some disk reads (probably overkill).

src/main.rs
jrvidal, line 143: Maybe `.trim()` before checking for length?

src/main.rs
```diff
@@ -223,7 +298,14 @@ fn watch(exercises: &[Exercise], verbose: bool) -> notify::Result<()> {
...
                         let pending_exercises = exercises
                             .iter()
-                            .skip_while(|e| !filepath.ends_with(&e.path));
+                            .filter(|e| filepath.ends_with(&e.path))
+                            .chain(
+                                exercises
+                                    .iter()
+                                    .filter(|e| !e.looks_done() && !filepath.ends_with(&e.path))
+                            );
+                            // .skip_while(|e| !e.is_solved());
```
jrvidal, line 307: Forgotten?

jrvidal (commented):
Some minor comments, but this looks good @AbdouSeck :rocket:

src/main.rs
AbdouSeck: Hi @jrvidal thanks for the very thorough review! I struggled with this update, but I see your point about somehow keeping `skip_while` for the maintenance of a logical thread. However, it should be mentioned that keeping `skip_while` and `chaining` may lead to the `pending_exercises` iterator containing duplicates. In my naive thinking, going back to the skipped exercise sounded preferable to doing duplicate compilations. But like you said, it's hard to fix this issue with maintaining some actual state (not just whether the "I AM NOT DONE" string is removed from the files.)
Do you know of a cleaner way of keeping `skip_while` without introducing duplicates?

src/main.rs
AbdouSeck: Yes, that would be more idiomatic here. Done.

src/main.rs
AbdouSeck: Yes, this was definitely needed. Done.

src/main.rs
AbdouSeck: I switched the pattern separator to a comma.

src/main.rs
AbdouSeck: Yes, you're right. Done.

src/main.rs
AbdouSeck: Definitely forgotten. Fixed.

src/main.rs
AbdouSeck: I personally don't want to assume that someone has `grep`, `cut`, `sed`, or `awk` on their machine to split and extract fields, so I think that the filtering adds quite a bit of value here. Controlling the displayed fields allows for the _shell utility use cases_ I mentioned in the PR. Otherwise, a user would have to skip the header row first (with `head` or `tail` which may not be available on a user's machine) before going about extracting and filtering on fields.
I would prefer to keep these features in, but I am happen to edit/update for better UX.

src/main.rs
jrvidal: Do those duplicates have an observable effect for the end user? I suspect not (though I haven't given it much thought).

src/main.rs
jrvidal: I think that is safe to assume that there are _some_ CLI tools in the user's operating system that would allow for similar capabilities, _and_ that the use cases you showed above already presuppose a good deal of familiarity with CLI tools.

But as I said, this is a matter of taste, I don't really have a strong opinion :+1:

src/main.rs
AbdouSeck: Yes. The updated file does not get compiled again when the following is used:
```
let pending_exercises = exercises
    .iter()
    .skip_while(|e| !e.path.ends_with(&filepath))
    // .filter(|e| filepath.ends_with(&e.path))
    .chain(
        exercises
            .iter()
            .filter(|e| !e.looks_done() && !filepath.ends_with(&e.path))
);
```
Perhaps, I missing something here?

src/main.rs
AbdouSeck: Let's leave it as is for now. It's one file update, so it should be easy to retract if it ends up being a nightmare of a feature.

src/main.rs
jrvidal: But this iterator is passed to `verify()` which does not necessarily compile and run every single exercise within it. It stops as soon as it finds one that is not actually done. So I'm having a hard time imagining a case where we end up actually compiling twice the same file.

AbdouSeck: The iterator itself will contain the duplicate, but like you said `verify` does stop upon encountering a failed exercise. So, it may not matter at all if the iterator contains duplicates.
However, the chaining using `skip_while` is not working at all, since the updated file does not get compiled. Please feel free to toggle between `filter` and `skip_while` and let me know if you can observe when I am referring to.

jrvidal: I do see weird behaviour toggling between the commented out version and the current one. However,  when I suggested keeping `skip_while`, I was referring to the old version:
```rust
.skip_while(|e| !filepath.ends_with(&e.path))
```
I think it works fine in this way.

AbdouSeck: Keeping that line was the cause of the bug reported in the target issues. I have explained why in issue #585.

When you have the original line, it makes it so that if you skip the order given by the `watch` command and solve the last problem (which is currently `exercises/conversions/from_str.rs`), the program exits and claims that you completed all the exercises, even when you've still got pending exercises.

Please feel free to try and reproduce that bug by invoking `watch` and then finishing `exercises/conversions/from_str.rs`. Regardless of whether or not you've got other pending exercises, the program will exit and print out an erroneous message you've finished all exercises.

jrvidal: Sorry, I think I might have been too terse, or maybe it's just my obtuseness :sweat_smile:

What I'm suggesting is something along the lines of:
```rust
// Original line, unchanged
.skip_while(|e| !filepath.ends_with(&e.path))
// NEW
.chain(
  exercises
    .iter()
    .filter(|e| /* your thing */)
)
```

AFAICT this:

* Avoids #585 by doing a second pass and looking for unfinished exercises.
* Avoids #465 by using this second pass as fallback when we get an event about an extraneous file.
* Keeps the current behavior where we progress through the exercises list in order (at least, by default).
* Never verifies an exercise twice.

AbdouSeck: I think the weird behavior I was observing with `skip_while` is due to how I set up the condition: `.skip_while(|e| !e.path.ends_with(&filepath))`. It should have been `.skip_while(|e| !filepath.ends_with(&e.path))`. `filepath` is the canonical path. That's why the least recently modified file was not being checked. I think that's what I was missing. Thank you!

While I am more than fine with this current behavior, I think this behavior _does_ verify files twice. You can see this by updating an already solved exercises that precedes the one `rustlings watch` is suggesting you solve. For example, if you have already solved the first 5 exercises and you're being prompted to solve the 6th. However, you decide, for whatever reason, to go back to exercise 1 and update it. Upon updating exercise 1, `watch` will verify exercises 1 through 6 again. I am happy to demo this over slack or Zoom, whenever time allows. My description of the behavior may not be clear enough.

In any case, like I said, I am fine with this behavior but I do want to make sure we're on the same page and are seeing the same behavior reflected.

jrvidal: Ah, I see, yes, that is a non ideal UX. <sub>Though note that my concern with "double verification" was more about not observing rustlings verify the same exercise twice on the same verification batch.</sub> If a user wanted to re-evaluate their solution, they'd need to re-add `I AM NOT DONE`, which might not be entirely obvious :confused:

I think that we can live with that, it is after all the _current_ behavior. I think the `chain`ed fallback is an improvement anyway. @fmoko Thoughts?

AbdouSeck: @fmoko, the current code is in the state @jrvidal suggested. It works fine, to me. It can be merged. If this comes back again in the form of an issue, we could toggle back to the `filter` solution as an alternative.

Thanks,
Abdou

fmoko (approved):

PR #600, opened 2020-12-14
docs: use git commands to find the latest tag

numbertheory:
We don't need to manually copy and paste the latest tag, we can have bash do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,12 @@ To install Rustlings. Same as on MacOS/Linux, you will have access to the `rustl
 
 ## Manually
 
-Basically: Clone the repository, checkout to the latest tag, run `cargo install`.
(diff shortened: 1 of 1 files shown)


jrvidal:
Hey @numbertheory thanks for the suggestion :+1:  It would be ideal if we could have equivalent instructions for Windows/Powershell, so we can keep the README useful for all platforms.

fmoko:
Hmm, @jrvidal makes a good point, plus this doesn't work in Fish (the shell I personally use) as well as any other shell that doesn't use bash interpolation syntax, whereas keeping the version manually updated is just one chore of many when releasing a new version. I think I like it better as it is now, unless there's some sort of way we can keep it platform-agnostic without adding 10 different code blocks to the readme.

numbertheory:
Oh ok, I'll revisit this later then.

PR #608, opened 2020-12-30
Update structs3.rs

sundevilyang:
as a totally newbie to Rust, I don't know panic statement from https://doc.rust-lang.org/book/ and rustlings in the beginning. After a hard searching of [should_panic], then I figure out panic statement. 

So it's helpful to tell the learner that write a panic statement here.

### DIFF
--- a/exercises/structs/structs3.rs
+++ b/exercises/structs/structs3.rs
@@ -15,7 +15,7 @@ struct Package {
 impl Package {
     fn new(sender_country: String, recipient_country: String, weight_in_grams: i32) -> Package {
         if weight_in_grams <= 0 {
-            // Something goes here...
+            // panic statement goes here...
         } else {
             return Package {
                 sender_country,


jfchevrette:
As a newcomer to rust I also got a bit confused on that one. Looking at the tests and seeing `#[should_panic]` gave me a hint but I don't think understanding macros and tests should be expected of new users at this point in the rustlings exercices.

As such I like the idea behind this change, however I personally think this is giving too much as I think we want the user to think about that should happen if the new function can't create a new Package object.

I think a better idea would be to hint at it without directly telling where to put the panic. For example, could we add a sentence or two at the top of the file to tell the the user to pay attention to the two functions which should return something (the `???` markers) in comparison to the new function which should `return a new Package, otherwise what can be done if it can't? (and hint at: https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html)`. I think this would guide the user better into understanding when and where panics are appropriate.

Zerotask:
@jfchevrette 
I agree with you. A hint is probably better here. Otherwise it could be too easy. There should be a little challenge and if people have to read the book to find the solution, then they will learn more.
Otherwise they could recognize panic as a macro and just use it without knowing why / what it does.

diannasoreil (approved):

diannasoreil:
@all-contributors please add @sundevilyang for content

PR #609, opened 2020-12-30
fix: typo in default out text

xehpuk:


fmoko (approved):
Thanks! Figured there would be at least one.

PR #615, opened 2021-01-03
docs: mention flatten in the options2 hint

wsh:
[`Option::flatten`](https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten) is so useful, I thought I would mention it as an alternative to "stacking Options".

Thanks for considering this, I've been having fun with `rustlings` and hope to make more substantial contributions in the future. :smile:


fmoko (approved):
Thank you!

fmoko:
@all-contributors please add @wsh for content

PR #643, opened 2021-02-09
Changed errors3 mode from test to compile.

apogeeoak:
The mode for `errors3` is set as `test` but the file does not contain any tests.

The generated output is hidden because it is in test mode. This obfuscates a successful result, impeding understanding.

An unrelated auto-formatting change removes trailing whitespace further down in the file.


jrvidal:
@all-contributors please add @apogeeoak for code

PR #646, opened 2021-02-10
Added iterators5.rs exercise.

apogeeoak:
Added an iterator exercise that counts values in a hash map that equal a given value.

Incorporates more iterator methods into the exercises. Provides a use case for the `fold` method.

### DIFF
--- a/exercises/standard_library_types/iterators5.rs
+++ b/exercises/standard_library_types/iterators5.rs
@@ -1,11 +1,14 @@
 // iterators5.rs
 
-// Rustling progress is modelled using a hash map. The name of the exercise is
-// the key and the progress is the value. Two counting functions were created
-// to count the number of exercises with a given progress. These counting
-// functions use imperative style for loops. Recreate this counting
-// functionality using iterators.
-// Execute `rustlings hint iterators5` for hints.
+// Let's define a simple model to track Rustlings exercise progress. Progress
+// will be modelled using a hash map. The name of the exercise is the key and
+// the progress is the value. Two counting functions were created to count the
+// number of exercises with a given progress. These counting functions use
+// imperative style for loops. Recreate this counting functionality using
+// iterators. Only the two iterator methods (count_iterator and
+// count_collection_iterator) need to be modified.
+// Execute `rustlings hint
+// iterators5` for hints.
 //
 // Make the code compile and the tests pass.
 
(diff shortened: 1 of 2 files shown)


jfchevrette:
Overall I like the idea behind this exercise and the fact it is 'meta' in the sense that it is using progress data from rustlings itself which is pretty neat.

I'm however concerned that it may easily break when new exercises are added in the future. Is there some way we can ensure it doesn't?

apogeeoak:
This exercise does not use actual progress data. It only simulates progress. The strings that are used as keys can be anything. They just happen to be filenames from the exercises. If the exercise files are renamed, exercises are deleted, or new exercises are added this exercise will still work.

jfchevrette:
Sorry for the confusion @apogeeoak, I'll update my previous comment

exercises/standard_library_types/iterators5.rs
```diff
@@ -0,0 +1,113 @@
+// iterators5.rs
+
+// Rustling progress is modelled using a hash map. The name of the exercise is
```
fmoko, line 3:
```suggestion
// Let's reimplement a simple version of Rustlings exercise progress!
// This will be modelled using a hash map. The name of the exercise is
```

exercises/standard_library_types/iterators5.rs
```diff
@@ -0,0 +1,113 @@
...
+// the key and the progress is the value. Two counting functions were created
+// to count the number of exercises with a given progress. These counting
+// functions use imperative style for loops. Recreate this counting
+// functionality using iterators.
+// Execute `rustlings hint iterators5` for hints.
+//
+// Make the code compile and the tests pass.
+
+// I AM NOT DONE
+
```
fmoko, line 13: Generally, I would leave comments somewhere in this exercise with notes that explain:

- What a "stack" is (aka a vec of hashmaps)
- That you probably shouldn't touch the `_for` methods, but rather implement the iterative version in the other methods
- A delineator above the tests module telling the user not to touch what's below it

exercises/standard_library_types/iterators5.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut count = 0;
+    for val in map.values() {
+        if val == &value {
+            count += 1;
+        }
+    }
+    count
+}
+
+fn count(map: &HashMap<String, Progress>, value: Progress) -> usize {
```
fmoko, line 33: I'd rename these to `count_iterator` or something, to keep them separate from the `_for` methods.

fmoko (commented):
Overall a great concept! I have some ideas that could make this exercise a little easier to approach, feel free to implement those.

apogeeoak:
Thanks for the suggestions. I improved the exercise explanation and clarified which functions need modified.

I renamed `stack` to `collection` to avoid confusion and added a comment describing the collection.

fmoko (approved):
Looks good to me!

PR #647, opened 2021-02-11
fix: added missing required section before quiz1

iliasbhal:
The comment was missing a line.
It forced me out of the `rustling watch` to run the quiz separately.


iliasbhal:
My bad, the quizzes are not meant to cover all previous modules.

PR #648, opened 2021-02-12
Moved iterators2.rs errors out of tests.

### LINKED ISSUE #359: iterators2 doesn't compile before step 2
The instructions say to first make the first 2 tests pass, but the last 2 tests have compilation errors. Of course, it can easily be fixed by commenting them out. Still, maybe it would be nicer if Step 1 could be completed without having to do that?

apogeeoak:
Moved all errors out of tests and into the base module.
The exercise compiles after step 1 allowing the user to test `capitalize_first`. The remaining steps fix the other two failing tests.

The exercise still showcases the power of `collect` without requiring the user to complete the entire exercise before checking their work.

The hint to solve the `capitalize_first` function was moved to the hint in `info.toml`.

Closes #359


fmoko (approved):
Thanks!

PR #649, opened 2021-02-12
Enabled iterators3.rs to run without commented out tests.

apogeeoak:
Adapted all but one compiler error into logic errors. The exercise can be tested after completing step one, without the need for the tests to be commented out.

The assertions against debug output are kept to allow the code to compile before `result_with_list` and `list_of_results` are completed. They also prevent the compiler from revealing the answer too eagerly.
﻿


fmoko (approved):
Thank you!

PR #660, opened 2021-02-27
Fix incorrect hashmap2 test

arthas168:
The test `at_least_five_types_of_fruits` implies that there should be `at least` five types of fruit in the basket, but the test is actually checking for `exactly` five types of fruit, which was a bit misleading for newcomers like me :) 

A simple change from `==` to `>=` should do the trick and properly check for the `at least 5` condition.


AbdouSeck (approved):

fmoko (approved):

fmoko:
@all-contributors please add @arthas168 for content

PR #668, opened 2021-03-13
fix(installation): first PowerShell command

cseltol:
ExecutionPolicy to RemoteSigned command must be fixed because the old command was getting denied access to change the Execution Policy property


fmoko (approved):

fmoko:
@all-contributors please add @cseltol for docs

PR #670, opened 2021-03-15
Fix Rust's HashMap equivalent in C++

hpwxf:
C++ `map` is more like BTreeMap in Rust.

`unordered_map` in C++(11) is the equivalent of `HashMap` in Rust.
(+ additional like for references).


jrvidal:
@all-contributors please add @hpwxf for content

Issue #651, opened 2021-02-15
Collections exercises run and hint command does not match standard pattern


arnoldn1:
rustc version: rustc 1.49.0 (e1884a8e3 2020-12-29)
rustlings version: rustlings 4.3.0

The README gives the `run` and `hint` commands as:
```bash
rustlings run myExercise1
```
``` bash
rustlings hint myExercise1
```
where `myExercise1` is the exercise file name. This remains true for other exercise groups where the file names do not match their parent directory name e.g., conversions.

However, for the collections exercises the pattern is instead:
`rustlings run collections1` to run `vec1`
`rustlings run collections2` to run `vec2`
`rustlings run collections3` to run `hashmap1`
`rustlings run collections4` to run `hashmap2`

There is a comment in each file that specifies the correct `hint` command but not `run` command. The lack of consistency leads to some confusion.

fmoko:
Yup, I think this has been an inconsistency ever since these exercises came into being, since the folder name is `collections`. A PR renaming the files themselves to `collections1` - `collections4` would be appreciated!

lauralindzey:
@fmoko -- It looks like this Issue could be closed?

fmoko:
@lauralindzey Thanks for the nudge!

PR #674, opened 2021-03-18
Fix/collections exercises naming

Morsicus:
Hello,

## Issue

Apparently, the collections exercises are not following the standard naming.

Inside `collections/` directory we have files named `vecX.rs` and `hashmapX.rs` but the rustlings CLI expect them to be named `collectionsX.rs`

## Proposal

As mention in https://github.com/rust-lang/rustlings/issues/651, we could rename files inside `collections/` but in my opinions the possibility to have different names inside a directory is good for clarity. 

Furthermore, the same pattern (ie. Exercises name different from directory name) is also present in `error_handling/`, `standard_directory_types/` and `conversions/` directories.

Obviously, if you think that the solution in #651 is better, i'll change it :)

### DIFF
--- a/exercises/collections/hashmap1.rs
+++ b/exercises/collections/hashmap1.rs
@@ -8,7 +8,7 @@
 //
 // Make me compile and pass the tests!
 //
-// Execute the command `rustlings hint collections3` if you need
+// Execute the command `rustlings hint hashmap1` if you need
 // hints.
 
 // I AM NOT DONE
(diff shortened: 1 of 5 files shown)


WowSuchRicky:
👍 , just ran into this as well and was about to make a PR.

AbdouSeck (approved):
This looks good to me. @jrvidal @fmoko any thoughts?

fmoko:
Yeah, looks OK to me too.

PR #675, opened 2021-03-18
Updated option2.rs

k12ish:
Renaming uninformative names like `optional_value`, `value`, `optional_values_vec` and `value` helps users distinguish between the two parts of the task.


fmoko (approved):

fmoko:
@all-contributors please add @k12ish for content

Issue #364, opened 2020-04-21
Excess use of colour and emoji is an accessibility issue


rjek:
Hello,

While I can disable the use of colour using CLICOLOR=0 in the environment, sadly I can find no way to disable the use of emoji.

Is it possible to have rustlings respond to EMOJI=0 in the environment?

fmoko:
@rjek Good idea! Thanks for the suggestion!

blerchy:
@fmoko Thank you for accepting my PR! May I suggest closing this issue?

fmoko:
I think so! Thanks for implementing it~

PR #676, opened 2021-03-19
Replace emojis when NO_EMOJI env variable present

blerchy:
When the `NO_EMOJI` environment variable is set, replace/eliminate emojis in output.

For instance, without `NO_EMOJI` set (normal behaviour):
Screen Shot 2021-03-19 at 2 25 48 AM

With `NO_EMOJI` set:
Screen Shot 2021-03-19 at 2 26 30 AM

Addresses #364.

### DIFF
--- a/src/exercise.rs
+++ b/src/exercise.rs
@@ -127,9 +127,10 @@ name = "{}"
 path = "{}.rs""#,
                     self.name, self.name, self.name
                 );
-                let cargo_toml_error_msg = match env::var("NO_EMOJI").is_ok() {
-                    true => "Failed to write Clippy Cargo.toml file.",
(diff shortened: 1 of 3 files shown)


fmoko:
Out of curiosity, is there any prior work for this environment variable in other projects? Is there some sort of standard naming here that works with multiple CLI applications when you have it set?

blerchy:
Unfortunately, not that I could find. I started with issue #364 and looked for an environment variable to suit it, eventually settling on `NO_EMOJI` after the similar `NO_COLOR` from [https://no-color.org](https://no-color.org/). There is however no precedent that I could find for `NO_EMOJI` in particular.

fmoko (approved):

fmoko:
@all-contributors please add @blerchy for code

PR #687, opened 2021-03-31
fix: improve function argument type in functions3.rs

flakolefluk:
Hi!
Going through the exercise functions3.rs, it feels like using u32 instead of i32  could be an improvement. Even though passing a negative number won't make the compiler complaint and the loop will exit immediately, it feels that it lacks of intention (to the eyes of someone who's learning like me). I hope this small change makes sense.

### DIFF
--- a/exercises/functions/functions3.rs
+++ b/exercises/functions/functions3.rs
@@ -7,7 +7,7 @@ fn main() {
     call_me();
 }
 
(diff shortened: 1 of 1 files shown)


fmoko (approved):

fmoko:
Thanks!

@all-contributors add @flakolefluk for content

Issue #691, opened 2021-04-04
conversions/try_from_into doesn't work so well with `?` operator


tlyu:
[Thanks to multiple people on Discord for helping me to understand some of these issues.]

I tried to solve conversions/try_from_into by using the `?` operator in a few places to propagate errors. This led me down a surprisingly deep rabbit hole where I discovered:
* The `?` operator can't convert a `num::TryFromIntError` into a `String` because it relies on `From`, not `ToString` or something
* Trying to use `Box<dyn error::Error>` doesn't work because some of the tests explicitly use `Result<Color, String>` instead of `Result<Color, _>`
* Additionally, the tests that `assert_eq!(c, Ok(Color{...}))` don't work even with `Result<Color, _>` as the target, because `Box<dyn error::Error>` doesn't implement `PartialEq` for what turn out to be very complicated reasons

Suggested fixes:
* Replace `Result<Color, String>` with `Result<Color, _>` in the tests; this allows users to attempt to use custom error types to propagate errors (you might want to add a warning against using `Box<dyn error::Error>` unless you implement the following item)
* Replace `assert_eq!(c, Ok(Color{...}))` with `assert!(c.is_ok()); assert_eq!(c.unwrap(), Color{...})` so that using `Box<dyn error::Error>` as the error type works (there would no longer be an `assert_eq!()` that tries to enforce `PartialEq` on the error type)
* Replace `type Error = String` in the skeletal `TryFrom` implementations with `type Error = Box<dyn error::Error>` or similar (though maybe there's a pedagogical reason to avoid trait objects in this particular exercise? I think it's ok to use because this is one of the last exercises in the suggested sequence)

Fixes elsewhere?:
* Add a feature to the standard library that implements `PartialEq<T> for Box<dyn T>` on appropriate types

Do my interpretations seem correct? I could work on a pull request for these suggestions, if you like. Or do you have alternative solutions that you would rather pursue? Thanks!

tlyu:
Fixed in #693.

PR #693, opened 2021-04-04
fix: use trait objects for try_from_into

tlyu:
Use `Box<dyn error::Error>` to allow solutions to use `?` to propagate
errors.  In the tests, explicitly check `is_ok()` instead of trying to
force the error type to `String` (or other `PartialEq` type) using
`assert_eq!()`.

### DIFF
--- a/exercises/conversions/from_str.rs
+++ b/exercises/conversions/from_str.rs
@@ -2,6 +2,7 @@
 // Additionally, upon implementing FromStr, you can use the `parse` method
 // on strings to generate an object of the implementor type.
 // You can read more about it at https://doc.rust-lang.org/std/str/trait.FromStr.html
+use std::error;
 use std::str::FromStr;
 
 #[derive(Debug)]
@@ -23,7 +24,7 @@ struct Person {
 // If everything goes well, then return a Result of a Person object
 
 impl FromStr for Person {
-    type Err = String;
+    type Err = Box<dyn error::Error>;
     fn from_str(s: &str) -> Result<Person, Self::Err> {
     }
 }
--- a/exercises/conversions/try_from_into.rs
+++ b/exercises/conversions/try_from_into.rs
@@ -3,6 +3,7 @@
 // instead of the target type itself.
 // You can read more about it at https://doc.rust-lang.org/std/convert/trait.TryFrom.html
 use std::convert::{TryFrom, TryInto};
+use std::error;
 
 #[derive(Debug, PartialEq)]
 struct Color {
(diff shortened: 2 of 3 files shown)


tlyu:
This should fix #691.

AbdouSeck (changes requested):
Hi @tlyu, this is a great idea. Thanks for both the issue and the PR. While you're already in the `conversions` exercises, would you mind updating the [`from_str.rs`](https://github.com/rust-lang/rustlings/blob/main/exercises/conversions/from_str.rs) exercise as well? The signature implementation of the `std::str::FromStr` trait could use an update similar to the one done here.

Thank you!
Abdou

tlyu:
Thanks for the review! I added a change for `from_str.rs` as well. I didn't try to propagate the error from `parse::<usize>()` in that exercise, so I didn't run into that problem. The difference is that with `try_from_into.rs`, there might be a lot of code repetition if the `?` operator isn't usable for error propagation. (This is especially true with the tuple implementation, because as far as I can tell, it's not possible to directly iterate over tuples.)

AbdouSeck (approved):
Looks great. Thank you!

AbdouSeck:
@all-contributors add @tlyu for code and content

PhilMiller:
(newcomer to Rust here; I'm encountering this as I work through the exercises)

Being able to use `?` to propagate errors is a definite improvement in the solution to `try_from_into`, and in parsing failures in `from_str`.

However, for bad value cases in `from_str` (e.g. wrong array length, or empty name string), it becomes very obscure how one should write the `Err(...)` to be returned. Am I supposed to implement and return a concrete type? Someone on Discord (`cysant`) suggested `Err("error".into())`, which works, but I have no idea what actual type it's constructing/converting.

tlyu:
> (newcomer to Rust here; I'm encountering this as I work through the exercises)
> 
> Being able to use `?` to propagate errors is a definite improvement in the solution to `try_from_into`, and in parsing failures in `from_str`.
> 
> However, for bad value cases in `from_str` (e.g. wrong array length, or empty name string), it becomes very obscure how one should write the `Err(...)` to be returned. Am I supposed to implement and return a concrete type? Someone on Discord (`cysant`) suggested `Err("error".into())`, which works, but I have no idea what actual type it's constructing/converting.

I opened #755 to add hints that should help with this. Please give feedback there if you like. Thanks!

dahifi:
What's the idiomatic version of this look like now? 

Specifically looking at this one, it seems a bit hairy. https://github.com/alexxroche/rustlings-idiomatic-solutions/blob/master/exercises/conversions/try_from_into.rs_/try_from_into_match.rs

PR #715, opened 2021-04-20
feat: Replace clap with argh

fmoko:
I’ve been wanting to do this for a while, but always procrastinated on it. We’ve been using Clap since the 2.0 rewrite, but Clap is known to be a fairly heavy library. Since Rustlings is usually peoples’ first contact with a Rust compilation, I think it’s in our best interests that this complation is as fast as possible. In effect, replacing Clap with the smaller, structopt-style `argh` reduces the amount of crates needing to be compiled from 82 to 60.

I also think this makes the code way easier to read, we don’t need to use Clap’s methods anymore, but can switch over to using pure Rust methods, e.g., switches are booleans, options are Option<String>s or the like, and subcommands are just structs.

I'll leave this up for a couple of days for people to review and test, and this'll also be the capstone for the next version release, I think.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -201,13 +201,13 @@ fn main() {
         }
 
         Subcommands::Run(subargs) => {
-            let exercise = find_exercise(subargs.name, exercises);
+            let exercise = find_exercise(&subargs.name, &exercises);
 
             run(&exercise, verbose).unwrap_or_else(|_| std::process::exit(1));
         }
 
         Subcommands::Hint(subargs) => {
-            let exercise = find_exercise(subargs.name, exercises);
+            let exercise = find_exercise(&subargs.name, &exercises);
 
             println!("{}", exercise.hint);
         }
@@ -285,14 +285,12 @@ fn spawn_watch_shell(failed_exercise_hint: &Arc<Mutex<Option<String>>>) {
     });
 }
 
-fn find_exercise(name: String, exercises: Vec<Exercise>) -> Exercise {
-    let matching_exercise = |e: &Exercise| name == e.name;
-
+fn find_exercise<'a>(name: &str, exercises: &'a [Exercise]) -> &'a Exercise {
     exercises
-        .into_iter()
-        .find(matching_exercise)
+        .iter()
+        .find(|e| e.name == name)
         .unwrap_or_else(|| {
-            println!("No exercise found for your given name!");
+            println!("No exercise found for '{}'!", name);
             std::process::exit(1)
         })
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,7 @@
 use crate::exercise::{Exercise, ExerciseList};
 use crate::run::run;
 use crate::verify::verify;
-use clap::{crate_version, App, Arg, SubCommand};
+use argh::FromArgs;
 use console::Emoji;
 use notify::DebouncedEvent;
 use notify::{RecommendedWatcher, RecursiveMode, Watcher};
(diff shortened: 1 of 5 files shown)


apogeeoak:
I like this change. It simplifies the code base and removes multiple dependencies while retaining existing functionality. I just have a few clippy suggestions.

src/main.rs
```diff
@@ -22,85 +22,91 @@ mod exercise;
 mod run;
 mod verify;
 
+// In sync with crate version
+const VERSION: &'static str = "4.3.0";
```
apogeeoak, line 26: Constants by default have static lifetime so the lifetime does not need to be specified.

https://rust-lang.github.io/rust-clippy/master/index.html#redundant_static_lifetimes

src/main.rs
```diff
@@ -130,144 +136,144 @@ fn main() {
...
-        println!("{}", exercise.hint);
-    }
+            run(&exercise, verbose).unwrap_or_else(|_| std::process::exit(1));
+        }
 
-    // Handle the verify command
-    if matches.subcommand_matches("verify").is_some() {
-        verify(&exercises, verbose).unwrap_or_else(|_| std::process::exit(1));
-    }
+        Subcommands::Hint(subargs) => {
```
apogeeoak, line 216: Carryover from previous commit but `Subcommands::Run` and `Subcommands::Hint` repeat the same logic.

src/main.rs
```diff
@@ -130,144 +136,144 @@ fn main() {
...
+            let mut exercises_done: u16 = 0;
+            let filters = subargs.filter.clone().unwrap_or_default().to_lowercase();
+            exercises.iter().for_each(|e| {
+                let fname = format!("{}", e.path.display());
+                let filter_cond = filters
+                    .split(',')
+                    .filter(|f| !f.trim().is_empty())
+                    .any(|f| e.name.contains(&f) || fname.contains(&f));
+                let status = if e.looks_done() {
+                    exercises_done = exercises_done + 1;
```
apogeeoak, line 160: Carryover from previous commit but this can be shortened.
https://rust-lang.github.io/rust-clippy/master/index.html#assign_op_pattern
```suggestion
                    exercises_done += 1;
```

fmoko:
@apogeeoak Fixed, thanks!

AbdouSeck (commented):
Hi @fmoko, this is a great way to trim the code size and reduce complication times. The only thing that departs from the previous behavior is the handling of the `--solved` and `--unsolved` options to the `list` subcommand. With `clap`, these options are mutually exclusive, i.e. `rustlings list --solved --unsolved` causes an argument parsing error. I took a quick glance at the docs for `argh` and could not find any support for mutual exclusions. In any case, if it's at all possible to handle those options in the code, I would recommend it. Otherwise, we can just document that `--solved` and `--unsolved` are mutually exclusive. Providing both has no real filtering effect, as the whole list of exercises will be displayed. Thoughts?

fmoko:
@AbdouSeck You're right, my thought was to not document that solved and unsolved are exclusive, since I don't think anyone will use them. It's sad that argh doesn't support something like this, but I think `clap` is the only Rust CLI crate that _does_.

And good call on the integration tests, I'll fix those.

src/main.rs
```diff
@@ -130,144 +136,129 @@ fn main() {
...
+                exercises.len(),
+                percentage_progress
+            );
+            std::process::exit(0);
+        }
 
-        run(&exercise, verbose).unwrap_or_else(|_| std::process::exit(1));
-    }
+        Subcommands::Run(subargs) => {
+            let exercise = find_exercise(subargs.name, exercises);
```
apogeeoak, line 204: Borrow parameters if `find_exercise` signature changes.
```suggestion
            let exercise = find_exercise(&subargs.name, &exercises);
```

src/main.rs
```diff
@@ -130,144 +136,129 @@ fn main() {
...
 
-        let exercise = exercises
-            .iter()
-            .find(|e| name == e.name)
-            .unwrap_or_else(|| {
-                println!("No exercise found for your given name!");
-                std::process::exit(1)
-            });
+        Subcommands::Hint(subargs) => {
+            let exercise = find_exercise(subargs.name, exercises);
```
apogeeoak, line 210: Borrow parameters if `find_exercise` signature changes.
```suggestion
            let exercise = find_exercise(&subargs.name, &exercises);
```

AbdouSeck (approved):
Everything looks good to me 👍🏿

src/main.rs
```diff
@@ -22,85 +22,91 @@ mod exercise;
...
+    /// show the executable version
+    #[argh(switch, short = 'v')]
+    version: bool,
+    #[argh(subcommand)]
+    nested: Option<Subcommands>,
+}
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand)]
+enum Subcommands {
```
jrvidal, line 43: `Subcommand`?

jrvidal (approved):
Sorry, my review did not arrive in time :disappointed: Nothing important though, I like this :+1:

PR #727, opened 2021-04-24
Support to start "watch" at any exercise

panpansuannai:
When we work on the latter exercises it takes a lot of time to compile from the first exercise. So if we can start "watch" from the exercise which is being done, it will save a lot of time.


Zerotask:
Additionally rustlings could remember the last succesfull exercise (with a simple text file) and with a flag you could start there. This would be even easier than remembering it by yourself and typing it.

panpansuannai:
> Additionally rustlings could remember the last successful exercise (with a simple text file) and with a flag (`-l`, `--last` or `-c`, `--continue`) you could start there. This would be even easier than remembering it by yourself and typing it.

Sorry, I didn't notice that.

Issue #657, opened 2021-02-24
"compilation failed" on "rustlings watch"


lemmy04:
I've done the manual installation of rustlings, now when I try to run "rustlings watch" i get some compilation error:

! Compiling of exercises/variables/variables1.rs failed! Please try again. Here's the output:
error[E0425]: cannot find value `x` in this scope
  --> exercises/variables/variables1.rs:12:5
   |
12 |     x = 5;
   |     ^ not found in this scope

Any ideas?

RansomTime:
Hi @lemmy04.

This is actually what's expected to happen. Your goal in the exercises is to get the code to compile. There's a README.md in each directory with a link to the Rust book about the topic (this first one is [How Functions Work](https://doc.rust-lang.org/book/ch03-03-how-functions-work.html)). The rustlings watch command will automatically update whenever you make a change to the file and attempt to re-compile the code. You can also type hint to get a hint on the current exercise.

Hope that helps.

lemmy04:
...so its basically not for people who don't know anything about rust at all...?

RansomTime:
Sorry I gave the wrong link. The book chapter I meant to link to is [Variables and Mutability](https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html)

Reading that section of the book should give you the information you need to do the first set of exercises. The main project's README.md describes doing exercises like this:

----
## Doing exercises

The exercises are sorted by topic and can be found in the subdirectory `rustlings/exercises/<topic>`. For every topic there is an additional README file with some resources to get you started on the topic. We really recommend that you have a look at them before you start.

The task is simple. Most exercises contain an error that keeps them from compiling, and it's up to you to fix it! Some exercises are also run as tests, but rustlings handles them all the same. To run the exercises in the recommended order, execute:

```bash
rustlings watch
```

This will try to verify the completion of every exercise in a predetermined order (what we think is best for newcomers). It will also rerun automatically every time you change a file in the `exercises/` directory. If you want to only run it once, you can use:

```bash
rustlings verify
```

This will do the same as watch, but it'll quit after running.

In case you want to go by your own order, or want to only verify a single exercise, you can run:

```bash
rustlings run myExercise1
```

In case you get stuck, you can run the following command to get a hint for your
exercise:

``` bash
rustlings hint myExercise1
```

---

I think that without reading the book then the exercises won't really make sense. If this doesn't seem to fit with the way you'd like to learn there are links to alternative resources that might be a better fit for you in the README

[...]

PR #733, opened 2021-04-24
feat(intro): Proposal to add successfully compiling exercise as the first exercise.

apogeeoak:
### Proposal to add introductory section

The first exercise in the section successfully compiles to acclimate new users to Rustlings and introduce the I AM NOT DONE comment.

The second exercise simply asks the user to complete a Hello World! program.

The intro section is more to introduce Rustlings than to teach Rust.

### DIFF
--- a/exercises/intro/README.md
+++ b/exercises/intro/README.md
@@ -1,0 +1,8 @@
+# Intro
+
+Rust uses the `print!` and `println!` macros to print text to the console.
+
+## Further information
+
+- [Hello World](https://doc.rust-lang.org/rust-by-example/hello.html)
+- [Formatted print](https://doc.rust-lang.org/rust-by-example/hello/print.html)
--- a/exercises/intro/intro1.rs
+++ b/exercises/intro/intro1.rs
@@ -16,4 +16,8 @@ fn main() {
     println!(r#" | |  | |_| \__ \ |_| | | | | | (_| \__ \ "#);
     println!(r#" |_|   \__,_|___/\__|_|_|_| |_|\__, |___/ "#);
     println!(r#"                               |___/      "#);
+    println!();
+    println!("This exercise compiles successfully. The remaining exercises contain a compiler");
+    println!("or logic error. The central concept behind Rustlings is to fix these errors and");
+    println!("solve the exercises. Good luck!");
 }
--- a/exercises/intro/intro1.rs
+++ b/exercises/intro/intro1.rs
@@ -1,0 +1,19 @@
+// intro1.rs
+// About this `I AM NOT DONE` thing:
+// We sometimes encourage you to keep trying things on a given exercise, even
+// after you already figured it out. If you got everything working and feel
+// ready for the next exercise, remove the `I AM NOT DONE` comment below.
+// Execute the command `rustlings hint intro1` for a hint.
+
+// I AM NOT DONE
+
+fn main() {
+    println!("Hello and");
+    println!(r#"       welcome to...                      "#);
+    println!(r#"                 _   _ _                  "#);
+    println!(r#"  _ __ _   _ ___| |_| (_)_ __   __ _ ___  "#);
+    println!(r#" | '__| | | / __| __| | | '_ \ / _` / __| "#);
+    println!(r#" | |  | |_| \__ \ |_| | | | | | (_| \__ \ "#);
+    println!(r#" |_|   \__,_|___/\__|_|_|_| |_|\__, |___/ "#);
+    println!(r#"                               |___/      "#);
+}
--- a/exercises/intro/intro2.rs
+++ b/exercises/intro/intro2.rs
@@ -1,0 +1,9 @@
+// intro2.rs
+// Make the code print a greeting to the world.
+// Execute `rustlings hint intro2` for a hint.
+
+// I AM NOT DONE
+
+fn main() {
+    println!("Hello {}!");
+}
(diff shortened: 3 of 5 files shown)


Zerotask (commented):
I really like your idea of an introduction to Rustlings.

Although I would implement that in another way.
At this step a learner usually has rustlings installed and knows the `rustlings watch` command.

Therefore I'd try to cover the following topics:
- what sorts of exercises will come later? (add functions, edit functions, edit tests, quizes etc. without going to deep into details)
- **how can I finish an exercise** (`I AM NOT DONE`)
- **structure of an exercise** (description / instructions in the header block, code below, TODO comments, hints in the code)
- **how can I find help** (README.md and hints)
- **reading and understanding compiler errors** (this is very imporant in my opinion)
- covered sections (also which sections are missing at the moment)
- rustlings commands, e. g. `rustlings list` to check your progress

Since most of the things I mentioned are more informative, probably a **quiz** where a user has to do **console inputs** (`Answer with (Y)es or (N)o`) or a selection (depends on the library; would offer 1 out of n anwers) could be an idea.
This way we could ensure a learner knows how rustlings work (if he/she is not guessing the answers of course) and this would probably improve the user experience and it is more likely that he/she finishes all exercises

apogeeoak:
A console quiz is an interesting way to introduce Rustlings but I do not think it flows with the rest of the exercises.

How would the console quiz be triggered, using the run command? What happens with the verify and watch commands?

I like the gradual way Rustlings currently introduces topics but I thought a successful compiling exercise would help reassure the user that everything is working correctly.

Zerotask:
> A console quiz is an interesting way to introduce Rustlings but I do not think it flows with the rest of the exercises.

The rest of the exercises are about learning Rust.  This Introduction is about learning # Rustlings, the learning Tool.

> How would the console quiz be triggered, using the run command? What happens with the verify and watch commands?

I think the best way of doing it is to separate it from the exercises (maybe with an own introduction/ folder). Otherwise it will result in pseudo-exercises with some work-arounds.

That way the exercises folder would only have actual Rust sections in it (introduction isn't a section)

crd477:
What about having a preliminary "exercise" that is already solved for everyone out of the box and just prints out something along the lines of "Congratulations on getting rustlings installed!  Welcome to the exercises.  This is the very first exercise and it has been solved for you.  From this point on, you will be required to fix any errors that keep the exercise solutions from compiling", etc.

This might prevent people from filing the recurring "I got a compilation error", e.g., #657.

Zerotask:
No, I don't like this idea. From a learning perspective it's important to ensure that the structure and the learning process is clear.
If you just have an exercise where you dont have to do anything, there is not much of learning here

I also installed it manually and then I run rustlings watch and it worked as expected.

If people don't know that they should use this command, that it could be highlighted in the readme.

crd477:
If some users are already missing the key point about fixing compilation errors when it's right in the README, I'm not optimistic that a separate introduction with serve them any better.

Zerotask:
An introduction would teach that and more, more like your suggestion would.
That's exactly the reason for an introduction.

apogeeoak:
@crd477 this pull request provides an already solved exercise that prints a welcome message. Adding information about needing to correct compiler errors in future exercises is a good idea.

@Zerotask the main purpose of the first exercise (the already solved exercise) in this pull request is to teach the I AM NOT DONE comment flow. I like how the Rustlings exercises only teach one thing at a time. A separate introduction command seems too heavy weight for a beginner course. This information is contained in the README. If the user does not read the README, how would they even know about the intro command?

Zerotask:
I'm not talking about a new command and I don't talk about something heavyweight.

It is always a good practice to introduce the learning process. That's how university learning materials work.

Also a Rustling exercise will often teach you more than one thing.
Just to teach the I AM NOT DONE thing isn't enough.
The Readme isn't complete and just gives an overview which doesn't cover the whole learning process.

apogeeoak:
If the intro is separate and does not follow the normal flow how is it invoked? I assumed it would need to use a separate command instead of the normal verify, watch, or run commands. And if it used the existing commands how would it differ from the regular exercises?

apogeeoak:
@Zerotask I like your introduction topics but I think it might be better to add that information directly to the README. New user will be able to find it without even having to download the source.

Zerotask:
It's always the beginning if you run rustlings watch.
That's the nature of an introduction.
It's like an own section, but it's about explaining and a quiz to ensure you understand it (but I'm repeating myself).

> And if it used the existing commands how would it differ from the regular exercises?

I already explained that.

Zerotask:
> New user will be able to find it without even having to download the source.

It doesn't make much sense to explain the learning process in detail if you don't download and want to try it.

This would just bloat the Readme file up and  in todays time many people just skip long texts. That's why I think an interactive way of explaining it is better.

But it seems you just want to "defend" your PR now.

apogeeoak:
Does the intro use the I AM NOT DONE mechanic as well? I am still trying to understand your idea.

Zerotask:
> Does the intro use the I AM NOT DONE mechanic as well?

It could be, but it doesn't have to. It depends on how this introduction will be designed.
But that's not important now. It will explain that the actual exercises will have this mechanic

> I am still trying to understand your idea.

What exactly is unclear? Maybe just read my first post again

apogeeoak:
I think having two separate folders (exercises and introduction) using the same command but different continuation mechanics would only serve to confuse users. I am confused myself. It would also add more logic the main program to separate out the introduction from the regular exercises.

diannasoreil (approved):

PR #735, opened 2021-04-25
docs(exercises): consistent exercise description

Zerotask:
To have a consistent exercise description block, every exercise will have the following structure:
```
// <filename.rs>
//
// description / instructions
//
// help line
```

This way every exercise has the same structure. Learners will know that the first line is the filename, the last is the help line and in between is a description or instructions.
At the moment, some exercises don't have a filename at the beginning and / or don't show help information (e. g. the hint for this exercise). Sometimes help information or links are in between the instructions. In some exercises instructions come after some lines of code, which is inconsistent and could confuse some people.
Additionally to the current hint, I also added an reference to the corresponding README.md, in case learners will overlook it.

### DIFF
--- a/exercises/clippy/clippy1.rs
+++ b/exercises/clippy/clippy1.rs
@@ -1,10 +1,12 @@
 // clippy1.rs
+//
 // The Clippy tool is a collection of lints to analyze your code
 // so you can catch common mistakes and improve your Rust code.
 //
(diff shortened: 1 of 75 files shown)


diannasoreil:
This is probably something that will be done directly on the main branch at some point, having it in a PR will just lead to a lot of merge conflicts.

PR #741, opened 2021-05-01
Add hints on how to get gcc installed

highb:
Just tried installing Rustlings using the `install.sh` script on a fresh Ubuntu install and I forgot to install `build-essential` so it failed once cargo started trying to compile. This commit adds a check to the script for `cc` as well as a note to `README.md`.

I'm not on an OSX system right now, but iirc `cc` is what the Xcode CLI tools uses, as well so it should work across both platforms. Not sure about Windows.

If you would prefer that I make a PR with only the `README.md` change, I'm happy to do that.


diannasoreil (approved):

diannasoreil:
@all-contributors please add @highb for doc

PR #742, opened 2021-05-04
iterators5: Change Progress to derive Clone and Copy

hyperparabolic:
Sorry in advance if this is just me being ignorant.  I'm pretty new to rust, and I'm still sorting out the ownership model.

I personally found it really unintuitive to not call `count_iterator` as a part of my `count_collection_iterator` implementation.

Rather than writing the problem solution here, I'll give an example with `count_for` and `count_collection_for`:
```
fn count_for(map: &HashMap<String, Progress>, value: Progress) -> usize {
    let mut count = 0;
    for val in map.values() {
        if val == &value {
            count += 1;
        }
    }
    count
}

fn count_collection_for(collection: &[HashMap<String, Progress>], value: Progress) -> usize {
    let mut count = 0;
    for map in collection {
        count = count + count_for(map, value);
    }
    count
}
```
It feels very natural to implement the iterators more similar to this, but this results in the following error:
```
! Compiling of exercises/standard_library_types/iterators5.rs failed! Please try again. Here's the output:
error[E0382]: use of moved value: `value`
  --> exercises/standard_library_types/iterators5.rs:45:40
   |
42 | fn count_collection_for(collection: &[HashMap<String, Progress>], value: Progress) -> usize {
   |                                                                   ----- move occurs because `value` has type `Progress`, which does not implement the `Copy` trait
...
45 |         count = count + count_for(map, value);
   |                                        ^^^^^ value moved here, in previous iteration of loop

For more information about this error, try `rustc --explain E0382`.
```

I spent some time trying to find a way to not move `value` also without duplicating code, and it led me to either implementing the Copy / Clone traits, or using Arc to pass a shared atomic value.

Both of these require changes outside of the functions, which is counter to the suggestions in the top level comment:
> Only the two iterator methods (count_iterator and count_collection_iterator) need to be modified.

If there is some other solution that I did not find that calls `count_iterator` from `count_collection_iterator` I would love to know it.  Otherwise I would like to suggest adding these traits to Progress to allow this solution.

Thanks.

### DIFF
--- a/exercises/standard_library_types/iterators5.rs
+++ b/exercises/standard_library_types/iterators5.rs
@@ -15,7 +15,7 @@
 
 use std::collections::HashMap;
 
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Clone, Copy)]
 enum Progress {
     None,
     Some,


apogeeoak:
Good catch. `Progress` should be `Copy`.
For non-`Copy` types, the value could be borrowed (`&T`) instead of using `Arc`.

Does implementing `count_collection_for` in terms of `count_for` make the exercise more intuitive? Should that also be changed?

hyperparabolic:
Okay.  Good to know.  That makes a lot of sense, but also would require refactoring outside the specified functions since you'd have to change how they're called.

I don't really have a problem with the original implementation.  If I implemented `count_collection_for` or `count_collection_iterator` in isolation I probably wouldn't split it into two functions, but with both of them being implemented in the same file it seems like a bit of an oversight to not make this implementation easy.

I'd also be willing to refactor all of the functions by reference if that's what is wanted here.  Just let me know.

apogeeoak:
Changing `Progress` to be `Copy` is the correct approach.

It was an oversight from building down from the `count_collection_iterator` function.

PhilMiller (approved):

apogeeoak:
Closes #763

fmoko:
superseded by #771

PR #746, opened 2021-05-06
feat: Add move_semantics5 exercise.

sateeshkumarb:
- Added an exercise for `move_semantics` which helps in improving the understanding of mutable references.
- Added an exercise for `option` which helps in understanding of usage of keyword _ref_

### DIFF
--- a/exercises/option/option3.rs
+++ b/exercises/option/option3.rs
@@ -1,0 +1,20 @@
+// option2.rs
+// Make me compile! Execute `rustlings hint option3` for hints
+
+// I AM NOT DONE
+
+struct Point {
+    x: i32,
+    y: i32,
+}
+
+fn main() {
+    let y: Option<Point> = Some(Point { x: 100, y: 200 });
+
+    match y {
+        Some(p) => println!("Co-ordinates are{},{} ", p.x, p.y),
+        _ => println!("no match"),
+    }
+    // Fix without deleting this line.
+    y;
+}
--- a/exercises/option/option3.rs
+++ b/exercises/option/option3.rs
@@ -12,9 +12,8 @@ fn main() {
     let y: Option<Point> = Some(Point { x: 100, y: 200 });
 
     match y {
-        Some(p) => println!("Co-ordinates are{},{} ", p.x, p.y),
+        Some(p) => println!("Co-ordinates are {},{} ", p.x, p.y),
         _ => println!("no match"),
     }
-    // Fix without deleting this line.
-    y;
+    y; // Fix without deleting this line.
 }
--- a/exercises/option/option3.rs
+++ b/exercises/option/option3.rs
@@ -1,4 +1,4 @@
-// option2.rs
+// option3.rs
 // Make me compile! Execute `rustlings hint option3` for hints
 
 // I AM NOT DONE
--- a/info.toml
+++ b/info.toml
@@ -210,6 +210,13 @@ So the end goal is to:
    - since we're not creating a new vec in `main` anymore, we need to create
      a new vec in `fill_vec`, similarly to the way we did in `main`"""
 
+[[exercises]]
+name = "move_semantics5"
+path = "exercises/move_semantics/move_semantics5.rs"
+mode = "compile"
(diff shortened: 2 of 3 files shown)


exercises/option/option3.rs
```diff
@@ -0,0 +1,20 @@
+// option2.rs
```
Zerotask, line 1:
```suggestion
// option3.rs
```

info.toml
```diff
@@ -578,6 +585,16 @@ For example: Some(Some(variable)) = variable2
 Also see Option::flatten
 """
 
+[[exercises]]
+name = "option3"
+path = "exercises/option/option3.rs"
+mode = "compile"
+hint = """
+The compiler says a partial move happened in the `match`
+statement. How can this be avoided ? The compiler shows the correction
```
Zerotask, line 594:
```suggestion
statement. How can this be avoided? The compiler shows the correction
```

info.toml
```diff
@@ -210,6 +210,13 @@ So the end goal is to:
    - since we're not creating a new vec in `main` anymore, we need to create
      a new vec in `fill_vec`, similarly to the way we did in `main`"""
 
+[[exercises]]
+name = "move_semantics5"
+path = "exercises/move_semantics/move_semantics5.rs"
+mode = "compile"
+hint = """Carefully reason about the range in which each mutable reference is in vogue. Does updating the value of referrent (x) immediately after the mutable reference is taken helps ? Read more about 'Mutable Referenes' in the book's section References and Borrowing': https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references."""
```
Zerotask, line 217: Can you please add line breaks here? Then it's consistent with the other hints.

exercises/option/option3.rs
sateeshkumarb: Fixed.

info.toml
sateeshkumarb: Fixed.

info.toml
sateeshkumarb: Done.

exercises/option/option3.rs
```diff
@@ -0,0 +1,20 @@
...
+}
+
+fn main() {
+    let y: Option<Point> = Some(Point { x: 100, y: 200 });
+
+    match y {
+        Some(p) => println!("Co-ordinates are{},{} ", p.x, p.y),
+        _ => println!("no match"),
+    }
+    // Fix without deleting this line.
```
fmoko, line 18: you can move this comment after line 19 so it's more obvious which line "this line" is referring to

exercises/option/option3.rs
```diff
@@ -0,0 +1,20 @@
...
+struct Point {
+    x: i32,
+    y: i32,
+}
+
+fn main() {
+    let y: Option<Point> = Some(Point { x: 100, y: 200 });
+
+    match y {
+        Some(p) => println!("Co-ordinates are{},{} ", p.x, p.y),
```
fmoko, line 15:
```suggestion
        Some(p) => println!("Co-ordinates are {},{} ", p.x, p.y),
```

exercises/option/option3.rs
sateeshkumarb: Moved the comment to be inline which I think makes the intent clear.

exercises/option/option3.rs
sateeshkumarb: Done.

exercises/option/option3.rs
fmoko: oh yeah sorry that's what I meant

fmoko (approved):

sateeshkumarb:
@fmoko Now that this PR is approved can you please merge this change. Or are there any additional steps I need to take for this PR to be merged. Thanks.

fmoko:
@all-contributors please add @sateeshkumarb for code and content

milesfrain:
My first instinct was to solve this with `&` instead of the `ref` keyword, which worked:
```rust
    match &y {
        Some(p) => println!("Co-ordinates are {},{} ", p.x, p.y),
```

I assumed this is the envisioned solution:
```rust
    match y {
        Some(ref p) => println!("Co-ordinates are {},{} ", p.x, p.y),
```

I'm still hazy on the difference between `&` vs `ref`, even after reading https://doc.rust-lang.org/std/keyword.ref.html#-vs-ref

Wondering if this exercise could be made more restrictive to force the `ref` solution.

sateeshkumarb:
@milesfrain  That is a good point and after your comment I did searched around and experimented but I couldn't come up with a way to make this exercise to be restrictive to force _ref_ solution. Thanks for pointing this out.

PR #747, opened 2021-05-08
feat(deps): added dependabot for weekly updates

Zerotask:


fmoko:
we don't have that many dependencies, and keeping them up to date isn't a priority high enough to warrant 1 PR/week, monthly would work better

fmoko:
oh, i was gonna just suggest changing the "weekly" to "monthly" (assuming dependabot works with that?)

PR #755, opened 2021-05-15
fix(try_from_into, from_str): hints for dyn Error

tlyu:
Add hints about how to return the correct type for functions that
return `Result<_, Box<dyn Error>`. Some feedback from Discord suggests
that people run into trouble with that.


fmoko (approved):

fmoko:
Thanks!

PR #762, opened 2021-05-23
 fix: move_semantics5 hints

tlyu:
Improve hints and comments for move_semantics5 based on repeated instances of people expressing confusion about this exercise on Discord.

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -216,8 +216,8 @@ path = "exercises/move_semantics/move_semantics5.rs"
 mode = "compile"
 hint = """
 Carefully reason about the range in which each mutable reference is in
-vogue. Does updating the value of referrent (x) immediately after the
-mutable reference is taken helps? Read more about 'Mutable Referenes'
+vogue. Does updating the value of referent (x) immediately after the
+mutable reference is taken helps? Read more about 'Mutable References'
 in the book's section References and Borrowing':
 https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references."""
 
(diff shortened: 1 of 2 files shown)


fmoko (approved):

dtaralla:
Personally, me and some colleagues had a lot of troubles on this exercise, even though we do understand the concept of overlapping scopes. What we missed was actually in the problem statement.

The comments are still confusing - to us, the "or changing any of the lines" part entailed that we couldn't reorder them either. The exercise seems more focused on "thinking outside of the box" and discover that "AHAH! But they didn't say we can't REORDER!" than actually learning something about overlapping scopes.
Could we maybe just say something like "Make me compile by ONLY reordering lines in main."?

diannasoriel:
@dtaralla Yes, that seems like a good way to phrase the instructions for this exercise, and a fun little challenge at the same time. Thanks for the suggestion, I'll commit this on `main`.

diannasoriel:
Fixed (again) in https://github.com/rust-lang/rustlings/commit/df25684cb79f8413915e00b5efef29369849cef1.

info.toml
```diff
@@ -216,10 +216,14 @@ path = "exercises/move_semantics/move_semantics5.rs"
 mode = "compile"
 hint = """
 Carefully reason about the range in which each mutable reference is in
-vogue. Does updating the value of referrent (x) immediately after the
-mutable reference is taken helps? Read more about 'Mutable Referenes'
+vogue. Does it help to update the value of referent (x) immediately after
```
hamirmahal, line 219: #1974 gets rid of `referent` altogether because it doesn't seem to help the explanation.

Issue #567, opened 2020-10-22
/exercises/threads/threads1.rs compile fail


DrYaling:
Arc DerefMut between thread is not safe.
Maybe we should change status to
 let status = Arc::new(Mutex::new(JobStatus { jobs_completed: 0 }));

AbdouSeck:
Hi @DrYaling the point of the exercise is to have you pick the right data types so that the exercise can compile and run. The exercise, by default, will not compile. Your changes should make it compile.

Please let us know if you have any other question or suggestion.

Thanks,
Abdou

PR #764, opened 2021-05-27
docs(exercises): Updated challenge instructions for threads1.rs

metalogica:
I've seen from a cursory glance on the repo that a lot of people struggle with this problem set because of the ambiguity in the instructions (See issues #287, #743, #567).

I'd like to recommend perhaps putting a TODO on the lines that need changing to stop people from falling down rabbit holes and pursuing completely different solution to the exercise that avoids the core pedagogical content of the problem, which is an understanding of how `Arc` is an atomic reference counting primitive that relies on `.lock()` and `.unwrap()` to solve shared state concurrency problems with the `Mutex` primitive.

I've also added comments in the instructions to highlight why it is that the solution is solved when the program prints the counter to the screen 6 times, as this was unclear and was distracting students from the core takeaway of the exercise.

### DIFF
--- a/exercises/threads/threads1.rs
+++ b/exercises/threads/threads1.rs
@@ -4,7 +4,13 @@
 // monitoring progress until 10 jobs are completed. Because of the difference between the
 // spawned threads' sleep time, and the waiting threads sleep time, when you see 6 lines
 // of "waiting..." and the program ends without timing out when running,
-// you've got it :)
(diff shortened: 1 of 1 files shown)


diannasoreil:
Hopefully this is somewhat indirectly resolved via the new threads exercise structure? Let me know if not.

PR #771, opened 2021-06-06
fix(iterators5): derive Clone, Copy

### LINKED ISSUE #763: Second part of standard_library_types/iterators5.rs can be confusing
When I reached this exercise, I ran into a wall of sorts on the second half (implementation of `count_collection_iterator`). This was because, from a functional programming perspective, I should be implementing it using the existing `count_iterator` function that I'd just done for the first part. This isn't possible, because the `Progress` enum doesn't include `Clone` or `Copy` in the derived traits. I had considered making the passing of `value` be by-reference in `count_iterator`, but that would mean changing the calls to that function in the tests, and that seemed to go against the spirit of the exercise as laid out in the prelude comment.

I was able to solve this by essentially duplicating the `count_iterator` code within `count_collection_iterator`, but this doesn't really "feel right" in a functional programming sense.

My suggestion is that the `derive` directive on line ~19 include `Clone` and `Copy`. I have attached my file from my workspace, which has both a working solution (with duplicated code) and a commented-out version of the solution that uses composition. (Note that the file has a *.txt suffix, as GitHub's issues editor would not attach a *.rs file.)
[iterators5.rs.txt](https://github.com/rust-lang/rustlings/files/6548551/iterators5.rs.txt)

Randy

tlyu:
To allow more flexibility in solutions, derive `Clone` and `Copy`
for `Progress`.

Fixes #763.

Also add a hint about conciseness by using some method other than `fold`.

### DIFF
--- a/exercises/standard_library_types/iterators5.rs
+++ b/exercises/standard_library_types/iterators5.rs
@@ -15,7 +15,7 @@
 
 use std::collections::HashMap;
 
-#[derive(PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 enum Progress {
     None,
     Some,
--- a/info.toml
+++ b/info.toml
@@ -788,7 +788,10 @@ test count_iterator.
 The collection variable in count_collection_iterator is a slice of HashMaps. It
 needs to be converted into an iterator in order to use the iterator methods.
 
-The fold method can be useful in the count_collection_iterator function."""
+The fold method can be useful in the count_collection_iterator function.
+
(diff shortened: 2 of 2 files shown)


apogeeoak:
Deriving `Clone, Copy` for `Progress` was also proposed in #742. It just has not be implemented yet. It normally takes a bit.

The hint to look for alternate ways of solving the problem is a good addition.

fmoko:
merging this one, but i'll give kudos to @hyperparabolic for coming up with it first

fmoko:
@all-contributors please add @hyperparabolic for code

PR #772, opened 2021-06-07
feature: improve error_handling exercises

### LINKED ISSUE #719: Bug: error_handling/result1.rs order is wrong
The error_handling section starts with *errors1.rs* and then stops after *errorsn.rs* and then the `rustlings watch` want to continue with *generics/generics1.rs* instead of doing the last error_handling exercise *result1.rs*

image

### LINKED ISSUE #756: error-handling difficulty jump; teach custom error types
Maybe we should add more error-handling exercises. This would both deal with a large difficulty jump in the existing exercise sequence, and introduce some error-handling concepts that seem to be considered best practices now.

errors3 introduces the idea how to use the `?` operator to propagate an error out of `main()`. errorsn introduces `dyn error::Error` in a roundabout way (you have to read the hints, or pay close attention to how the tests are written). I think this is maybe a large jump in difficulty? Especially because errorsn uses stuff like mock types that implement `io::BufRead`.

Might we want some exercises with intermediate difficulty, such as a simple introduction of `dyn error::Error` with simpler error types that don't involve `std::io`?

Also, I've been reading stuff like https://github.com/rust-lang/project-error-handling/blob/master/CHARTER.md and https://blog.burntsushi.net/rust-error-handling/ and realizing that we might want to introduce the idea of custom error types. Exercises like try_from_into use `Box<dyn error::Error>`, when they should probably use a custom error type instead of the trait object. This could be a bit challenging because of the need to implement multiple traits on the custom error type: `From`, `Error` (including `Display`), etc.

I'm willing to work on pull requests to implement these ideas, but I wanted to hear some feedback first.

tlyu:
Move result1 to errors4 and put it in the error handling section. Fixes #719.

Add new exercises errors5 and errors6, to introduce boxed errors and
custom error enums more gently. Delete errorsn, because it tried to do
too much too soon. Fixes #756.


exercises/error_handling/errors5.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+}
+
+// This is required so that `CreationError` can implement `error::Error`.
+impl fmt::Display for CreationError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let description = match *self {
+            CreationError::Negative => "Number is negative",
+            CreationError::Zero => "Number is zero",
```
yaahc, line 47:
```suggestion
            CreationError::Negative => "number is negative",
            CreationError::Zero => "number is zero",
```

Error messages should be lower case so they concatenate with other error messages nicely

exercises/error_handling/errors6.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+fn parse_pos_nonzero(s: &str)
+    -> Result<PositiveNonzeroInteger, ParsePosNonzeroError>
+{
+    // TODO: change this to return an appropriate error instead of panicking
+    // when `parse()` returns an error.
+    let x: i64 = s.parse().unwrap();
+    PositiveNonzeroInteger::new(x)
+        .or(Err(ParsePosNonzeroError::CreationError))
```
yaahc, line 27: This should use `map_err` and wrap the original error

That plus the above makes me think this belongs in that advanced error section you suggested

exercises/error_handling/errors6.rs
```diff
@@ -0,0 +1,86 @@
...
+// what to do next when our function returns an error.
+
+// Make these tests pass! Execute `rustlings hint errors6` for hints :)
+
+// I AM NOT DONE
+
+// This is a custom error type that we will be using in `parse_pos_nonzero()`.
+#[derive(PartialEq, Debug)]
+enum ParsePosNonzeroError {
+    CreationError,
```
yaahc, line 16: this should wrap the source error, also, the Error variant's shouldn't include `Error` in their identifier.

```suggestion
    Creation {
        source: CreationError,
    }
```

yaahc (changes requested):

tlyu:
Thanks for the review! I agree with most of your changes.

I was planning to have `map_err` in one of the advanced exercises, but now that we've talked about passing functions instead of closures to `map_err` (because closures aren't used yet at this point in the exercise sequence), I guess it could go in this exercise.

Is there a consensus for error enum variants with payloads? I think I've seen both 1-tuple structs and structs with named fields. Are there tradeoffs involved? (I suspect there are, involving public vs private visibility of fields, but I'm not sure.)

`std::num::ParseIntError` does have a payload, though it's a private one at the moment, so I guess we shouldn't wrap that?

yaahc:
> Is there a consensus for error enum variants with payloads? I think I've seen both 1-tuple structs and structs with named fields. Are there tradeoffs involved? (I suspect there are, involving public vs private visibility of fields, but I'm not sure.)

I don't think there is a clear consensus. There should be no difference in field visibility since all fields in enums are pub. The main difference is how the constructor is exposed. For a struct variant it uses the constructor syntax, but for unit variants the compiler essentially adds a function with the same name for the constructor. And I think I can see where you're going with this question. 

```rust
enum WrappingError {
    Inner(InnerError),
}

struct InnerError;

fn main() {
    Err::<(), _>(InnerError).map_err(WrappingError::Inner);
}
```

This is possible with tuple variants but not with struct variants, so it will probably be easiest to teach the tuple version since it provides the needed conversion function out of the box.

yaahc:
> `std::num::ParseIntError` does have a payload, though it's a private one at the moment, so I guess we shouldn't wrap that?

I don't understand what you mean. The ParseIntPayload is internal so we don't have to worry about it, but if anything the fact that they're not all identical and they have different runtime instantiations of that type means that we really need to wrap it, if we try to replace it with our own equivalent error type we will probably provide less information than the original error did.

tlyu:
> I don't understand what you mean. The ParseInt payload is internal so we don't have to worry about it, but if anything the fact that they're not all identical and they have different runtime instantiations of that type means that we really need to wrap it, if we try to replace it with our own equivalent error type we will probably provide less information than the original error did.

Yeah, I guess you're right. Even if we can't make decisions based on those private variants, they're still relevant for display purposes.

tlyu:
> This is possible with tuple variants but not with struct variants, so it will probably be easiest to teach the tuple version since it provides the needed conversion function out of the box.

I hadn't realized that tuple struct constructors actually were proper functions! I thought they were special struct expression syntax that happened to look like function calls. I'm not sure this exercise is the right place to introduce that concept. Or maybe it is? It seems likely to cause confusion if it hasn't been mentioned before.

yaahc:
> I hadn't realized that tuple struct constructors actually were proper functions! I thought they were special struct expression syntax that happened to look like function calls. I'm not sure this exercise is the right place to introduce that concept. Or maybe it is? It seems likely to cause confusion if it hasn't been mentioned before.

hehe, yea. This comes up when you're re-exporting items in a library. If you a type alias to bring it in it will only import the type, but not the function, where as if you export it with a `pub use` it will bring in all items with the same name from all 3 namespaces.

yaahc (commented):
lgtm

fmoko (approved):
i trust your judgement @yaahc

PR #773, opened 2021-06-07
fix(variables5): confine the answer further

kayuapi:
let mut number = 3; can lead to a correct answer, so the comment helps to direct the users to the intended answer.


fmoko (approved):

fmoko:
@all-contributors please add @kayuapi for content

PR #781, opened 2021-06-25
feature: advanced errors

tlyu:
Based on #780.

Add new exercises to demonstrate traits that are useful to help other code consume our custom error types.

For `advanced_errs2`, I'm somewhat concerned that the `source` method usage in the ignored test might be lead to confusing error messages about that method, so I'm OK with taking out that test (and the related challenge), possibly deferring it to another exercise.


diannasoriel (approved):
Fantastic, thank you!

PR #787, opened 2021-06-29
chore: Update hint of iterators3

kolbma:
`collect()` needs some hint for standard_library_types/iterators3 exercise with doc link for understanding different return types via `FromIterator`.

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -774,7 +774,7 @@ case is a vector of integers and the failure case is a DivisionError.
 
 The list_of_results function needs to return a vector of results.
 
-See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect how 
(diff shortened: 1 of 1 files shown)


info.toml
```diff
@@ -772,7 +772,10 @@ The division_results variable needs to be collected into a collection type.
 The result_with_list function needs to return a single Result where the success
 case is a vector of integers and the failure case is a DivisionError.
 
-The list_of_results function needs to return a vector of results."""
+The list_of_results function needs to return a vector of results.
+
+See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect how
```
fmoko, line 777:
```suggestion
See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect for how 
```

PR #794, opened 2021-07-05
chore: Updated quiz1.rs question description

sinharaksh1t:
The original text `order amount` appears to be confusing as it seems to imply that the function accepts an amount/price. However, the function is instead returning a price while the argument accepted into the function is the quantity of apples bought.


fmoko (approved):
Thanks!

fmoko:
@all-contributors please add @sinharaksh1t for content

PR #805, opened 2021-07-09
Error handling

suryapandian:
Rustlings - gives exercises with minor errors for us to fix and learn.

Originally the error I was getting was

```
the trait `From<CreationError>` is not implemented for `ParseIntError`
```

To fix this I tried:

```
impl From<CreationError> for ParseIntError{
    fn from(e: CreationError) -> ParseIntError{
        return ParseIntError::CreationError(e);
    }
}

impl From<ParseIntError> for CreationError{
	fn from(e: ParseIntError) -> CreationError{
		return CreationError::ParseIntError(e);
	}
}
```

This created more errors, so I tried
```
type Result<T, E> = std::result::Result<T, Box<dyn E>>;
```

This didn't help as well.

My guess is that I am making an error in the syntax of the boxed error. 🤔 

Any suggestion or lead would help.

I was referring:
https://doc.rust-lang.org/stable/rust-by-example/error/multiple_error_types/boxing_errors.html
https://stackoverflow.com/questions/59214210/rust-implement-trait-from-for-associated-type-error
https://stackoverflow.com/questions/52651566/associated-type-not-found-in-the-current-scope-when-returning-from-pattern-match

### DIFF
--- a/exercises/collections/hashmap1.rs
+++ b/exercises/collections/hashmap1.rs
@@ -11,18 +11,15 @@
 // Execute the command `rustlings hint hashmap1` if you need
 // hints.
 
-// I AM NOT DONE
-
 use std::collections::HashMap;
 
 fn fruit_basket() -> HashMap<String, u32> {
-    let mut basket = // TODO: declare your hash map here.
+    let mut basket = HashMap::new();
 
     // Two bananas are already given for you :)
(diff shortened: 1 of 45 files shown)


suryapandian:
Mistakenly created, meant to create it in my fork.

PR #807, opened 2021-07-21
Typo in tests3

BertrandBev:


exercises/tests/tests3.rs
```diff
@@ -1,5 +1,5 @@
 // tests3.rs
-// This test isn't testing our function -- make it do that in such a way that
+// This test isn't testing our function -- make it does that in such a way that
```
Anthras, line 2: "does" should be changed to "so"
"... make is so that in such a way ..."
or
"... make is so that the test passes. ..."

suryapandian (approved):

exercises/tests/tests3.rs
diannasoriel: I think "make it do so in such a way [...]" also works

suryapandian: well, the way the desc is right now itself seems correct. 🤷

BertrandBev: Woops you're right I must've misread it! Never mind! :)

Issue #816, opened 2021-08-06
conversions/from_str + try_from_into, struggle with Box<dyn error::Error>


jebos:
Hey,

thanks for all those lessons. Though I had some hard time to solve **from_str** and **try_from_into** due to the use of `Box<dyn error::Error>.`

It is totaly possible that I had the issue due to the lack of knowledge, but how are you supposed to return an custom Error - i.e. for an invalid slice size?

Of course on the parse you could add "?" that is all fine - but how to return an error i.e. if the slice is to small? 

I did it by adding this:

`use std::fmt;
use std::error::Error;

#[derive(Debug)]
struct MyError(String);

impl fmt::Display for MyError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "There is an error: {}", self.0)
    }
}

impl Error for MyError {}`

and then returning:

`return Err(Box::new(MyError("invalid".into())));`

However needing to do this seems totaly out of place for the lesson. What would be the expected way to solve those?

LeoniePhiline:
Hi Jeremias,

I did struggle with this one as well. Probably a hint or helpful inline comment would be good!

If you check out the first item at https://doc.rust-lang.org/stable/std/convert/trait.From.html#implementors you will be able to go further with far less typing overhead:

```rust
impl From<&'_ str> for Box<dyn Error>
```

So `Box<dyn Error>` implements `from(error_message: &str)`, which means you can just `return Err("My error message".into())`.
(Or more explicityly: `return Err(Self::Error::from("My error message")`, using the pre-defined alias `type Error = Box<dyn error::Error>`.)

See also 
- https://users.rust-lang.org/t/help-understanding-return-for-box-dyn-error/33748/4
- https://doc.rust-lang.org/stable/src/std/error.rs.html#303-324
- https://doc.rust-lang.org/stable/src/std/error.rs.html#282-301
- https://doc.rust-lang.org/stable/src/std/error.rs.html#239-280

Example for exercises/conversions/try_from_into.rs:

```rust
// Slice implementation
impl TryFrom<&[i16]> for Color {
    type Error = Box<dyn error::Error>;
    fn try_from(slice: &[i16]) -> Result<Self, Self::Error> {
        match slice.len() {
            3 => Ok(Color { 
                red: slice[0].try_into()?, 
                green: slice[1].try_into()?, 
                blue: slice[2].try_into()? 
            }),
[...]

PR #818, opened 2021-08-09
fix(try_from_into): Hint at `&str` to `Box<dyn std::error:Error>` conversion.

LeoniePhiline:
Proposal to improve UX at `try_from_into`.

Fixes https://github.com/rust-lang/rustlings/issues/816

Up for discussion:
- Is the hint too explicit?

### DIFF
--- a/exercises/conversions/try_from_into.rs
+++ b/exercises/conversions/try_from_into.rs
@@ -19,9 +19,14 @@ struct Color {
 // You need to create an implementation for a tuple of three integers,
 // an array of three integers and a slice of integers.
 //
-// Note that the implementation for tuple and array will be checked at compile time,
+// Note that correct RGB color values must be integers in the 0..=255 range.
+//
+// Also note that the implementation for tuple and array will be checked at compile time,
 // but the slice implementation needs to check the slice length!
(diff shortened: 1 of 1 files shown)


exercises/conversions/try_from_into.rs
```diff
@@ -19,9 +19,14 @@ struct Color {
 // You need to create an implementation for a tuple of three integers,
 // an array of three integers and a slice of integers.
 //
-// Note that the implementation for tuple and array will be checked at compile time,
+// Note that correct RGB color values must be integers in the 0..=255 range.
+//
+// Also note that the implementation for tuple and array will be checked at compile time,
 // but the slice implementation needs to check the slice length!
-// Also note that correct RGB color values must be integers in the 0..=255 range.
+// A little hint:
```
Randgalt, line 26: This is tripping me up right now. I hadn't come across this syntax in any of my Rust research and I don't know how to get past it. I'll try this "little hint" now but I think Rustlings should have a dedicated rustling prior to this so we can learn how to do this properly.

diannasoreil:
I think this is closed via https://github.com/rust-lang/rustlings/commit/81d25aecff01643221131743b48f6b63377eeb59

PR #824, opened 2021-08-22
Fixed hint for move semantics ex. 1 hint

nickytonline:
## Description

Fixes the move semantics exercise 1 hint mentioning a piece of code on an incorrect line number.

```
So `vec0` is being *moved* into the function `fill_vec` when we call it on
line 10,
```

is now

```
So `vec0` is being *moved* into the function `fill_vec` when we call it on
line 7,
```

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -171,7 +171,7 @@ path = "exercises/move_semantics/move_semantics2.rs"
 mode = "compile"
 hint = """
 So `vec0` is being *moved* into the function `fill_vec` when we call it on
-line 10, which means it gets dropped at the end of `fill_vec`, which means we
+line 7, which means it gets dropped at the end of `fill_vec`, which means we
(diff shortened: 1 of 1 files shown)


info.toml
```diff
@@ -771,7 +771,7 @@ case is a vector of integers and the failure case is a DivisionError.
 
 The list_of_results function needs to return a vector of results.
 
-See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect for how 
+See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect for how
```
nickytonline, line 774: Looks like some extraneous whitespace was removed when I edited the file on github.dev. I can remove the white space change here and below if you want to keep the PR to just the line number  change.

nickytonline:
Nevermind. I was looking at the wrong hint. 🙃

PR #826, opened 2021-08-22
Fix inconsistent wording

dbednar230:
The second test expects the function to return 80 when there is an order of 40 apples, but the current wording implies returning 40 will pass as well


suryapandian:
Nice catch!

suryapandian (approved):

diannasoriel (approved):
Thanks!

diannasoriel:
@all-contributors please add @dbednar230 for content

Issue #823, opened 2021-08-19
move_semantics2: mutable and immutable exist same time


ziyouwa:
Rule of Refernces 👍 
At any given time, you can have either one mutable reference or any number of immutable references.

but  this code compile success!  

```
fn main() {
    let mut vec0 = Vec::new();

    // Do not change the following line!
    println!("{} has length {} content `{:?}`", "vec0", vec0.len(), vec0);   // vec0  immutable borrow

    let  vec1 = fill_vec(&mut vec0);   //vec0  mutable borrow   ?!!
    vec1.push(88);

    println!("{} has length {} content `{:?}`", "vec1", vec1.len(), vec1);
}

fn fill_vec(vec: &mut Vec<i32>) -> &mut Vec<i32> {
    // let mut vec = vec.clone();

    vec.push(22);
    vec.push(44);
    vec.push(66);

    vec
}
```

FutureHet:
I want to work on this issue, can you please assign it to me.

FutureHet:
I created a pull request named, updated move_semantics2.rs.
Please check that and merge that...
If there is any update let me know.

zachs18:
> **At any given time**, you can have either one mutable reference or any number of immutable references.

The lifetimes of the two borrows of vec0 mentioned in comments do not overlap (the borrows do not exist at the same time), so the rule of references is not broken.

SkiFire13:
To expand on @zachs18's explanation, borrows are ended as soon as you no longer use them. The first immutable borrow of `vec0` is only used inside the `println!`, so after that there aren't borrows active anymore and you're free to create a mutable borrow of `vec0` when calling `fill_vec`

PR #827, opened 2021-08-22
Updated move_semantics2.rs

FutureHet:
In the previous code, the address which was stored in vec was passed, so it was showing immutable but instead of we can pass the address of vec and can clone that array into another one.

### DIFF
--- a/exercises/move_semantics/move_semantics2.rs
+++ b/exercises/move_semantics/move_semantics2.rs
@@ -7,7 +7,7 @@
 fn main() {
     let vec0 = Vec::new();
 
-    let mut vec1 = fill_vec(vec0);
+    let mut vec1 = fill_vec(&vec0);
 
     // Do not change the following line!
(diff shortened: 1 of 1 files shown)


SkiFire13:
I don't think this is right. The code is supposed to not compile, the changes you're proposing are what the user has to figure out.

diannasoriel:
Yep, this fixes a non-issue.

FutureHet:
When I tried to compile the default file, there were some errors, it was not compiling.

SkiFire13:
That's supposed to happen, the aim of this repository is helping you (and other people who also clone this repository) learn rust by fixing these errors.

FutureHet:
okay, so I was not able to compile then it means that issue is closed, right?

FutureHet:
Issue no: #823

diannasoriel:
@FutureHet Yes, I think the issue you're linking to is more of a question regarding how a feature of Rust works, not an issue with the actual Rustlings program.

Issue #810, opened 2021-07-29
Can you please add numbers to the exercises folders so that they come in the same order as in rustlings watch?


TimothyAlexisVass:
I think it would be helpful to get started easily.

suryapandian:
This is a nice suggestion.

You could do that yourself by making the necessary change and raising a pull request.
If the repository owners feel good about the change then it would go into rustlings :)

anuk909:
It will make adding exercises folder in the middle very annoying.

diannasoriel:
Personally, I think this would to more harm than good. There's already suffix numbers (e.g. `variables1`), so I fear adding more numbers in the file path would make it more confusing to parse. That's my opinion, though, what do you think @AbdouSeck @jrvidal?

VMois:
@diannasoriel All data and hints are defined in [info.toml](https://github.com/rust-lang/rustlings/blob/main/info.toml). So adding support to folder numbering will not make parsing more confusing.

@anuk909 yep, it is true. You will need to update all folder numbers after the new folder is inserted.

diannasoriel:
@VMois Sorry, I meant more confusing to parse visually, as someone who runs through these exercises.

Svintooo:
What about just give the ability to list the folders in the correct order, without adding numbers to them?
This way there will not be any hassle adding more exercises in-between existing ones.

As a suggestion, it could be a sub-command to the rustlings binary. And it could maybe also show the solution status for each folder.
```bash
$ rustlings list
intro/       [ PASS ]
variables/   [ FAIL ]  # Unit tests exists which fails
functions/   [ERRORS]  # Compilation errors exists
...
```

As of now you can get the correct order of the folders (and quizzes) with the following command:
```bash
$ grep -Po '(?<=path = "exercises/)[^/]+(/|\.rs)' info.toml | uniq
```

DC4JG:
I just encountered this problem.

Started the course via `rustlings watch` then had to search for the file I should edit. Luckily the `hint intro1` gave it away. For all following exercises I relied on VSCodes ctrl+click feature on the file path in the compiler message to open the relevant file without searching.

Since there are 20+ subdirectories and potentially more to come I suggest some way to clarify this for new users, especially for the very first intro without compiler error explicitly stating the file path. Numbering the exercises would be an easy solution but as mentioned above maintainability might suffer. Another issue to keep in mind is that not all operating systems sort folders consistently.

[...]

PR #841, opened 2021-09-09
Add numbers to exercises folders names according to watch order

VMois:
Closes https://github.com/rust-lang/rustlings/issues/810


VMois:
I just realized that the order of the folders doesn't look good. I can update "1" => "01", etc.

But before I would like to get feedback from maintainers. Will they at all accept numbering in folders. 

Thank you

diannasoreil:
This PR will be pretty out of date by the time 5.0 releases, so feel free to reopen it at some point if this is still a concern to you.

Issue #842, opened 2021-09-13
no way to quit `rustlings watch`


abusch:
I ran `rustlings watch` and started working my way through the exercices and wanted to stop for a bit, only to find out there was absolutely no way to quit the process (except by finishing all the exercises...). I tried `C-c`, `C-\`, `C-d`, `quit`, `exit`, ...

It would be handy to add a `quit` command or similar (and maybe a `help` one too).

RusimbiP:
I use CTRL + C and it works

abusch:
hmm, it might be a `fish` shell thing maybe...

diannasoriel:
A quit command would be a good idea, yeah.

PR #843, opened 2021-09-14
feat: Add "quit" command to `rustlings watch`

abusch:
closes: #842

Took a crack at adding a "quit" command to the `rustlings watch` prompt, as at the moment the only way to exit that shell is to finish all the exercises in one go.

The implementation is a bit clunky... Ideally I would have another channel to signal the intention to quit from the thread spawned by `spawn_watch_shell()`, and the `watch` function would `select` on both receivers, but I don't think there's a way to do that with `std` channels. We can't move to `crossbeam` channels either as the `notify` crate has a hard dependency on `std` channels :( (this is fixed in the upcoming version 5, but it hasn't been released yet...).

I probably should have waited for some discussion on the issue, but it seemed easy enough to give it a shot anyway... Hope this can be useful.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,8 @@ use std::fs;
 use std::io::{self, prelude::*};
 use std::path::Path;
 use std::process::{Command, Stdio};
-use std::sync::mpsc::channel;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc::{RecvTimeoutError, channel};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -217,65 +218,76 @@ fn main() {
...
+                        "If you noticed any issues, please don't hesitate to report them to our repo."
+                    );
+                    println!("You can also contribute your own exercises to help the greater community!");
+                    println!();
+                    println!("Before reporting an issue or contributing, please read our guidelines:");
+                    println!("https://github.com/rust-lang/rustlings/blob/main/CONTRIBUTING.md");
+                }
+                Ok(WatchStatus::Unfinished) => {
+                    println!("We hope you're enjoying learning about Rust!");
+                    println!("If you want to continue working on the exercises at a later point, you can simply run `rustlings watch` again");
```
abusch, line 267: More than happy to get better copy here :)

diannasoriel:
@abusch Oops, totally accidentally superseded part of this PR with mine (#846). Would you be up for converting this PR into one that only adds the quit-after-finish functionality? That part looks good, but it'll lead to merge conflicts if we both add the quit command.

abusch:
@diannasoriel ha, no worries :)
I see your PR just got merged so I'll rebase mine and see how it goes.

diannasoriel:
@abusch Thank you! Looks good.

diannasoriel:
@all-contributors please add @abusch for code

PR #854, opened 2021-09-29
Does not follow the function naming convention snake_case

xuesongbj:
This exercise does not follow the snake_case specification, cargo test has the following warning:

warning: function `is_FooBar` should have a snake case name.

To ensure the meaning is correct, try to add allow non_snake_case.


diannasoriel:
In this case it'd be okay to change the offending function names too, since it's against convention and may confuse people that encounter it.

xuesongbj:
is_FooBar rename is_foo_bar ? @diannasoriel

xuesongbj:
If the problem is not fixed, there is a warning message

diannasoriel:
> is_FooBar rename is_foo_bar ?

Yes, that sounds right.

solariens:
Yes, I also encountered this warning problem. when will this commit merge into master branch?  @diannasoriel

diannasoriel:
Again, I would rather fix the problem properly than rely on suppressing the warning. Therefore, changing the non-snake-case function names to snake-case names would be the correct solution here.

xuesongbj:
Has been modified according to snake-case:

[cfdae46](https://github.com/rust-lang/rustlings/pull/854/commits/cfdae46eed63cac359c68f6b6985fd30a059b8d1)

diannasoriel:
@all-contributors please add @xuesongbj for content

PR #865, opened 2021-10-18
fix(advanced_errors): Fix unit test unreachable_patterns warning

xuesongbj:
After the code is perfected, the following warnings will appear when the unit test is executed:

```
  --> src/main.rs:72:13
   |
72 |             _ => write!(f, "unhandled error!"),
   |             ^
   |
   = note: `#[warn(unreachable_patterns)]` on by default
```

It is recommended to delete this line of code.


diannasoriel:
This should already be merged as part of #854 :)

PR #868, opened 2021-10-26
feat: Add exercises for understanding borrowing

jagill:
I recently ran a "learn Rust" session with some coworkers using Rustling.  Rustling is great!  But it doesn't talk about one of Rust's most iconic features, borrowing.  Here are a couple simple exercises to explore immutable borrowing, mutable borrowing, and "aliasing xor mutability".


diannasoreil:
move_semantics exercises already do most of this, I feel like there's too much overlap. If you feel like there's something missing from those, feel free to adjust it in a separate PR.

PR #877, opened 2021-11-13
error[E0425]: cannot find value `x` in this scope   --> exercises/variables/variables1.rs:12:5

chicagoist:


diannasoreil:
That is part of the solution of the exercise, but the point is for people to figure that out themselves 😄

chicagoist:
> 

Thanks! When I understood it, it was already late. I apologize for anxiety.

PR #882, opened 2021-12-02
refactor(exercises): shifts smart pointer exercises into new folder

### LINKED ISSUE #813: arc1 way too difficult for progression
Hi,

I'm going through the exercises, and it seems that `arc1.rs` is a bit out of place. It doesn't relate to any other exercises in `standard_library_types`, and the solver probably has to read most of the concurrency chapter before being able to do it. Perhaps a better home for this would be in `threads`, and we could add more exercises about concurrency?

seaerchin:
# What's changed
Previously, the exercises for smart pointers (`Box` and `Arc`) were grouped together under standard library types. This led to an unexpected jump in difficulty and divergence from the rust book, where the chapter for standard library types mainly covered iterators and closures. 

The relevant content is under smart pointers (for the book) and hence, the relevant exercises have been shifted into the newly created `smart_pointers` folder and `info.toml` + `exercises/README.md` updated to reflect this change

closes #813


rlpowell:
This doen't quite match either; the book has iterators, then Box and Rc and RefCell in the smart pointers chapter, but Arc is in yet another chapter.  Arc does not belong in the smart_pointers directory, only Box.

PR #893, opened 2021-12-23
feat: Adding threads1.rs with a focus on JoinHandles

### LINKED ISSUE #892: threads1.rs breakdown suggestion
I recently finished the rustlings course at version 4.6.0. I noted in the hint for threads1.rs looked like a cry for help: _Ok, so, real talk, this was actually tricky for *me* to do too. And I could see a lot of different problems you might run into, so at this
point I'm not sure which one you've hit :)_ I also had trouble implementing the exercise. I borrowed heavily from the last section of the book on threads. It involved adding the Mutex, moving the 'status.clone()' into the loop, and then unwrapping the mutex to check for if the number of threads has been completed. It was a lot of edits for the single exercise. It seemed odd to check the value to see if the threads had finished.

My recommendation, is to split threads1.rs into two exercises
1. threads1.rs would focus on using the JoinHandle result from spawning 10 threads and waiting the threads to finish.
2. threads2.rs starts with the result from threads1.rs. It then stubs out the Arc and hints for the Mutex implementation to change shared data. Ask for the final value held by the mutex at the end of the exercise.

jaystile:
feat: Adding threads1.rs with a focus on JoinHandles and waiting for spawned threads to finish. Moved the original threads1.rs to threads2.rs with the focus on the Mutex and modifying shared data.

resolves #892

### DIFF
--- a/exercises/threads/threads1.rs
+++ b/exercises/threads/threads1.rs
@@ -1,32 +1,31 @@
 // threads1.rs
-// Make this compile! Execute `rustlings hint threads1` for hints :)
-// The idea is the thread spawned on line 22 is completing jobs while the main thread is
-// monitoring progress until 10 jobs are completed. Because of the difference between the
-// spawned threads' sleep time, and the waiting threads sleep time, when you see 6 lines
-// of "waiting..." and the program ends without timing out when running,
-// you've got it :)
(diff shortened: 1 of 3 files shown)


diannasoreil:
@all-contributors please add @jaystile for content

PR #899, opened 2021-12-29
fix(quiz1): update to say quiz covers "If"

tan-zx:
The correct order for doing this quiz should be after "Variables", "Functions", and "If", the last of which is not mentioned in the comments.


diannasoreil (approved):
Thank you!

diannasoreil:
@all-contributors please add @tan-zx for content

PR #902, opened 2021-12-31
fix(structs3): Add panic! statement into structs3

### LINKED ISSUE #685: Structs3: relies on panic! which is not introduced in the official book until chapter 9
The exercise of Structs3 requires knowledge of `panic!` to complete the task.

```rust
    fn new(sender_country: String, recipient_country: String, weight_in_grams: i32) -> Package {
        if weight_in_grams <= 0 {
            // Something goes here...
        } else {
            return Package {
                sender_country,
                recipient_country,
                weight_in_grams,
            };
        }
    }
```

However `panic!` is not introduced in the official book until chapter 9 (Error Handling), so user who closely follows the book and does exercise will have additional challenge to find the right answer here.

Maybe we should just provide the `panic!` implementation here instead of asking user to find it?

camperdue42:
Adds panic! statement into Structs3 instead of requiring users to implement themselves. New users following along with Rust Book likely will not be exposed to panic! statements as discussed in issue #685.

closes #685

### DIFF
--- a/exercises/structs/structs3.rs
+++ b/exercises/structs/structs3.rs
@@ -15,7 +15,7 @@ struct Package {
 impl Package {
     fn new(sender_country: String, recipient_country: String, weight_in_grams: i32) -> Package {
         if weight_in_grams <= 0 {
-            // Something goes here...
+            panic!("Can not ship a weightless package.")
         } else {
             return Package {sender_country, recipient_country, weight_in_grams};
         }


diannasoreil:
Merged into `5.0-dev`: https://github.com/rust-lang/rustlings/commit/4531c21bf1b54ff5fa8cc796584043b826cf78e8

diannasoreil:
@all-contributors please add @camperdue42 for content

PR #903, opened 2022-01-01
refactor: Changed function name to be more accurate

neilpate:
A new programmer looking at the method name might assume that the purpose of the function is to return just the price of a single apple rather than the total price taking the discount into effect.

### DIFF
--- a/exercises/quiz1.rs
+++ b/exercises/quiz1.rs
@@ -11,14 +11,14 @@
 // I AM NOT DONE
 
 // Put your function here!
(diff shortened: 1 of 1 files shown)


diannasoreil:
Merged in `5.0-dev`: https://github.com/rust-lang/rustlings/commit/c265b681b188ea21b3f8585e65ea363fc02c4b50

diannasoreil:
@all-contributors please add @neilpate for content

PR #908, opened 2022-01-08
feat: Add move_semantics6.rs exercise (importance of & in arguments of function)

Kallu-A:
Could help to learn the difference between ownership and borrowed value in arguments of function and why use one or the other depending of the situation.

### DIFF
--- a/exercises/move_semantics/move_semantics6.rs
+++ b/exercises/move_semantics/move_semantics6.rs
@@ -6,8 +6,8 @@
 
 fn main() {
     let data = "Rust is great!".to_string();
-    println!("{}", get_char(data));
 
+    get_char(data);
 
     string_uppercase(&data);
 }
@@ -20,5 +20,6 @@ fn get_char(data: String) -> char {
 // Should take ownership
 fn string_uppercase(mut data: &String) {
     data = &data.to_uppercase();
+
     println!("{}", data);
 }
--- a/exercises/move_semantics/move_semantics6.rs
+++ b/exercises/move_semantics/move_semantics6.rs
@@ -1,0 +1,24 @@
+// move_semantics6.rs
+// Make me compile! `rustlings hint move_semantics6` for hints
+// You can't change anything except adding or removing reference
+
+// I AM NOT DONE
+
+fn main() {
+    let data = "Rust is great!".to_string();
+    println!("{}", get_char(data));
+
+
+    string_uppercase(&data);
+}
+
+// Should not take ownership
+fn get_char(data: String) -> char {
+    data.chars().last().unwrap()
+}
+
+// Should take ownership
+fn string_uppercase(mut data: &String) {
+    data = &data.to_uppercase();
+    println!("{}", data);
+}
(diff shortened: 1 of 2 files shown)


exercises/move_semantics/move_semantics6.rs
```diff
@@ -0,0 +1,24 @@
+// move_semantics6.rs
+// Make me compile! `rustlings hint move_semantics6` for hints
+// You can't change anything except adding or removing reference
```
diannasoreil, line 3:
```suggestion
// You can't change anything except adding or removing references
```

exercises/move_semantics/move_semantics6.rs
```diff
@@ -0,0 +1,24 @@
...
+
+// I AM NOT DONE
+
+fn main() {
+    let data = "Rust is great!".to_string();
+    println!("{}", get_char(data));
+
```
diannasoreil, line 10: This line should be deleted, for consistency's sake

exercises/move_semantics/move_semantics6.rs
```diff
@@ -0,0 +1,24 @@
...
+// Should not take ownership
+fn get_char(data: String) -> char {
+    data.chars().last().unwrap()
+}
+
+// Should take ownership
+fn string_uppercase(mut data: &String) {
+    data = &data.to_uppercase();
+    println!("{}", data);
+}
```
diannasoreil, line 24: Can you add a UNIX-style extra line ending here?

info.toml
```diff
@@ -221,6 +221,20 @@ in the book's section References and Borrowing':
 https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references.
 """
 
+[[exercises]]
+name = "move_semantics6"
+path = "exercises/move_semantics/move_semantics6.rs"
+mode = "compile"
+hint = """
+Here if you want to find the answer by yourself, you can read in the book's section References and Borrowing:
```
diannasoreil, line 229:
```suggestion
To find the answer, you can consult the book section "References and Borrowing":
```

info.toml
```diff
@@ -221,6 +221,20 @@ in the book's section References and Borrowing':
...
+https://doc.rust-lang.org/stable/book/ch04-02-references-and-borrowing.html
+Else the first problem is that `get_char` is taking ownership of the string
```
diannasoreil, line 231:
```suggestion
The first problem is that `get_char` is taking ownership of the string.
```

info.toml
```diff
@@ -221,6 +221,20 @@ in the book's section References and Borrowing':
...
+So `data` is moved and can't be used for `string_uppercase`
```
diannasoreil, line 232: I would rewrite everything after this to not explicitly tell the solution to the user, e.g.:

`data` is moved to `get_char` first, meaning that `string_uppercase` cannot manipulate the data. Once you've fixed that, `string_uppercase`'s function signature will also need to be adjusted - can you figure out how? Another hint: it has to do with the `&` character.

Kallu-A: At first, I was looking to do something like that, but I found it very frustrating that you didn't have a solution to the exercise with the best methods to solve it. So I gave it in the hints.

Maybe it could be a feature.
Because with the correct answer, you could see what you did wrong or could upgrade, and it could work like that? 
`rustlings solve <exercise>` and it's print the answer.

PR #920, opened 2022-01-23
feat(cli): adding colors in list command

vdusart:
Displays the list of exercises in color


diannasoreil:
I don't think that this is neccessary.

vdusart:
@diannasoreil You choose, but I personally think that having color in the list is much nicer. After it's just a bonus feature and really not necessary :smile:

PR #922, opened 2022-01-25
feat: Add flake.nix for nix users

### LINKED ISSUE #683: Add Nix (flake) support
It would be great to be able to build `rustlings` with nix and have it automatically available in a local shell. This will allow new users to start working on rustlings with nothing but the Nix package manager installed on their system without having to worry about trying to install rustlings.

It also exposes the rustlings program as a flake itself so that it can be imported and used elsewhere.

I've already done this on my fork of the repository, which you can find [here](https://github.com/jakeisnt/rustlings-nix), but as per the contributing guidelines for new features, I'm opening an issue to discuss!

MoritzBoehme:
This pull request is an attempt to close #683 by adding flake.nix and adding a separate section in the README.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,53 @@
+{
+  description = "Small exercises to get you used to reading and writing Rust code";
+
+  inputs = {
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, flake-utils, nixpkgs, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
+        rustlings =
+          pkgs.rustPlatform.buildRustPackage {
+            name = "rustlings";
+            version = "5.2.1";
+
+            src = with pkgs.lib; cleanSourceWith {
+              src = self;
+              # a function that returns a bool determining if the path should be included in the cleaned source
+              filter = path: type:
+                let
+                  # filename
+                  baseName = builtins.baseNameOf (toString path);
+                  # path from root directory
+                  path' = builtins.replaceStrings [ "${self}/" ] [ "" ] path;
+                  # checks if path is in the directory
+                  inDirectory = directory: hasPrefix directory path';
+                in
+                inDirectory "src" ||
+                inDirectory "tests" ||
+                hasPrefix "Cargo" baseName ||
+                baseName == "info.toml";
+            };
+
+            cargoLock.lockFile = ./Cargo.lock;
+          };
+      in
+      {
+        devShell = pkgs.mkShell {
+          buildInputs = with pkgs; [
+            cargo
+            rustc
+            rust-analyzer
+            rustlings
+          ];
+        };
+      });
+}
--- a/README.md
+++ b/README.md
@@ -28,6 +28,19 @@ curl -L https://raw.githubusercontent.com/rust-lang/rustlings/main/install.sh | 
 
 This will install Rustlings and give you access to the `rustlings` command. Run it to get started!
 
+### Nix
+Basically: Clone the repository at the latest tag, finally run `nix develop` or `nix-shell`.
+
+```bash
+# find out the latest version at https://github.com/rust-lang/rustlings/releases/latest (on edit 5.2.1)
+git clone -b 5.2.1 --depth 1 https://github.com/rust-lang/rustlings
+cd rustlings
+# if nix version > 2.3
+nix develop
+# if nix version <= 2.3
+nix-shell
+```
+
 ## Windows
 
 In PowerShell (Run as Administrator), set `ExecutionPolicy` to `RemoteSigned`:
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,60 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1650374568,
+        "narHash": "sha256-Z+s0J8/r907g149rllvwhb4pKi8Wam5ij0st8PwAh+E=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "b4a34015c698c7793d592d66adbab377907a2be8",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1659877975,
+        "narHash": "sha256-zllb8aq3YO3h8B/U0/J1WBgAL8EX5yWf5pMj3G0NAmc=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "c0e246b9b83f637f4681389ecabcb2681b4f3af0",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1666629043,
+        "narHash": "sha256-Yoq6Ut2F3Ol73yO9hG93x6ts5c4F5BhKTbcF3DtBEAw=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "b39fd6e4edef83cb4a135ebef98751ce23becc33",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,6 @@
+(import (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+in fetchTarball {
+  url =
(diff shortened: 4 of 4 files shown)


jdbosser:
I would love if this pr got merged! I hope it does.

diannasoreil:
Please help me get up to speed here - running `nix develop` here would provide you with a shell where Rust, Cargo, etc are all installed, as well as the `rustlings` binary?

winterqt:
@diannasoreil That's correct, yeah.

I do believe there are some changes that can be made here to make things better (using fenix over rust-overlay for a smaller download size, as well as using a build system that doesn't require updating the Cargo vendor hash).

@MoritzBoehme If you'd like to make these changes (I'm happy to review them if you do), feel free. If not, I'd also be happy to open a PR with these changes.

MoritzBoehme:
Sorry for the delayed answer, I'm currently a bit preoccupied due to my exams coming up.
Sadly, I have not yet experimented as much as I would like with Nix + Rust. So please feel free to open a PR either on my fork or a separate one in this repo to  improve the flake.

winterqt:
> Okay, I tried my best using fenix + naersk as per the fenix examples. This also solves the issue of having to specify cargo vendor hash, as you mentioned.

I somehow completely forgot that Nixpkgs' Rust build infrastructure can use a Cargo lock file (simply add `cargoLock.lockFile = ./Cargo.lock` to the `buildRustPackage` arguments) when writing that comment, do you mind switching back to it? Happy to review afterwards. Very sorry about this, and also about the late response.

MoritzBoehme:
@winterqt Could you have another look? Thanks in advance :)

winterqt (changes requested):
Sorry for forgetting about this again.

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+      flake = false;
+    };
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, flake-compat, flake-utils, nixpkgs }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
```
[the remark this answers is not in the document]
MoritzBoehme: Yeah, I also see the latter pattern with `legacyPackages` used more often, but think the former is cleaner. However, if you're strongly against keeping this as is, I can change it.

MoritzBoehme:
> Sorry for forgetting about this again.

No problem. Thanks for the help =)

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+        rustlings = pkgs.rustPlatform.buildRustPackage {
+          name = "rustlings";
+          version = "5.0.0";
+
+          src = ./.;
```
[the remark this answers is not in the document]
MoritzBoehme: Thanks :+1:

flake.nix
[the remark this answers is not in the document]
MoritzBoehme: @winterqt what do you say?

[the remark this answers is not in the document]
winterqt: It's bad practice to use it, see [here](https://zimbatm.com/notes/1000-instances-of-nixpkgs) (tl;dr: evaluations can be cached across flakes).

I don't think it *really* matters here since we don't depend on any flakes that we pass Nixpkgs to, but other flakes *may* depend on us for whatever reason, so maybe it'd be good to just go with this? Not sure, what do you think?

winterqt:
Also: I screwed up the initial source cleaning implementation, here's a fixed one:
```nix
          src = with pkgs.lib; cleanSourceWith {
            src = self;
            filter = path: type:
              let
                directories = [ "src" "tests" ];
                files = [ "Cargo.toml" "Cargo.lock" ];
                basePath = "${self}/";
                path' = builtins.replaceStrings [ basePath ] [ "" ] path;
              in
              (type == "directory" && any (dir: path' == dir) directories) ||
              (type == "regular" && (any (f: builtins.baseNameOf path' == f) files) || (any (dir: hasPrefix "${dir}/" path') directories));
          };
```

Let me know if you can think of any way to improve this code, not really sure if there's a better way to do this 😅

flake.nix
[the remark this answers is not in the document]
MoritzBoehme: Oh I see, in this case I will change this just in case. I always thought is was a stylist choice...

MoritzBoehme:
> Also: I screwed up the initial source cleaning implementation, here's a fixed one:
> 
> ```nix
>           src = with pkgs.lib; cleanSourceWith {
>             src = self;
>             filter = path: type:
>               let
>                 directories = [ "src" "tests" ];
>                 files = [ "Cargo.toml" "Cargo.lock" ];
>                 basePath = "${self}/";
>                 path' = builtins.replaceStrings [ basePath ] [ "" ] path;
>               in
>               (type == "directory" && any (dir: path' == dir) directories) ||
>               (type == "regular" && (any (f: builtins.baseNameOf path' == f) files) || (any (dir: hasPrefix "${dir}/" path') directories));
>           };
> ```
> 
> Let me know if you can think of any way to improve this code, not really sure if there's a better way to do this 😅

I have recently found this project https://github.com/numtide/nix-filter
I'm not sure if it's any good but maybe it's worth a look.

MoritzBoehme:
I have simplified the source cleaning a little bit, but unfortunately by removing the exercises directory from source the first integration test is now failing. I created a patch to remove this particular test, which seems to have worked and results in a rustlings binary which does not need to be rebuilt if we change something in the exercises.
If you have a better idea of how to fix the integration test, please let me know, because I don't really like my current solution...

MoritzBoehme:
Okay, so I found the error which caused the test to fail. Apparently, rustlings depends on both `default_out.txt` and `info.toml`.  With both of them present, the failing test from before now succeeds :tada: .

MoritzBoehme:
@winterqt if you should find yourself with some free time on your hands I'd appreciate your view on the state of the merge request :)

flake.nix
```diff
@@ -0,0 +1,52 @@
+{
+  description = "Rustlings devShell.";
```
winterqt, line 2: Might be better to use the description in this repo.

```suggestion
  description = "Small exercises to get you used to reading and writing Rust code";
```

flake.nix
```diff
@@ -0,0 +1,52 @@
...
+  inputs = {
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, flake-compat, flake-utils, nixpkgs }:
```
winterqt, line 13: `flake-compat` isn't used here, so IMO it's cleaner to not reference it? Up to you, though.

```suggestion
  outputs = { self, flake-utils, nixpkgs, ... }:
```

flake.nix
```diff
@@ -0,0 +1,52 @@
...
+                baseName = builtins.baseNameOf (toString path);
+                # path from root directory
+                path' = builtins.replaceStrings [ "${self}/" ] [ "" ] path;
+                # checks if path is in the directory
+                inDirectory = directory: hasPrefix directory path';
+              in
+              inDirectory "src" ||
+              inDirectory "tests" ||
+              hasPrefix "Cargo" baseName ||
+              baseName == "default_out.txt" ||
```
winterqt, line 36: What is this path? I don't see it mentioned anywhere in the source.

winterqt (commented):
Apologies for taking so long to get back to you about this.

Really great work cleaning up my code!

Last nit in addition to my other comments: if you want to add me as a co-author like you did (which I greatly appreciate, but not sure how much credit I deserve at this point 😅), would you mind switching it to `Winter <winter@winter.cafe>`, which is in line with how I commit everywhere?

Thanks, and my sincerest of apologies, again.

MoritzBoehme:
> Last nit in addition to my other comments: if you want to add me as a co-author like you did (which I greatly appreciate, but not sure how much credit I deserve at this point sweat_smile), would you mind switching it to `Winter <winter@winter.cafe>`, which is in line with how I commit everywhere?

Oh yeah, my bad I'll change it right away...

flake.nix
MoritzBoehme: Sorry, but what path do you mean?

flake.nix
MoritzBoehme: Good call!

flake.nix
MoritzBoehme: :+1:

MoritzBoehme:
I have now also added fenix back to the flake, so you do not need to set the RUST_SRC_PATH variable to use rust-analyzer. This was the only way I was able to use rust-analyzer in emacs for me. I guess other editors will face the same problem.

winterqt:
> For this, I have added fenix back to the flake, so you do not need to set the `RUST_SRC_PATH` variable to use rust-analyzer.

This is [already set in Nixpkgs' rust-analyzer](https://github.com/NixOS/nixpkgs/blob/de80d1d04ee691279e1302a1128c082bbda3ab01/pkgs/development/tools/rust/rust-analyzer/wrapper.nix#L13-14).

Considering this, I'm not sure if we want to stick with Fenix or not.

I say this since really the only thing we'd be gaining by using Fenix is quicker Rust updates, but not sure if that would be beneficial at all.

MoritzBoehme:
> This is [already set in Nixpkgs' rust-analyzer](https://github.com/NixOS/nixpkgs/blob/de80d1d04ee691279e1302a1128c082bbda3ab01/pkgs/development/tools/rust/rust-analyzer/wrapper.nix#L13-14).

Oh, it seems it was probably just user error on my part then.

I don't think we need fenix for the quicker updates to Rust, plus the speed to setup rustlings gained by getting to use cachix is probably also nice. So I'll just remove it and if it should be needed we can still add it again.

MoritzBoehme:
I think this should now be ready if you agree @winterqt and @diannasoreil.

winterqt (approved):

diannasoreil (approved):

diannasoreil:
Thank you both for your work in landing this PR!

@all-contributors please add @winterqt and @MoritzBoehme for code

diannasoreil:
hm, it can only do one at a time

@all-contributors please add @MoritzBoehme for code

PR #923, opened 2022-01-27
Quiz 2 - missing space

Plotnus:
I believe a space is missing before the `--`.
It is inconsistent with quiz3.rs

P.S. thanks for these cool exercises.


diannasoreil:
This is out of date, in `5.0-dev` quiz2 has been completely rewritten

PR #932, opened 2022-02-16
chore: bump minimum rust version

### LINKED ISSUE #931: [BUG] install-rustlings fails for rust version 1.49 (unsupported edition: 2021)
The `install-rustlings` script runs as long you have at least rust version 1.39. But the script can then complain that it cannot parse Cargo.toml because cargo is too old to know about edition 2021.

Edition 2021 was introduced in rust version 1.56 (if my knowledge is correct).
Source: https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html

I therefore suggest bumping the minimal rust version to 1.56 in `install-rustlings` script:
```bash
101| RustVersion=$(rustc --version | cut -d " " -f 2)
102| MinRustVersion=1.39
103| vercomp $RustVersion $MinRustVersion
```

Genysys:
This  PR closes #931

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -53,7 +53,7 @@ function vercomp($v1, $v2) {
 }
 
 $rustVersion = $(rustc --version).Split(" ")[1]
-$minRustVersion = "1.39"
+$minRustVersion = "1.56"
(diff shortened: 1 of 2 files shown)


diannasoreil:
I think this has already been done?

PR #935, opened 2022-02-20
fix(if2): Rename if2 exercise function to foo_if_fizz.

theherk:
The reasoning here is pretty straightforward: you don't say "Hungry, if
eat." That doesn't make sense. We want to get "foo" back if given
"fizz", so it seems this makes far more sense as "Eat, if hungry," or in
this case, return `foo_if_fizz` is given.

### DIFF
--- a/exercises/if/if2.rs
+++ b/exercises/if/if2.rs
@@ -6,7 +6,7 @@
 
 // I AM NOT DONE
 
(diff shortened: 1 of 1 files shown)


diannasoreil:
Picked on `5.0-dev`: https://github.com/rust-lang/rustlings/commit/5812f1f27b0e3c2eed6e4606635bb810d0a5727e

PR #954, opened 2022-04-03
fix(move_semantics) : add move_semantics6.rs to its mod

hpwxf:
Recent exercise `move_semantics6` was not referenced in its `mod.rs`.


diannasoreil (approved):
Thanks!

PR #957, opened 2022-04-05
Add comment about avoiding "cheesing" a solution

ragreenburg:
In `move_semantics2` you can get it to compile and pass by simply moving `vec1` below the first `println!` function, so this is to tell people to avoid doing that.


diannasoreil (approved):

diannasoreil:
@all-contributors please add @ragreenburg for content

PR #961, opened 2022-04-08
fix(errors6.rs): remove one answer code

J-S-Kim:
Although marked as 'TODO', three tests pass without any implementation because the correct answer code already exists.


diannasoreil (approved):

diannasoreil:
@all-contributors please add @J-S-Kim for content

ConchuOD:
Since this commit, the wording of this comment is now somewhat misleading as it still refers to the now removed example:
"// TODO: add another error conversion function here."

That implies that a) only one handler needs to be added b) that the first case is handled by doing nothing.

PR #968, opened 2022-04-19
refactor(arc1): improve readability by using functional style

Fointard:
Functional style is neat and shorter, I think it's good practice to promote it over imperative style.

### DIFF
--- a/exercises/standard_library_types/arc1.rs
+++ b/exercises/standard_library_types/arc1.rs
@@ -32,12 +32,7 @@ fn main() {
     for offset in 0..8 {
         let child_numbers = // TODO
         joinhandles.push(thread::spawn(move || {
-            let mut i = offset;
-            let mut sum = 0;
-            while i < child_numbers.len() {
-                sum += child_numbers[i];
-                i += 8;
-            }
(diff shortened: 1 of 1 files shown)


diannasoreil (approved):

diannasoreil:
@all-contributors please add @Fointard for content

oriontvv:
@cyril-marpaud Hello, shouldn't it be like `let sum: u32 = child_numbers.iter().filter(|&n| *n % 8 == offset).sum();`?
because lambda argument is taken twice by reference instead of current version with `filter(|n| *n % 8 == offset)`. As I understand the implicit derefernce is saving situation here but I believe we should avoid implicit stuff in tutorials. What do you think about it?

cyril-marpaud:
I agree, it could be cleaner without the implicit deref. Let's even do `filter(|&&n| n % 8 == offset)`, this way the type is explicit & no deref is needed at all, agreed ?

oriontvv:
Yes, looks good. Will you fix it?

cyril-marpaud:
> Yes, looks good. Will you fix it?

Done. Thank you for the heads up !

PR #975, opened 2022-04-22
docs(move_semantics5): Replace "in vogue" with "in scope" in hint

luskwater:
The hint for `move_semantics5` refers to "the range in which each
mutable reference is in vogue". Unless this is a deliberate
introduction of "vogue" (an admittedly-useful term because "scope"
isn't purely lexical, as in many other languages), it may be in error:
I have been unable to find the term used with reference to *Rust*
references.

Thus, I'm suggesting the replacement, in case it's been overlooked.

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -231,7 +231,7 @@ path = "exercises/move_semantics/move_semantics5.rs"
 mode = "compile"
 hint = """
 Carefully reason about the range in which each mutable reference is in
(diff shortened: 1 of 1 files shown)


diannasoreil (approved):
Thank you for catching this!

diannasoreil:
@all-contributors please add @luskwater for content

PR #987, opened 2022-05-03
fix(install.sh):fix `rust version` check

### LINKED ISSUE #985: Rustlings installation script not working on MacOS
When installing using `curl + bash`
```
curl -L https://raw.githubusercontent.com/rust-lang/rustlings/main/install.sh | bash
```

I got the following output:
```
Let's get you set up with Rustlings!
Checking requirements...
SUCCESS: Git is installed
SUCCESS: cc is installed
SUCCESS: Rust is installed
SUCCESS: Cargo is installed
bash: line 86: v2[$i]: unbound variable
```

By setting the installation script in debug mode, it seems the issue located in `vercomp` function
Bellow the debug output:

```
RustVersion=1.56.1
+ MinRustVersion=1.56
+ vercomp 1.56.1 1.56
+ [[ 1.56.1 == 1.56 ]]
+ v1=(${1//./ })
+ v2=(${2//./ })
+ len1=3
+ len2=2
+ max_len=3
+ [[ 3 -lt 2 ]]
++ seq 0 3
+ for i in '`seq 0 $max_len`'
+ '[' -z 1 ']'
+ '[' -z 1 ']'
+ '[' 1 -gt 1 ']'
+ '[' 1 -lt 1 ']'
+ for i in '`seq 0 $max_len`'
+ '[' -z 56 ']'
+ '[' -z 56 ']'
+ '[' 56 -gt 56 ']'
+ '[' 56 -lt 56 ']'
+ for i in '`seq 0 $max_len`'
+ '[' -z 1 ']'
bash: line 86: v2[$i]: unbound variable
```

likzn:
### What problem does this PR solve?
Issue Number: close #985 

Problem Summary: Our `install.sh` has `MinRustVersion` such as 1.56.And bash will have a version check,it will split to array and to check one by one.But there is a cornor case,when user use the same version but has `z` version (x.y.z).The short will out of bound(version=x.y).So i add a `pad right` to ensure the length of them is the same.
https://github.com/rust-lang/rustlings/blob/102a4b92a0c2e98ff4ba9f060850d81b4ebbe42f/install.sh#L73-L99
### What is changed and how it works?
Add a `pad right` to ensure the length of them is the same to avoid out of bound of array.
```shell
#pad right in short arr
    if [[ len1 -gt len2 ]];
    then
        for ((i = len2; i < len1; i++));
        do
            v2[$i]=0
        done
    else
        for ((i = len1; i < len2; i++));
        do
            v1[$i]=0
        done
    fi
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -75,6 +75,21 @@ function vercomp() {
     then
         max_len=$len2
     fi
+
+    #pad right in short arr
+    if [[ len1 -gt len2 ]];
+    then
+        for ((i = len2; i < len1; i++));
+        do
+            v2[$i]=0
+        done
+    else
+        for ((i = len1; i < len2; i++));
+        do
+            v1[$i]=0
+        done
+    fi
+
     for i in `seq 0 $max_len`
     do
         # Fill empty fields with zeros in v1


likzn:
@liby @diannasoreil PTAL if free.

diannasoreil (approved):
Looks good! Thanks!

PR #994, opened 2022-05-12
Add Necessary information in cargo

Jzow:
I'm not sure whether the documentation url and homepage url are correct, but I think it's necessary to add them.


diannasoreil:
Not really neccessary IMO - this isn't a published crate, and will likely never be.

Jzow:
@diannasoreil  thanks, I understand. Then I'll close this PR

PR #1025, opened 2022-06-15
fix(errors5): Add more context to exercise description

nacairns1:
This change alleviates some of the awkwardness while going through the Rustlings suggested path. In its current iteration, errors5 asks the user to create a Box<dyn err:Error> without having learned about boxes or trait objects in any meaningful way. The new description and updated hint gives a brief use case for boxes as well as explaining exactly what the exercise is looking for.

The exercise has also been modified to scaffold out a basic Box<dyn ???> in order to point the user more precisely to what the exercise is asking for.

### DIFF
--- a/exercises/error_handling/errors5.rs
+++ b/exercises/error_handling/errors5.rs
@@ -1,7 +1,17 @@
 // errors5.rs
 
-// This program uses a completed version of the code from errors4.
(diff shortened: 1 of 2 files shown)


diannasoreil:
@all-contributors please add @nacairns1 for content

PR #1032, opened 2022-06-20
feat: add strings3.rs exercise

exdx:
Add a simple exercise demonstrating how to append to a String, a pretty common operation, in two different ways.


exdx:
Closing -- I'm not super happy about my code style in this exercise, even though it would helpful to learn more about strings. I'm going to create another new exercise instead.

PR #1079, opened 2022-07-18
fix: Improve hint of lifetimes2

### LINKED ISSUE #1071: solution for `lifetimes2` is unclear
is the challenge asking to use lifetime coercion? is it asking to remove the block scope in `main()`? in my opinion there is not enough information provided for beginners to solve this, not even with the `help` command.

Tostapunk:
fixes #1071

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -795,7 +795,10 @@ name = "lifetimes2"
 path = "exercises/lifetimes/lifetimes2.rs"
 mode = "compile"
 hint = """
-What is the compiler checking? How could you change how long an owned variable lives?"""
+Remember that the generic lifetime 'a will get the concrete lifetime that is equal to the smaller of the lifetimes of x and y.
+You can take at leats two paths to achieve the desidered result while keeping the inner block:
+1. move string2 declaration to make it live as long as string1 (how is result declared?)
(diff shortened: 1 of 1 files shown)


info.toml
```diff
@@ -795,7 +795,10 @@ name = "lifetimes2"
 path = "exercises/lifetimes/lifetimes2.rs"
 mode = "compile"
 hint = """
-What is the compiler checking? How could you change how long an owned variable lives?"""
+Remember that the generic lifetime 'a will get the concrete lifetime that is equal to the smaller of the lifetimes of x and y.
+You can take at leats two paths to achieve the desidered result while keeping the inner block:
+1. move string2 declaration to make it live as long as string1 (how is result declared?)
```
diannasoreil, line 800:
```suggestion
1. Move the string2 declaration to make it live as long as string1 (how is result declared?)
```

info.toml
```diff
@@ -795,7 +795,10 @@ name = "lifetimes2"
...
+2. move println! into the inner block"""
```
diannasoreil, line 801:
```suggestion
2. Move println! into the inner block"""
```

PR #1102, opened 2022-07-23
Fix/hints typos

### LINKED ISSUE #1062: variables5 hint typo
Nothing serious but it says:
> In variables3 we already learned how to make an immutable variable mutable ...

And it was variables**4** where we learn about mutable variables.

It's better to say "In the previous exercise" or something like that

thatlittleboy:
* fixes #1062 -- I merely changed `variables3` -> `variables4`, let me know if you want to future-proof it using the suggestion mentioned in the issue.
* In `functions4.rs`, there's a reference made to the error being on line 14, but it is actually line 17 (presumably because of the `// I am not done` comment block.)

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -63,7 +63,7 @@ name = "variables5"
 path = "exercises/variables/variables5.rs"
 mode = "compile"
 hint = """
-In variables3 we already learned how to make an immutable variable mutable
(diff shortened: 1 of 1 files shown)


diannasoreil (approved):

diannasoreil:
@all-contributors please add @thatlittleboy for content

PR #1105, opened 2022-07-24
feat: add structs4.rs exercise

SAbruzzo:
Added an excise in order to test methods that just read data and methods that also change data in a struct.


SAbruzzo:
closed because opened my branch from master and not main. I will reopen

PR #1107, opened 2022-07-25
feat: add structs4.rs exercise

SAbruzzo:
Added an excise in order to test methods that just read data and methods that also change data in a struct.


RubenPX:
looks like [struct3.rs](https://github.com/rust-lang/rustlings/blob/main/exercises/structs/structs3.rs) but different

mo8it:
Resolved in https://github.com/rust-lang/rustlings/pull/2296

PR #1110, opened 2022-07-26
fix: remove extra " typo in info.toml

clairew:
When I run `rustlings watch`, main.rs panics when trying to read info.toml which is caused by the extra " character on line 748 in the info.toml file. 

edit - realized we would want to keep `"use std::fmt::Display;"` so open to a better fix that the parser will be happy with

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -746,7 +746,7 @@ path = "exercises/quiz3.rs"
 mode = "test"
 hint = """
 To find the best solution to this challenge you're going to need to think back to your
-knowledge of traits, specifically Trait Bound Syntax -  you may also need this: "use std::fmt::Display;""""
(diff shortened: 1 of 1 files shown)


elyscape:
This actually seems less like a typo and more of a failure to account for the way it's parsed.

The text immediately preceding the multi-line string delimiter is:
> you may also need this: "use std::fmt::Display;"

This is followed by 3 double-quotes to close the multi-line string. Since there's nothing between the final double-quotes of the text and the string terminator, it gets parsed as the string ending and then having an extra double-quote immediately afterwards.

Looking at the rest of the file, probably the right answer here is to replace the double-quotes in the hint text with backticks, and also throw a trailing period after them.

diannasoreil (approved):

diannasoreil:
@all-contributors please add @clairew for content

PR #1114, opened 2022-07-26
Feat submission 1

OllyDS:
Completed the first 8 tasks and first quiz.

Some more reading required on the following:
- `move_semantics` aka ownership and borrowing.
- `enums` especially using with match expressions.


OllyDS:
Accidentally pushed to this repo rather than my own :/

PR #1117, opened 2022-07-30
refactor(box1): prefer todo! over unimplemented!

Lioness100:
`todo!` better matches the exercise's intent than `unimplemented!`

> The difference between `unimplemented!` and `todo!` is that while `todo!` conveys an intent of implementing the functionality later and the message is "not yet implemented", `unimplemented!` makes no such claims. Its message is "not implemented". Also some IDEs will mark `todo!`s.

Furthermore, all other exercise's use `todo!`, so it's also a matter of consistency.


diannasoreil (approved):

diannasoreil:
@all-contributors please add @Lioness100 for content

PR #1128, opened 2022-08-07
Fix typo in quiz2

Caruso33:
Changes
// TODO: What `to` we have to import to have `transformer` in scope?
into
// TODO: What `do` we have to import to have `transformer` in scope?


diannasoreil:
Superseded by #1132

PR #1144, opened 2022-08-15
Update options1.rs

nico-vromans:
Added extra test for before 10PM and updated the test for at 10PM (when it's 10PM there should already not be any ice cream left, as per the description). Also fixed the `raw_value` test, as it is later than 10PM, so there should be no more ice cream left.

### DIFF
--- a/exercises/options/options1.rs
+++ b/exercises/options/options1.rs
@@ -19,7 +19,8 @@ mod tests {
 
     #[test]
     fn check_icecream() {
-        assert_eq!(maybe_icecream(10), Some(5));
+        assert_eq!(maybe_icecream(9), Some(5));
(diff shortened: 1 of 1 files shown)


diannasoreil (approved):

ioistired:
The raw_value test asks how much ice cream is available at 12 PM, which is before 10 PM. Therefore the old value, 5, is correct.

nico-vromans:
@ioistired, you are absolutely correct. For some reason I flipped AM with PM, so the original tests were right as well. I'll create another PR. Thanks for pointing it out.

ioistired:
@nico-vromans, no need, one has already been created at https://github.com/rust-lang/rustlings/pull/1148

PR #1148, opened 2022-08-17
fix(options1): wrong assertions

vostok92:
At 10**AM** and noon, there must be 5 pieces left.


diannasoreil:
@all-contributors please add @vostok92 for content

PR #1157, opened 2022-08-18
fix options1 assert with option

tobikris:
The assertion is incorrectly asserting for an integer instead of an Option.


tobikris:
Ah, just realized that I got confused after rebasing my finished exercise after the latest updates to options1.

PR #1159, opened 2022-08-21
fixed exercise to cover all bases

PerryCameron:
The current version needs an extra entry that I added on line 57. This allows the tests to be thorough on what the exercise intended. An example of this can best be expressed on one of the answer repositories Here: https://github.com/nestoralonso/rustlings-answers/compare/answers...PerryCameron:rustlings-answers:answers


mo8it:
Thanks :) I will add the entry to version 6.

PR #1164, opened 2022-08-24
Fix Quiz 1 >= Typo

Kodylow:
2nd Assert is for greater than or equal to 40, not "more than" like it says in the instructions.


diannasoreil:
@all-contributors please add @Kodylow for content

PR #1165, opened 2022-08-25
 add same value test case

cito-lito:
add test case for same values - > same value condition can be tested


diannasoreil:
I don't really see any reason for this? You either return `a` or `b` here anyways, and if you don't, the other tests would fail, too.

cito-lito:
yeah sure, while I did the exercise I just thought it could be nice to have a test case for  same values, but for the purpose of the code, I see that it may not make sense : )

PR #1170, opened 2022-08-28
fix(quiz1): reverting commit that made description incorrect. 

### LINKED ISSUE #1169: quiz1 test case does not match problem description
The problem description for quiz1 is as follows:
```rust
// Mary is buying apples. One apple usually costs 2 Rustbucks, but if you buy
// 40 or more at once, each apple only costs 1! Write a function that calculates
// the price of an order of apples given the quantity bought. No hints this time!
```
This means that if you buy anything less than 40 apples, the cost is 2 Rustbucks per apple and the cost == 2 * the number of apples. Starting at 40 apples, the cost is reduced to 1, so for 40 or more apples, the cost == the number of apples.

However, the second test case is
```
let price2 = calculate_price_of_apples(40);
...
assert_eq!(80, price2);
```
This implies that the cost at 40 apples is still 2 Rustbucks per apple.

Recent commit https://github.com/rust-lang/rustlings/commit/d6e26bb350e423a8f2a90d9405060fdaf9893736 is incorrect and should be reverted.

LeinadAsid:
closes #1169

### DIFF
--- a/exercises/quiz1.rs
+++ b/exercises/quiz1.rs
@@ -5,7 +5,7 @@
 // - If
 
 // Mary is buying apples. One apple usually costs 2 Rustbucks, but if you buy
-// 40 or more at once, each apple only costs 1! Write a function that calculates
+// more than 40 at once, each apple only costs 1! Write a function that calculates
 // the price of an order of apples given the quantity bought. No hints this time!
 
 // I AM NOT DONE


DanielBauman88 (approved):

robinvandernoord:
Agreed, I was confused my test was failing as well. Either the test or the description should be updated

septatrix (approved):

diannasoreil:
Fixed on main

PR #1171, opened 2022-08-28
Add quotes for readability of output

memark:
Adding quotes to make it easier to see what part comes from which string.

Before
`The longest string is long string is long`

After
`The longest string is 'long string is long'`


diannasoreil (approved):

diannasoreil:
@all-contributors please add @memark for content

PR #1174, opened 2022-08-29
fix: quiz1 - price is cheaper for 40 or more

DanielBauman88:
The description says that the discount is for >= 40
but the test for the value of 40 doesn't apply the discount.


DanielBauman88:
Closing - already fixed by [this](https://github.com/rust-lang/rustlings/pull/1170/files) open PR.

PR #1178, opened 2022-08-31
Fix quiz1 description

ceth-x86:
Description:

> but if you buy 40 or more at once, each apple only costs 1

Code:

```rs
let price2 = calculate_price_of_apples(40);
assert_eq!(80, price2);
```

I get 40 apples. It equals to 40 ("40 or more at once" -> true). So each apple cost 1 bucks and the answer must be 40 for current condition.

### DIFF
--- a/exercises/quiz1.rs
+++ b/exercises/quiz1.rs
@@ -5,7 +5,7 @@
 // - If
 
 // Mary is buying apples. One apple usually costs 2 Rustbucks, but if you buy
-// 40 or more at once, each apple only costs 1! Write a function that calculates
+// more than 40 at once, each apple only costs 1! Write a function that calculates
 // the price of an order of apples given the quantity bought. No hints this time!
 
 // I AM NOT DONE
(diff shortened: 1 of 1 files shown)


Evgeny-SPB:
This pull request is a duplicate of https://github.com/rust-lang/rustlings/pull/1170

robinvandernoord:
Agreed, I was confused my test was failing as well. Either the test or the description should be updated

satyambnsal:
> Description:
> 
> > but if you buy 40 or more at once, each apple only costs 1
> 
> Code:
> 
> ```rust
> let price2 = calculate_price_of_apples(40);
> assert_eq!(80, price2);
> ```
> 
> I get 40 apples. It equals to 40 ("40 or more at once" -> true). So each apple cost 1 bucks and the answer must be 40 for current condition.

lol was about to raise PR for the same 😂

diannasoreil:
Fixed on main

PR #1182, opened 2022-09-03
Update quiz1.rs

girish1605:
line 8 : // 40 or more at once, each apple only costs 1! Write a function that calculates

The above statement doesn't mach with the 
line 24: assert_eq!(80, price2); 

it supposed to be assert_eq!(40, price2);


diannasoreil:
Fixed on main

PR #1184, opened 2022-09-05
fix(quiz1): Inaccurate instructions

HPaulson:
The code checks for an assertion of `assert_eq!(80, price2);` which means that we only double the cost when the number of apples >= 41, not 40; Changing the language here makes the instructions clear


HPaulson:
Duplicate of #1170, #1178, #1179, & #1182
Should've searched the PRs here before creating my own -- rookie mistake. Sorry!

PR #1187, opened 2022-09-05
Make changes to quiz1 so that tests match problem.

Zarzoom:
40 or more implies quantities greater than or equal receive the discount. Therefore, when I pass in 40 during the test I should return 40 instead of 80.


diannasoreil:
Fixed on main by rewording the comment

PR #1188, opened 2022-09-06
Revert #1164

yjhn:
#1164 introduced an error rather than fixing it. Second assert states that 40 apples have a price of 80, therefore, 40 apples still cost 2 Rustbucks per apple.


yjhn:
Added another assert to make the point perfectly clear.

diannasoreil:
I rewrote the comment for this exercise a bit on `main` to hopefully make it clearer, and added a fourth assert as well.

PR #1192, opened 2022-09-09
fix: Add a deref in the test code

aaarkid:
This change makes it possible to finish the last exercise without touching the test code, seems like an overlook when code was added since there's nothing in the hints about this extension neither.

It's virtually impossible to write a the `num_sq` function to take the Box since it doesn't implement `MulAssign`.

### DIFF
--- a/exercises/conversions/as_ref_mut.rs
+++ b/exercises/conversions/as_ref_mut.rs
@@ -54,7 +54,7 @@ mod tests {
     #[test]
     fn mult_box() {
         let mut num: Box<u32> = Box::new(3);
-        num_sq(&mut num);
+        num_sq(&mut *num);
         assert_eq!(*num, 9);
     }
 }


aaarkid:
I believe this PR is also a good moment to add something in the hints regard the `AsMut` part.

aaarkid:
Can this PR be merged? It only changes 1 line and saves people finishing Rustlings a good bunch of minutes up to an hour.

diannasoreil (approved):

diannasoreil:
@all-contributors please add @aaarkid for content

shepmaster:
> This change makes it possible to finish the last exercise without touching the test code

I'm surprised at this PR as **now** I can't implement the function _without_ changing the test code.

This was my implementation:

```rust
fn num_sq<T: AsMut<u32>>(mut arg: T) {
    let arg = arg.as_mut();
    *arg *= *arg;
}
```

I have to remove the `*` added by this PR to get it to compile.

@aaarkid can you share your solution?

aaarkid:
Sure, in a bit.

seritools:
I also can't get it to compile in the "obvious" way, and it certainly works without the `*` in the test code.

Another argument in favor of reverting this - the example right in the `AsMut` documentation shows it working *without* the `*`:
https://doc.rust-lang.org/std/convert/trait.AsMut.html#examples

aaarkid:
I see the issue. You're right the deref shouldn't be there. I'll open a new PR tomorrow to revert this. I apologize for the trouble this may have caused.

In the meanwhile, do you think that the `let arg = arg.as_mut();` should already be in the code?

shepmaster:
> In the meanwhile, do you think that the `let arg = arg.as_mut();` should already be in the code?

I'm not the target audience of Rustlings, but my gut says "no, it should not be there".

azzamsa:
My CI fails because of this change https://github.com/azzamsa/rustlings/actions/runs/3317605384/jobs/5480649114#step:6:12

PR #1202, opened 2022-09-16
Switch `arc` and `rc` exercises order

skaunov:
`rc` should totally be introduced _before_ `arc`


shadows-withal (approved):

shadows-withal:
Hm, I must have merged a similar PR already, `main` already has this ordering...

PR #1211, opened 2022-09-27
Open next pending exercise with `rustlings next`

### LINKED ISSUE #1201: Open next practice source code by specified editor
Some IDE allow us to open an source file in current windows if they are in the folder we opened.

For example, rustling run `code exercises/primitive_types/primitive_types3.rs` if user set `RUSTLING_EDITOR` as `code {}` and just finished primitive_types2.

This avoids bad experience switching to next chapter on some IDE.

For `vim` users, to make buffers order as pratices':
```bash
vim `rustlings list|head -n -1|tail -n +2|cut -f 2`
```

mwrites:
**Problem:**
It is cumbersome to find the next pending exercise. 
[closes #1201](https://github.com/rust-lang/rustlings/issues/1201)

**Solution:**
Added a new command to open the next pending exercise in $EDITOR:
```bash
rustlings next
# or EDITOR=vim rustlings next
# or EDITOR=code rustlings next
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,6 +98,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "524cbf6897b527295dff137cec09ecf3a05f4fddffd7dfcd1585403449e74198"
 
 [[package]]
+name = "edit"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c562aa71f7bc691fde4c6bf5f93ae5a5298b617c2eb44c76c87832299a17fbb4"
+dependencies = [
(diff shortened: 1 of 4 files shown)


diannasoreil:
I'm not sure how many people this would benefit, most people do Rustlings via `watch` mode, in which case this is nigh useless. It would be nice if there was a watch-mode command to do this.

Even then, I'd wager most people use a GUI editor instead of a TUI editor, so it's probably just going to be confusing to them when something pops open in their terminal. The sheer breadth of editors and the way people configure (or don't!) a "default" editor makes this a really tricky issue to get right.

I'm disinclined to add this to Rustlings for the above reasons, as it'd confuse most people more than anything.

PR #1216, opened 2022-09-30
feat: Add VSCode extension recommendation

gfaugere:
VSCode supports recommending extensions through a file : https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

This is what the prompt looks like:    
image

I think this allows us to help newcomers set-up their tooling easily, and that the recommended extension ([rust-analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer)) is pretty consensual.

Usually I am in favour of leaving IDE config files out of repositories, but I think this is a good fit for rustlings.

Let me know your thoughts!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ exercises/clippy/Cargo.toml
 exercises/clippy/Cargo.lock
 rust-project.json
 .idea
-.vscode
+.vscode/*
+!.vscode/extensions.json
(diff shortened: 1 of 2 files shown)


diannasoreil:
@all-contributors please add @gfaugere for tools

PR #1217, opened 2022-10-01
fix(verify): print feedback on verify pass for tests

jan25:
`rustlings run next` silently exits without any success message for examples with tests. This change fixes that to have similar output as `rustlings watch`.

Example output:
```
➜ rustlings run next
✅ Successfully tested exercises/if/if1.rs!

🎉 🎉  The code is compiling, and the tests pass! 🎉 🎉

You can keep working on this exercise,
or jump into the next one by removing the `I AM NOT DONE` comment:

 2 |  // Execute `rustlings hint if1` or use the `hint` watch subcommand for a hint.
 3 |  
 4 |  // I AM NOT DONE
 5 |  pub fn bigger(a: i32, b: i32) -> i32 {
 6 |      if a > b {
```

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -15,14 +15,15 @@ pub fn verify<'a>(
 ) -> Result<(), &'a Exercise> {
     let (num_done, total) = progress;
     let bar = ProgressBar::new(total as u64);
-    bar.set_style(ProgressStyle::default_bar()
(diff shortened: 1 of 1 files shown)


jan25:
Ah, looks like there is a flag to show test output `rustlings  --nocapture run if2`. Closing this PR, can reopen if it makes sense to show the prompt by default for `run` subcommand both for compile and test examples.

PR #1220, opened 2022-10-03
feat(verify): add progress percentage in watch mode

azzamsa:


src/verify.rs
```diff
@@ -29,6 +29,8 @@ pub fn verify<'a>(
         if !compile_result.unwrap_or(false) {
             return Err(exercise);
         }
+        let percentage = num_done as f32 / total as f32 * 100.0;
+        bar.set_message(format!("({:.2} %)", percentage));
```
diannasoreil, line 33: Could we round this to one decimal place instead? There's not that many exercises, so rounding it to 2 feels kinda overblown.

diannasoreil (approved):

diannasoreil:
@all-contributors please add @azzamsa for code

PR #1232, opened 2022-10-13
Use inline format argument syntax

nyurik:
The inline format is clearer and more concise.

### DIFF
--- a/exercises/clippy/clippy2.rs
+++ b/exercises/clippy/clippy2.rs
@@ -9,5 +9,5 @@ fn main() {
     for x in option {
         res += x;
     }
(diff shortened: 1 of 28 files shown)


shadows-withal:
I'd be partial to merging this, but I think we should at least explain that there is another way of doing `format!` arguments, maybe add a comment in `variables1`? Something like "This is a more concise version of `format!(..., ...)`.

shadows-withal:
Okay, this Clippy lint has been downgraded to pedantic in Rust 1.67.1, and rust-analyzer doesn't support it yet, so I'll close it until things change in Rust core.

PR #1235, opened 2022-10-14
feat: add strings5.rs exercise

bazik123:
Added an exercise that explores println/format syntax.


shadows-withal:
If Rustlings would ever have an exercise on `println!/format!` syntax, I think it'd be after Traits so that it could explore the `Display` and `Debug` trait. In this form, I'm going to have to say no.

PR #1236, opened 2022-10-16
style: explicitly use Arc::clone

mvanschellebeeck:
The [Rust book](https://doc.rust-lang.org/book/ch15-04-rc.html?highlight=clone()#cloning-an-rc-increases-the-reference-count) recommends `Arc::clone` over just `.clone()` to highlight the fact that this doesn't perform a deep copy the way other types' implementations do.


diannasoreil (approved):

diannasoreil:
@all-contributors please add @mvanschellebeeck for content

PR #1237, opened 2022-10-17
fix: test case with wrong expectation

illegalcall:


bazik123:
you didnt understand what 'Append' command does,  there is no error that needs fixing here.  'Append' appends 'bar' to whatever string is given, and not duplicates the given string as you seem to think.

illegalcall:
Oh Now I understand. Thank you @bazik123 .

Issue #1195, opened 2022-09-11
Clone-on-Write exercise is unclear


black-puppydog:
I am just going through `exercises/standard_library_types/cow1.rs` and while I think I understand the type (and have read up on it in the reference) I don't really see what I'm supposed to _do_ in this exercise.

Specifically, I find these lines confusing:

https://github.com/rust-lang/rustlings/blob/c923e7af73a91970d2e63e03babbca9cc0817551/exercises/standard_library_types/cow1.rs#L41-L47

As a learner I'm not sure what I'm expected to do here. I can just replace `Borrowed` by `Owned` and make the code compile & run. However, that seems too obvious as a solution, given that it's literally the same as a few lines above. And then the `panic!` also needs to be changed I guess, that assertion seems a bit confusing.

Maybe simply writing `// TODO: modify the following two lines to match the situation` or some such would take the uncertainty out of the situation. Or, if that seems _too_ easy then, maybe just leave the match incomplete and let the learner fill it completely?

I'm obviously nit-picking a bit here; including `cow` into the course is a great thing! 👍🏽

Jzow:
This code only tells you that the `input` has the ownership of the `slice` value, so it will not clone

black-puppydog:
yes I understood that, the reference of `cow` is quite clear about it. Maybe this is really just my particular brain that found this exercise a bit confusing. I think the reason might be that there are two lines, one of which affirming "expected borrowed value". I think I wouldn't have been confused if for example the match only contained the following:

```rust
match abs_all(&mut input) { 
    ... => println!("I own this slice!"), 
    _ => panic!("expected ... value"), 
} 
```

That makes the exercise even a little more obvious.
FWIW, I won't push hard on this, it's not a huge issue. But it was pretty much the first hiccup I had in the flow state of going through the rustlings so I found it noteworthy. 🙂

qqrm:
I agree, the description of panic is misleading.

dolzenko:
Confused here too, so the solution of the exercise is to `Cow::Borrowed => Cow::Owned` in the last match?

JonShort:
I also got confused here, for me it's because the rest of the course is TDD but here the assertion is unclear/incorrect

```rust
_ => panic!("expected borrowed value"),
```

Agree with https://github.com/rust-lang/rustlings/issues/1195#issuecomment-1246539662 that altering this text would improve the exercise

black-puppydog:
[...]

PR #1244, opened 2022-10-23
refactor(cow1):  replace main with tests

black-puppydog:
Following the discussion in #1195 this is the best I could come up with. The issue for me (and apparently a few other learners) was that the code needed to complete the exercise was not _missing_, but was rather there but wrong.

In the end, what made the difference between this exercise and others (for me) was that in this exercise I was supposed to learn what to *expect* of an output. So I think it makes sense here to let the learner modify the tests and not the code itself.

Side note: once [`assert_match!`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) becomes stable I'd like to integrate it here just to expose the learner to it and because it makes the tests much clearer to read. This is how I had written it before realizing that probably most people don't use nightly to play through rustlings. 😆 

```rust
#[test]
fn reference_no_mutation() {
    // No clone occurs because `input` doesn't need to be mutated.
    let slice = [0, 1, 2];
    let mut input = Cow::from(&slice[..]);
    assert_match!(
        abs_all(&mut input),
        Cow::Borrowed(_),
        "Expected borrowed value"
    );
}
````

### DIFF
--- a/exercises/smart_pointers/cow1.rs
+++ b/exercises/smart_pointers/cow1.rs
@@ -4,6 +4,9 @@
 // Cow is a clone-on-write smart pointer.
 // It can enclose and provide immutable access to borrowed data, and clone the data lazily when mutation or ownership is required.
 // The type is designed to work with general borrowed data via the Borrow trait.
+//
+// This exercise is meant to show you what to expect when passing data to Cow.
+// Fix the unit tests by checking for Cow::Owned(_) and Cow::Borrowed(_) at the TODO markers.
 
 // I AM NOT DONE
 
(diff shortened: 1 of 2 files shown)


shadows-withal:
@black-puppydog Looks good! Sorry for the late review, could you fix the merge conflicts? Then I'll merge it ASAP.

black-puppydog:
@shadows-withal no reason to be sorry. thanks for taking the time to look at this! 

I rebased and then rebased my solutions branch on top of this, I'd say this is good to go then. 🙂

shadows-withal (approved):

shadows-withal:
@all-contributors please add @black-puppydog for contetn

shadows-withal:
oops

@all-contributors please add @black-puppydog for content

PR #1245, opened 2022-10-24
fix(traits4): Updated `compare_license_types` to Use References

yakovmanshin:
* The `compare_license_types` function shouldn’t take ownership of its parameters

### DIFF
--- a/exercises/traits/traits4.rs
+++ b/exercises/traits/traits4.rs
@@ -33,7 +33,7 @@ mod tests {
         let some_software = SomeSoftware {};
         let other_software = OtherSoftware {};
 
-        assert!(compare_license_types(some_software, other_software));
+        assert!(compare_license_types(&some_software, &other_software));
     }
 
     #[test]
@@ -41,6 +41,6 @@ mod tests {
         let some_software = SomeSoftware {};
         let other_software = OtherSoftware {};
 
-        assert!(compare_license_types(other_software, some_software));
+        assert!(compare_license_types(&other_software, &some_software));
     }
 }


diannasoreil:
Why not? It's fine here.

yakovmanshin:
> Why not? It's fine here.

It may be fine from the technical standpoint, but it doesn’t make much sense in a real-life scenario. I mean, why would a function whose only responsibility is reading and comparing some fields on the objects consume them?

diannasoreil:
Hm, it's not really about real-life use here, though, it's about teaching `impl Trait`. Changing this to reference arguments would mean having to write `&impl Licensed`, which would be more clutter that detracts from the point of the exercise.

Either way, this is bikeshedding, but I'm against this change.

diannasoreil:
In the same vein, the error produced by the compiler when forgetting to add the ampersand to `impl Licensed` suggests _removing_ the reference in the function call:

CleanShot 2022-10-26 at 11 40 09@2x

yakovmanshin:
> Hm, it's not really about real-life use here, though, it's about teaching `impl Trait`. Changing this to reference arguments would mean having to write `&impl Licensed`, which would be more clutter that detracts from the point of the exercise.
> 
> Either way, this is bikeshedding, but I'm against this change.

I totally agree this exercise is about traits.

However, being consistent throughout the course in regard to when to move ownership and when to use references is beneficial for (long-term) comprehension. It doesn’t feel right to explain once when references are appropriate but then not stick to those principles in the rest of the exercises.

PR #1256, opened 2022-11-03
fix: Fixed instructions for options1.rs

officialabdulrehman:
options1.rs

The instruction said:

```
At 10PM, someone eats them all, so there'll be no more left :(
```

According to the tests, the instruction should have been something like this:

```
Once the clock hits 10PM, someone eats them all, so there'll be no more left from there onwards:(
```

Changed accordingly.

Thanks for this really nice learning tool.

options1.rs Before
```
// options1.rs
// Execute `rustlings hint options1` or use the `hint` watch subcommand for a hint.

// I AM NOT DONE

// This function returns how much icecream there is left in the fridge.
// If it's before 10PM, there's 5 pieces left. At 10PM, someone eats them
// all, so there'll be no more left :(
// TODO: Return an Option!
fn maybe_icecream(time_of_day: u16) -> Option<u16> {
    // We use the 24-hour system here, so 10PM is a value of 22 and 12AM is a value of 0
    // The Option output should gracefully handle cases where time_of_day > 23.
    ???
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn check_icecream() {
        assert_eq!(maybe_icecream(9), Some(5));
        assert_eq!(maybe_icecream(10), Some(5));
        assert_eq!(maybe_icecream(23), Some(0));
        assert_eq!(maybe_icecream(22), Some(0));
        assert_eq!(maybe_icecream(25), None);
    }

    #[test]
    fn raw_value() {
        // TODO: Fix this test. How do you get at the value contained in the Option?
        let icecreams = maybe_icecream(12);
        assert_eq!(icecreams, 5);
    }
}

```

options1.rs After
```
// options1.rs
// Execute `rustlings hint options1` or use the `hint` watch subcommand for a hint.

// I AM NOT DONE

// This function returns how much icecream there is left in the fridge.
// If it's before 10PM, there's 5 pieces left. Once the clock hits 10PM, someone eats them
// all, so there'll be no more left from there onwards:(
// TODO: Return an Option!
fn maybe_icecream(time_of_day: u16) -> Option<u16> {
    // We use the 24-hour system here, so 10PM is a value of 22 and 12AM is a value of 0
    // The Option output should gracefully handle cases where time_of_day > 23.
    ???
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn check_icecream() {
        assert_eq!(maybe_icecream(9), Some(5));
        assert_eq!(maybe_icecream(10), Some(5));
        assert_eq!(maybe_icecream(23), Some(0));
        assert_eq!(maybe_icecream(22), Some(0));
        assert_eq!(maybe_icecream(25), None);
    }

    #[test]
    fn raw_value() {
        // TODO: Fix this test. How do you get at the value contained in the Option?
        let icecreams = maybe_icecream(12);
        assert_eq!(icecreams, 5);
    }
}

```

### DIFF
--- a/exercises/options/options1.rs
+++ b/exercises/options/options1.rs
@@ -4,8 +4,8 @@
 // I AM NOT DONE
 
 // This function returns how much icecream there is left in the fridge.
-// If it's before 10PM, there's 5 pieces left. At 10PM, someone eats them
-// all, so there'll be no more left :(
+// If it's before 10PM, there's 5 pieces left. Once the clock hits 10PM, someone eats them
+// all, so there'll be no more left from there onwards:(
 // TODO: Return an Option!
 fn maybe_icecream(time_of_day: u16) -> Option<u16> {
     // We use the 24-hour system here, so 10PM is a value of 22 and 12AM is a value of 0


shadows-withal:
Hm, this feels like it says the exact same thing, just more verbose. I'm not against changing wording to make it more easily understandable, but I don't think it serves that purpose here. I find it more confusing now.

mo8it:
Sorry, closing because I don't think that the change is worth, especially after the improvements in version 6.

PR #1268, opened 2022-11-14
fix #1267

### LINKED ISSUE #1267: rustlings run clippy3 error
rustlings run clippy3 will notice error but rustlings watch and rustlings verify is ok

livstyle:
question description: rustlings run clippy3 will notice error but rustlings watch and rustlings verify is ok; 
fix: when the run function match Mode::Clippy type and then execute compile_and_run; maybe should execute compile_only

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,7 +1,7 @@
 use std::process::Command;
 
 use crate::exercise::{Exercise, Mode};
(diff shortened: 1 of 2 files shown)


mo8it:
Closing because this should be fixed in v6.

PR #1278, opened 2022-11-21
fix typo

hippietrail:
fix typo: what to we have to ... -> what do we have to ...


diannasoreil:
Superseded by #1279, I think?

PR #1290, opened 2022-12-07
docs: fix ordering of exercises

KentGrigo:
I've sorted the exercises in the README based on the first book chapter.
I think this will make it easier to find the exercise that matches the book chapter and also make it easier to get an overview of the chronological order.

### DIFF
--- a/exercises/README.md
+++ b/exercises/README.md
@@ -3,22 +3,22 @@
 | Exercise               | Book Chapter        |
 | ---------------------- | ------------------- |
 | variables              | §3.1                |
+| primitive_types        | §3.2, §4.3          |
 | functions              | §3.3                |
 | if                     | §3.5                |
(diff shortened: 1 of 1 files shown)


diannasoreil:
The exercise ordering is specifically like this because while Rustlings does use the book to learn alongside with, it doesn't match 100% because in our context (mostly teaching syntax), matching the book 100% doesn't make sense or slows the learner down.

KentGrigo:
Thank you, @diannasoreil, for the clarification.
I wasn't aware that the given order was the intended order of solving the exercises.
Would a note like "We recommend the given order of solving the exercises" (or something like that) be fitting?

diannasoreil:
> Would a note like "We recommend the given order of solving the exercises" (or something like that) be fitting?

That's already what `rustlings watch` does by default, so I don't see a reason for clarifying it.

PR #1305, opened 2022-12-26
fix(threads1): make program panic if threads are not joined

### LINKED ISSUE #1298: threads1 does not panick if not all handles are joined
It seems to me that the panic check in exercise `threads1` is incomplete.

Here's the exercise:
```rust
// threads1.rs
// Execute `rustlings hint threads1` or use the `hint` watch subcommand for a hint.
// This program should wait until all the spawned threads have finished before exiting.

// I AM NOT DONE

use std::thread;
use std::time::Duration;


fn main() {

    let mut handles = vec![];
    for i in 0..10 {
        thread::spawn(move || {
            thread::sleep(Duration::from_millis(250));
            println!("thread {} is complete", i);
        });
    }

    let mut completed_threads = 0;
    for handle in handles {
        // TODO: a struct is returned from thread::spawn, can you use it?
        completed_threads += 1;
    }

    if completed_threads != 10 {
        panic!("Oh no! All the spawned threads did not finish!");
    }
    
}
```
The solution (at least mine) is to add each thread to `handles` in the first loop and join each handle in the second loop like in the block below. The issue is that the panic check actually always succeeds, even if the handles are not joined. I think it should panic if some of the threads are not finished.
```rust
// threads1.rs
// Execute `rustlings hint threads1` or use the `hint` watch subcommand for a hint.
// This program should wait until all the spawned threads have finished before exiting.

// I AM NOT DONE

use std::thread;
use std::time::Duration;


fn main() {

    let mut handles = vec![];
    for i in 0..10 {
        handles.push(thread::spawn(move || { // <- push handle to handles so you can track it
            thread::sleep(Duration::from_millis(250));
            println!("thread {} is complete", i);
        }));
    }

    let mut completed_threads = 0;
    for handle in handles {
        // TODO: a struct is returned from thread::spawn, can you use it?
        handle.join();          // <- code happily compiles and runs without this
        completed_threads += 1;
    }

    if completed_threads != 10 { // <- this condition is never met
        panic!("Oh no! All the spawned threads did not finish!");
    }
    
}
```
Upon trying to find how to alter the exercise I quickly discovered it's not easy to change without convoluting the whole thing, so maybe this was more of a conscious decision than a bug.

I tried to come up with an alteration to the exercise and this is my best attempt. I don't think it's great as it introduces adding an empty result to a vector, which doesn't really make much sense.
```rust
// threads1.rs
// Execute `rustlings hint threads1` or use the `hint` watch subcommand for a hint.
// This program should wait until all the spawned threads have finished before exiting.

// I AM NOT DONE

use std::thread;
use std::time::Duration;


fn main() {

    let mut handles = vec![];
    for i in 0..10 {
        thread::spawn(move || {
            thread::sleep(Duration::from_millis(250));
            println!("thread {} is complete", i);
        });
    }

    let mut results = vec![];
    for handle in handles {
        // TODO: a struct is returned from thread::spawn, can you use it?
    }


    if results.len()!= 10 {
        panic!("Oh no! All the spawned threads did not finish!");
    }
}
```

platformer:
Attempts to close #1298

- program verifies that threads have been joined by checking if `results` vector has been filled
- solution now involves recognizing that `JoinHandle::join` returns a `Result`, unpacking it, and adding the value to a vector

### DIFF
--- a/exercises/threads/threads1.rs
+++ b/exercises/threads/threads1.rs
@@ -1,31 +1,38 @@
 // threads1.rs
 // Execute `rustlings hint threads1` or use the `hint` watch subcommand for a hint.
-// This program should wait until all the spawned threads have finished before exiting.
+
+// This program spawns multiple threads that each run for at least 250ms,
+// and each thread returns how much time they took to complete.
+// The program should wait until all the spawned threads have finished and
+// should collect their return values into a vector.
 
 // I AM NOT DONE
 
 use std::thread;
-use std::time::Duration;
-
+use std::time::{Duration, Instant};
 
 fn main() {
-
     let mut handles = vec![];
     for i in 0..10 {
-        thread::spawn(move || {
+        handles.push(thread::spawn(move || {
+            let start = Instant::now();
             thread::sleep(Duration::from_millis(250));
             println!("thread {} is complete", i);
-        });
+            start.elapsed().as_millis()
+        }));
     }
 
-    let mut completed_threads = 0;
+    let mut results: Vec<u128> = vec![];
     for handle in handles {
         // TODO: a struct is returned from thread::spawn, can you use it?
-        completed_threads += 1;
     }
 
-    if completed_threads != 10 {
+    if results.len() != 10 {
         panic!("Oh no! All the spawned threads did not finish!");
     }
     
+    println!();
+    for (i, result) in results.into_iter().enumerate() {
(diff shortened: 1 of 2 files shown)


shadows-withal (approved):

shadows-withal:
@all-contributors please add @platformer for content

PR #1306, opened 2022-12-27
feat: refactored as_ref_mut.rs to as_ref.rs and as_mut1.rs and add as_mut2.rs

jtcours:


mo8it:
I don't think that we should mix in other trait bounds. Users should focus on understanding `AsMut` in this exercise.

PR #1313, opened 2023-01-01
Reorder exercises

seporterfield:
Hi, I noticed the same issues as the creator of issue [#813](https://github.com/rust-lang/rustlings/issues/813).

It's more of an annoyance than a proper challenge for new Rust learners using Rustlings that the next exercise they come across in sequence has unintroduced concepts that go completely unexplained, along with the ones being introduced intentionally in the very same exercise. Also, "standard library types" is a bit clumsy, and I think we should try to be roughly in line with the book where it's possible and easy to do so.

So, I reordered these exercises, separated the smart pointer exercises from the iterator exercises, and made other updates accordingly.

Closes [#813](https://github.com/rust-lang/rustlings/issues/813).


diannasoreil (approved):

diannasoreil:
@all-contributors please add @seporterfield for content

PR #1319, opened 2023-01-03
fix(verify): progress bar percent now updates with the number of files verified

cloppingemu:
Noticed that the percent of exercises completed did not update in the progress bar when the sub-commands verify or watch were run. The percentage would update when working through the exercises - when the loop section section of the watch routine in src/main.rs had been entered.

Fixed the verify function in src/verify.rs so that the percentage would update by counting the number of exercises verified. A cleaner solution would be to use the _percent_ key in the progress bar template (https://docs.rs/indicatif/latest/indicatif/index.html#templates) to implicitly change the progress percentage instead of explicitly manipulating the message to display the progress percentage. The _percent_ key, however, does not show any digits after the decimal point (maybe another feature to implement there).

PS. First time contributing to a public project. Please point out if there are any deficiencies in the PR.

Thank you

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -13,13 +13,15 @@ pub fn verify<'a>(
     progress: (usize, usize),
     verbose: bool,
 ) -> Result<(), &'a Exercise> {
-    let (num_done, total) = progress;
+    let (mut num_done, total) = progress;
     let bar = ProgressBar::new(total as u64);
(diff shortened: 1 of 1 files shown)


diannasoreil (approved):

diannasoreil:
@all-contributors please add @cloppingemu for code

PR #1328, opened 2023-01-06
Correct borrowing

Anri-Lombard:
The previous code does not allow a good solution by just changing the function parameters, unless I am mistaken.

This code allows "&dyn Licensed" to be substituted for "??" and works well.


Anri-Lombard:
Gotcha

PR #1330, opened 2023-01-09
using costs::PI

mrsumanbiswas:
using std lib for value of PI


diannasoreil:
That's the solution of the exercise.

PR #1334, opened 2023-01-10
fix bug when showing percentage

focksor:
num_done here will alaways be zero, so the percentage here will always be 0%. 
fix bug by using bar.position() instead of num_done so that we can get the right percentage .


diannasoreil:
I'd reckon this is superseded by #1319

PR #1335, opened 2023-01-10
Added dikjstra algorithm

Irene-123:
Hello there, I've added an example of the implementation of the Dijkstra algorithm to find the shortest path between 2 nodes. Please review and let me know thank you so much!


diannasoreil:
This is out of scope for this project. Sorry!

PR #1342, opened 2023-01-17
Update hint for `exercises/conversions/as_ref_mut.rs`

Ttibsi:
It appears that the tasks for `exercises/conversions/as_ref_mut.rs` were updated [about 6 months ago](https://github.com/rust-lang/rustlings/blame/main/exercises/conversions/as_ref_mut.rs#L22) without updating the hint for the task. I'm not the most clear on the wording we'd want to use here for it, so I'm open to any feedback on this from the maintainers.

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -1160,4 +1160,6 @@ name = "as_ref_mut"
 path = "exercises/conversions/as_ref_mut.rs"
 mode = "test"
 hint = """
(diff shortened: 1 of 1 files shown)


Ttibsi:
Closing this as it turns out #1113 has the same thing, although it's been ignored for 6 months too? I'll leave a comment there instead

PR #1354, opened 2023-02-03
Cucumbers are not vegetables

HexExecute:
Replaced cucumber with broccoli in modules2 since cucumbers are not vegetables


shadows-withal:
Hmm, they aren't? Maybe not botanically, but culinarily, they are, so I think the distinction is so bikeshedd-y that it's not really worth debating about.

PR #1355, opened 2023-02-05
Fix the problem of different edition between rustc and rust-analyzer

wjwrh:
The edition args of `rustc` is set to 2018 by default. While rustling generate `rust-project.json` [with edition setting to 2021](https://github.com/rust-lang/rustlings/blob/main/src/project.rs#L46). It may cause [some problems](https://users.rust-lang.org/t/rust-analyzer-and-cargo-give-different-type-inferences-for-the-same-code/88790).


shadows-withal (approved):

shadows-withal:
@all-contributors please add @wjwrh for code

PR #1378, opened 2023-02-17
Updated README.md for the exercises with repective links.

An-Yay:
links for each exercise have been attached to the topics so that they can be opened up easily!


shadows-withal:
I don't think this is a huge issue, you're most likely looking at this file directly in your editor, at which point you ought to just click on the folder name in the sidebar. Adding Markdown link formatting to this would destroy the simplicity of the table, which is important because most people will view it in source view.

PR #1379, opened 2023-02-17
Removed verbose instructions from the uninstall section

Anush008:
The truncated lines from the uninstall instructions are gratuitous. Removing them makes the section more concise and clear.


shadows-withal:
@all-contributors please add @Anush008 for docs

PR #1381, opened 2023-02-18
fix: unify undisclosed type notation in errors5.rs

shgew:
I had to re-read the edited line a few times to understand that "..." and "???" refer to the same thing. I think, applying this change will keep the notation consistent across the comment and the body of the exercise, since `main`'s signature is initially defined with question marks as well.
```rust
fn main() -> Result<(), Box<dyn ???>>
```


shadows-withal (approved):

shadows-withal:
@all-contributors please add @shgew for content

PR #1396, opened 2023-02-24
fix(macros4.rs): Add rustfmt::skip to prevent auto-fix.

0lhi:
The `macros4.rs` challenge can automatically be solved by rustfmt without the user noticing.

Adding `#[rustfmt::skip]` above the `macro_rules!` line fixes this issue.


shadows-withal (approved):
good catch!

PR #1404, opened 2023-02-27
Always showing correct progress percentages.

nidhalmessaoudi:
## The issue
I noticed that the progress bar in watch mode is not showing the actual progress percentage when the compilation is failed (Showing 0.0%). This is because the progress percentage is only set when the exercise compilation is successful in verify.rs file.
## The fix
Setting the progress percentage before looping over the exercises and updating it when the exercise compilation is successful
## Capture of the issue
Capture

### DIFF
--- a/exercises/error_handling/errors5.rs
+++ b/exercises/error_handling/errors5.rs
@@ -4,7 +4,7 @@
 
 // This exercise uses some concepts that we won't get to until later in the course, like `Box` and the
 // `From` trait. It's not important to understand them in detail right now, but you can read ahead if you like.
-// For now, think of the `Box<dyn ???>` type as an "I want anything that does ???" type, which, given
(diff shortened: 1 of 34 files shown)


undali:
Hi @nidhalmessaoudi ,
Does it show correct progress percentages if run as `rustlings verify` ?

nidhalmessaoudi:
Yes, my fix shows correct percentages in both `rustlings verify` and `rustlings watch`. The current working version of rustlings doesn't show correct progress percentages when compilation is failed on watch mode.

nidhalmessaoudi:
Yes, @NANASHI0X74, exactly when you edit an exercise and the compilation fails. This is because the code that updates the percentage field will not be executed when the exercise is not done and that field is first initialized with 0.0. I have fixed that and made this pl.

shadows-withal (approved):

PR #1428, opened 2023-03-15
change order of references in README.md

javihernant:
Changed the order in which the references are presented. Now first reference is "Lifetimes (in Rust By Example)", showing a very brief introduction to the concept of lifetimes in references. Then, "Validating References with Lifetimes", where the whole concept is explained in full details.

The previous order didn't make sense as the second reference didn't contain anything new.

Also, first PR ever. Is there something missing? Should I use another name for my branch (something like "issue#")?

### DIFF
--- a/exercises/lifetimes/README.md
+++ b/exercises/lifetimes/README.md
@@ -13,5 +13,5 @@ restrictive of their callers.
 
 ## Further information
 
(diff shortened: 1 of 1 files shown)


shadows-withal:
All good, thank you for the contribution!

shadows-withal:
@all-contributors please add @javihernant for content

PR #1429, opened 2023-03-16
Update raw_value test's assertion value to meet exercise requirements

ezzhood:
Based on exercise requirement, raw_value test should assert value against Some(0), not Some(5) when input of function is 12.


shadows-withal:
Nope, we use the 24 hour system, meaning `12` is 12PM, which is noon. The expected return value only drops to `0` after 10PM (so 10 hours later).

ezzhood:
Thats why I propose this `assert_eq!(icecreams, 5);` to change to `assert_eq!(icecreams, 0);`

shadows-withal:
But 10PM is when the icecream vanishes, which, if converted to the 24-hour system, is `22`.

PR #1437, opened 2023-03-25
correct comments in errors2.rs

navicore:
trivial text correction - description of the total_cost function is wrong: it calculates total cost but instructions say it calculates token count.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @navicore for content

PR #1457, opened 2023-03-31
Fix (Hash-maps 3): Spelling mistake in Hash-maps 3 hint :crab:

MichaelVoropanov:
Just a bit annoying to me :) :heart: :crab:


shadows-withal:
This is intentional, there is actually a method called `or_insert` that helps here: https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert

PR #1460, opened 2023-04-02
fix: move lifetimes above tests in recommended order

sagarvora:
This order is better if one's working through the exercises in the sequence of "The Book" where lifetimes is covered in Ch. 10.3 whereas tests are covered in Ch. 11.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @sagarvora for content

PR #1463, opened 2023-04-05
feature: task that shows how to deal with panic while testing 

poneciak57:


exercises/tests/tests4.rs
```diff
@@ -0,0 +1,40 @@
+// tests4.rs
+// Correct the tests to
+// Do not change Rectangle::new body
```
shadows-withal, line 3: We don't need this comment in the introductory paragraph.

exercises/tests/tests4.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl Rectangle {
+    pub fn new(width: i32, height: i32) -> Self {
+        if width < 0 || height < 0 {
+            panic!("Rectangle width and height cannot be negative!")
+        }
+        Rectangle {width, height}
+    }
+}
+
```
shadows-withal, line 21: I'd add a comment here stating `Only change the test functions themselves` or something

exercises/tests/tests4.rs
```diff
@@ -0,0 +1,40 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn correct_width_and_height() {
```
shadows-withal, line 27: Is there any better way of expression what we want those tests to do? Maybe add a comment at the beginning of each test saying something like `This test should check if the rectangle is the size that we pass into its constructor`, and same for the other tests.

shadows-withal (changes requested):

exercises/tests/tests4.rs
```diff
@@ -25,16 +25,21 @@ mod tests {
...
-        let _rect = Rectangle::new(10, 10);
+        // This test should check if the rectangle is the size that we pass into its constructor
+        let rect = Rectangle::new(10, 20);
+        assert_eq!(???, 10); // check width
+        assert_eq!(???, 20); // check height
     }
 
     #[test]
     fn negative_width() {
+        // This test should check if thread panics when we try to create rectangle with negative width
```
shadows-withal, line 36: Let's not use "thread" here, instead let's say something like "program"

shadows-withal (approved):

shadows-withal:
@all-contributors please add @poneciak57 for content

PR #1470, opened 2023-04-05
feat: updated generics1

### LINKED ISSUE #1448: generics1: does not require use of explicit generics
The idiomatic solution for `generics1` is to simply remove the explicit type from the variable binding since it can be inferred:
```rust
fn main() {
    let mut shopping_list = Vec::new();
    shopping_list.push("milk");
}
```

The hint says:
```
Vectors in Rust make use of generics to create dynamically sized arrays of any type.
You need to tell the compiler what type we are pushing onto this vector.
```

Two problems with this:

1. The solution is better without specifying a type and thus is not teaching/requiring the explicit use of generics
2. The hint is wrong, because there is not _need_ to tell the compiler the type

poneciak57:
Updated `generics1.rs`.
Now it teaches how to define function that uses generics.

closes #1448

### DIFF
--- a/exercises/generics/generics1.rs
+++ b/exercises/generics/generics1.rs
@@ -1,11 +1,19 @@
-// This shopping list program isn't compiling!
-// Use your knowledge of generics to fix it.
+// generics1.rs
+// Use your knowledge of generics to fix the function signature.
 
 // Execute `rustlings hint generics1` or use the `hint` watch subcommand for a hint.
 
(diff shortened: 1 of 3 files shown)


mo8it:
`generics1` is fixed in the upcoming version 6. But we can think about adding something like this as a third generics exercise.

poneciak57:
ok i will add this as a third one

PR #1471, opened 2023-04-07
docs: update line numbers in move_semantics2

az0977776:


exercises/move_semantics/move_semantics2.rs
```diff
@@ -1,5 +1,5 @@
 // move_semantics2.rs
-// Make me compile without changing line 13 or moving line 10!
+// Make me compile without changing line 17 or moving line 14!
```
poneciak57, line 2: I would delete this comment from introduction paragraph because on line 16 theres another comment saying to not change the line 17 also this kind of comments seems unpractical

jeremymreed: I agree.

If the user removes "I AM NOT DONE", or adds custom code to this exercise, the line numbers could potentially be wrong depending on the exact changes.  This could cause the user confusion in the future if they choose to return to this code in the future.

Looking at the history for this file, it looks like there were extra comments added, and the line numbers were not updated.  See this [commit](https://github.com/rust-lang/rustlings/commit/fef8314d3d45aa1ec3655cd6e9a52f877da51c23).

I'd prefer having the comments directly above the lines the user isn't supposed to change or move.

This would make this exercise less fragile.

shadows-withal (approved):

shadows-withal:
@all-contributors please add @az0977776 for content

PR #1472, opened 2023-04-07
feature: added exercise to "traits" chapter

abhi3700:
This is an attempt to add 1 exercise in "traits" chapter.

The objective is to let the developer learn these:
- "How to implement at function that can return incompatible types?"
- "How to make the same function printable on console?"

Looking forward to getting feedback on this.


poneciak57:
Exercise `traits6` is unnecessarily large, tests seems redundant here.

abhi3700:
> Exercise `traits6` is unnecessarily large, tests seems redundant here.

I can add ?? in tests.

Or I just have only ‘traits7.rs’

poneciak57:
I think u can just delete tests and add main function with simple showcase.

abhi3700:
> I think u can just delete tests and add main function with simple showcase.

ok followed

Issue #1417, opened 2023-03-10
vecs2.rs - Iteration variable names arbitrary and inconsistent


adamhb123:
The names for `i` and `num` in the below code snippets are inexplicitly (former) and inconsistently (in juxtaposition) chosen:
https://github.com/rust-lang/rustlings/blob/5c5cd4c658ebb82f26f88b07660c1fc0fcb6ac14/exercises/vecs/vecs2.rs#L12
https://github.com/rust-lang/rustlings/blob/5c5cd4c658ebb82f26f88b07660c1fc0fcb6ac14/exercises/vecs/vecs2.rs#L23
`i` might lead one to, at first glance, believe we are iterating over the indexes of the elements. I believe changing `i` to `element` would reduce the likelihood of confusion. I also suggest changing `num` to `element` for the sake of consistency.

PR #1473, opened 2023-04-08
docs(vecs2): update hints

Jak-Ch-ll:
In #1417 the variables were renamed to be more expressive, which made the the first hint to become out-of-sync. I removed it completely, as I think the current variable names are conveying that information already.

Additionally, as I was working on the hints already, I added a bit more context for the dereference operator.


shadows-withal:
@all-contributors please add @Jak-Ch-ll for content

Issue #1387, opened 2023-02-19
Add some extra tests for iterators5


plaes:
It would be nice if there are some extra tests in iterators5 that would also check for other values than only `Progress::Complete`.

When I initially implemented my own solution for it, I had accidentally hardcoded this value in my function and did not use the `value` argument.

PR #1478, opened 2023-04-15
fix: Added some extra tests to validate iterators5 solution

Ben2917:
Firstly, I want to say that I've been really impressed with rustlings. This is my first foray into Rust programming and I've really enjoyed running through the exercises over the last few days so I wanted to contribute something in return.

Based on [this issue](https://github.com/rust-lang/rustlings/issues/1387), I've added some more detailed tests to the iterators5 exercise as I think the author has raised a valid point.

I've added new tests to validate that the output is correct for all progress types and I've also updated the tests that compare against the equivalent functions (that use for loops) to ensure that the output is the same for each type. 

Please let me know if there are any edits you would like me to make. Thanks!

### DIFF
--- a/exercises/iterators/iterators5.rs
+++ b/exercises/iterators/iterators5.rs
@@ -77,7 +77,7 @@ mod tests {
     }
 
     #[test]
-    fn count_complete_equals_for() {       
+    fn count_complete_equals_for() {
         let map = get_map();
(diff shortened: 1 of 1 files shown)


shadows-withal (approved):

shadows-withal:
@all-contributors please add @Ben2917 for content

PR #1484, opened 2023-04-20
fix(move_semantics2): fix line number comment

akgerber:
Commit fef8314 added three lines of comments, which left the line numbers expected to stay unchanged mentioned on line 2 out of date.

### DIFF
--- a/exercises/move_semantics/move_semantics2.rs
+++ b/exercises/move_semantics/move_semantics2.rs
@@ -1,5 +1,5 @@
 // move_semantics2.rs
-// Make me compile without changing line 13 or moving line 10!
+// Make me compile without changing line 17 or moving line 14!
 // Execute `rustlings hint move_semantics2` or use the `hint` watch subcommand for a hint.
 
 // Expected output:


akgerber:
Also note that techniques 2 and 3 suggested in the hint do not produce the expected output per fef8314:
```
So, `vec0` is passed into the `fill_vec` function as an argument. In Rust,
when an argument is passed to a function and it's not explicitly returned,
you can't use the original variable anymore. We call this "moving" a variable.
Variables that are moved into a function (or block scope) and aren't explicitly
returned get "dropped" at the end of that function. This is also what happens here.
There's a few ways to fix this, try them all if you want:
1. Make another, separate version of the data that's in `vec0` and pass that
   to `fill_vec` instead.
2. Make `fill_vec` borrow its argument instead of taking ownership of it,
   and then copy the data within the function in order to return an owned
   `Vec<i32>`
3. Make `fill_vec` *mutably* borrow a reference to its argument (which will need to be
   mutable), modify it directly, then not return anything. Then you can get rid
   of `vec1` entirely -- note that this will change what gets printed by the
   first `println!
```

shadows-withal (approved):

shadows-withal:
@all-contributors please add @akgerber for content

PR #1490, opened 2023-04-21
modules2.rs: add more information to hint

deedy5:
Add link to get more information
```https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#re-exporting-names-with-pub-use```


shadows-withal:
@all-contributors please add @deedy5 for content

PR #1504, opened 2023-05-01
feat(main, verify): Show hint(s) after exercise completion

rb5014:
Hello, I added an option in main.rs that we can use like this: 'rustlings --success-hints watch' or 'rustlings --success-hints verify'.
I had to change some functions in verify.rs to use this option as an additional boolean parameter, to pass it in the end to the prompt_completion function. In this function I added some println!() lines to display the hint(exercise.hint) in the same way the output is displayed too.
If the user uses this option with other subcommands like run, or lsp, nothing changes, no errors, I didn't handle these cases.
I also updated the help message for this.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,9 @@ struct Args {
     /// show the executable version
     #[argh(switch, short = 'v')]
     version: bool,
+    /// show hints on success
+    #[argh(switch)]
+    success_hints: bool,
     #[argh(subcommand)]
     nested: Option<Subcommands>,
 }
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -37,6 +37,9 @@ struct Args {
     /// show the executable version
     #[argh(switch, short = 'v')]
     version: bool,
+    /// show hints on success
+    #[argh(switch)]
+    success_hints: bool,
```
shadows-withal, line 42: This should be added to `WatchArgs` instead, because while this works, you'd need to type `rustlings --success-hints watch` to make it work, `rustlings watch --success-hints` is way more ergonomic.

rb5014: Thank you for your feedback, I added it to WatchArgs instead of Args. It works as you said. Now previously I put this option for the verify command too, because I modified the parameters of the function verify, but here I put false for the success_hints parameter when the command verify is called.

shadows-withal:
@all-contributors please add @rb5014 for content

PR #1512, opened 2023-05-09
fix(options3): panic when not matching to avoid false positives

### LINKED ISSUE #1503: options3.rs and similar tasks should panic when there is a logic error to avoid false positives
In the options3 task, you just have a println if there is no match:

```rust
let y: Option<Point> = Some(Point { x: 100, y: 200 });

match y {
    Some(p) => println!("Co-ordinates are {},{} ", p.x, p.y),
    _ => println!("no match"),
}
```

For this and similar tests, it should instead panic when matched incorrectly in case the user made a logic error. Otherwise, it would compile and show green lights unless they actually read the output, which I just skim most times, to be honest.

```diff
-    _ => println!("no match"),
+    _ => panic!("no match!"),
```

I'd be willing to make this contribution.

nmay231:
Closes #1503

I searched through the other `match` statements and did not really find any similar situations, so I only changed this file.

### DIFF
--- a/exercises/options/options3.rs
+++ b/exercises/options/options3.rs
@@ -13,7 +13,7 @@ fn main() {
 
     match y {
         Some(p) => println!("Co-ordinates are {},{} ", p.x, p.y),
-        _ => println!("no match"),
+        _ => panic!("no match!"),
     }
     y; // Fix without deleting this line.
 }


shadows-withal (approved):

shadows-withal:
@all-contributors please add @nmay231 for content

PR #1523, opened 2023-05-20
feat(iterators5): remove outdated part of hint

b1ue64:
The hint currently tells you to add `return 0` to the `count_collection_iterator` to make it compile. This is not necessary because the `todo!()` macro is used.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @b1ue64 for content

PR #1540, opened 2023-05-31
chore: update hint for vecs2 

tajo48:
It needs an update to match with exercise as it was recently updated to fix clarity.
https://github.com/rust-lang/rustlings/commit/7bab78c66d4c0fff4b9f7ca082964353215265e3 <- in this commit

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -260,8 +260,8 @@ name = "vecs2"
 path = "exercises/vecs/vecs2.rs"
 mode = "test"
 hint = """
-Hint 1: `i` is each element from the Vec as they are being iterated. Can you try
-multiplying this?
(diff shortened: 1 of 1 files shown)


shadows-withal:
Ah, didn't realize there were two PRs for this - it's fixed as per https://github.com/rust-lang/rustlings/pull/1544.

tajo48:
@shadows-withal 
I have seen pull request #1544, and I must admit I am somewhat disappointed with the phrase "`element` is each element" as it could be formulated differently for clarity. That's why I have decided that the entire sentence should be rewritten from scratch.
Imo for clarity we should try to avoid redundant statements whenever possible.

PR #1542, opened 2023-06-03
fix(threads, smart pointers): Swap order of threads and smart pointers exercises

### LINKED ISSUE #1541: Put the `smart_pointers` exercise before the `threads` exercise
While learning threading, you need to learn about the `Arc` type and why it is needed. Worse, there are no explicit exercises about the `Arc` type in the `thread` exercises. Immediately after the threading exercises, we do the `smart_pointers` exercises which do teach you about `Cow`, `Rc` and `Arc` types nicely with no prior dependencies on the threading exercise. 

I suggest that we swap the order of the `threads` and the `smart_pointers` exercises so help the concepts flow better.

luhem7:
closes #1541

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -905,67 +905,6 @@ The fold method can be useful in the count_collection_iterator function.
 For a further challenge, consult the documentation for Iterator to find
 a different method that could make your code more compact than using fold."""
 
-# THREADS
-
-[[exercises]]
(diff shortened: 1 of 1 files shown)


shadows-withal:
@all-contributors please add @luhem7 for content

PR #1543, opened 2023-06-04
Rustlings depends on git for rustlings reset commands

alex800121:
Rustlings depends on git for rustlings reset commands.
Add "nativeBuildInputs  = [ pkgs.git ];" to avoid unexpected behaviors when running bulid tests.


shadows-withal:
I think this is superseded by #1653? Let me know if I'm wrong.

PR #1549, opened 2023-06-08
fix(enums3): add test for message

### LINKED ISSUE #1548: Enums3 is not fully tested
I've the following code in my IDE and it is passing the tests but I've not correctly implemented my code.

> Note: I've not match'd on move but there is no test catching that!

```rust
// enums3.rs
// Address all the TODOs to make the tests pass!
// Execute `rustlings hint enums3` or use the `hint` watch subcommand for a hint.

// I AM NOT DONE

enum Message {
    // TODO: implement the message variant types based on their usage below
    Move(Point),
    Echo(String),
    ChangeColor(u8, u8, u8),
    Quit,
}

struct Point {
    x: u8,
    y: u8,
}

struct State {
    color: (u8, u8, u8),
    position: Point,
    quit: bool,
}

impl State {
    fn change_color(&mut self, color: (u8, u8, u8)) {
        self.color = color;
    }

    fn quit(&mut self) {
        self.quit = true;
    }

    fn echo(&self, s: String) {
        println!("{}", s);
    }

    fn move_position(&mut self, p: Point) {
        self.position = p;
    }

    fn process(&mut self, message: Message) {
        // TODO: create a match expression to process the different message variants
        // Remember: When passing a tuple as a function argument, you'll need extra parentheses: fn function((t, u, p, l, e))
        match message {
            Message::Move(point) => {self.move_position(point)},
            Message::Echo(string) => {},
            Message::ChangeColor(x, y, z) => { self.change_color((x, y, z))},
            Message::Quit => { self.quit() },
        }
    }
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_match_message_call() {
        let mut state = State {
            quit: false,
            position: Point { x: 0, y: 0 },
            color: (0, 0, 0),
        };
        state.process(Message::ChangeColor(255, 0, 255));
        state.process(Message::Echo(String::from("hello world")));
        state.process(Message::Move(Point { x: 10, y: 15 }));
        state.process(Message::Quit);

        assert_eq!(state.color, (255, 0, 255));
        assert_eq!(state.position.x, 10);
        assert_eq!(state.position.y, 15);
        assert_eq!(state.quit, true);
    }
}

```

IVIURRAY:
closes #1548

I've changed the exercie slightly so that all code paths are tested. Even the Echo code path asserts the message. In my issue I managed to pass the test without implementing anything for the `Message::Echo` match logic.

```rust
match message {
            Message::Move(point) => {self.move_position(point)},
            Message::Echo(string) => {}, // I pass the exercie without implementing this.
            Message::ChangeColor(x, y, z) => { self.change_color((x, y, z))},
            Message::Quit => { self.quit() },
        }
```

### DIFF
--- a/exercises/enums/enums3.rs
+++ b/exercises/enums/enums3.rs
@@ -17,6 +17,7 @@ struct State {
     color: (u8, u8, u8),
     position: Point,
     quit: bool,
+    message: String
 }
 
 impl State {
@@ -28,9 +29,7 @@ impl State {
         self.quit = true;
     }
 
-    fn echo(&self, s: String) {
-        println!("{}", s);
-    }
+    fn echo(&mut self, s: String) { self.message = s }
 
     fn move_position(&mut self, p: Point) {
         self.position = p;
@@ -52,6 +51,7 @@ mod tests {
             quit: false,
             position: Point { x: 0, y: 0 },
             color: (0, 0, 0),
+            message: "hello world".to_string(),
         };
         state.process(Message::ChangeColor(255, 0, 255));
         state.process(Message::Echo(String::from("hello world")));
@@ -62,5 +62,6 @@ mod tests {
         assert_eq!(state.position.x, 10);
         assert_eq!(state.position.y, 15);
         assert_eq!(state.quit, true);
+        assert_eq!(state.message, "hello world");
     }
 }


shadows-withal (approved):

shadows-withal:
@all-contributors please add @IVIURRAY for content

PR #1435, opened 2023-03-24
cargo update to resolve #1430

eugkhp:
I encountered the same issue as #1430 by trying to launch rustlings with nix. I'm using Apple Silicon that might be the reason, simple `cargo update` helped. Probably the rustling version update is required to account for the environment update, right? 

But the core issue seems to be that nix configuration should be in sync with cargo, which breaks the entire idea of nix. I would be glad to help if you think this is something worth doing

Issue #1430, opened 2023-03-17
`nix develop` is returning a Cargo.lock error during installation


tahoeschrader:
I'm kind of a nix newbie so I could be doing something wrong, but I'm running into an issue while following the install instructions in the readme for rustlings via nix. 

The version of rustlings I checked out from github is 5.4.1. I am on MacOS 13.1 (22C65), Apple M2 Chip.

`nix --version` returns:
```
nix (Nix) 2.13.0
```
`nix develop` returns:
```
error: builder for '/nix/store/77n03jiaaxk100dgc5880iq56alwzb80-rustlings.drv' failed with exit code 101;
       last 10 log lines:
       > Executing cargoSetupPostPatchHook
       > Validating consistency between /private/tmp/nix-build-rustlings.drv-0/source/Cargo.lock and /private/tmp/nix-build-rustlings.drv-0/cargo-vendor-dir/Cargo.lock
       > Finished cargoSetupPostPatchHook
       > updateAutotoolsGnuConfigScriptsPhase
       > configuring
       > building
       > Executing cargoBuildHook
       > ++ env CC_aarch64-apple-darwin=/nix/store/a79iskrr56f0dbggj1gs819y3g3vs37k-clang-wrapper-11.1.0/bin/cc CXX_aarch64-apple-darwin=/nix/store/a79iskrr56f0dbggj1gs819y3g3vs37k-clang-wrapper-11.1.0/bin/c++ CC_aarch64-apple-darwin=/nix/store/a79iskrr56f0dbggj1gs819y3g3vs37k-clang-wrapper-11.1.0/bin/cc CXX_aarch64-apple-darwin=/nix/store/a79iskrr56f0dbggj1gs819y3g3vs37k-clang-wrapper-11.1.0/bin/c++ cargo build -j 8 --target aarch64-apple-darwin --frozen --release
       > error: the lock file /private/tmp/nix-build-rustlings.drv-0/source/Cargo.lock needs to be updated but --frozen was passed to prevent this
       > If you want to try to generate the lock file without accessing the network, remove the --frozen flag and use --offline instead.
       For full logs, run 'nix log /nix/store/77n03jiaaxk100dgc5880iq56alwzb80-rustlings.drv'.
error: 1 dependencies of derivation '/nix/store/5rq3x7c1mrsppa3w9cgfpijsavzqck2f-nix-shell-env.drv' failed to build
```
`ls -la` returns:
```
total 576
[...]

PR #1550, opened 2023-06-11
chore(ci): add action step to check if lockfile is synced

proofconstruction:
This PR prevents issues like #1430 from reoccurring.

---
[Issue #1430](https://github.com/rust-lang/rustlings/issues/1430) was caused by `Cargo.lock` being out of sync with `Cargo.toml`; in [the commit](https://github.com/rust-lang/rustlings/tree/9acefe8b3a2bed6e23dc31e9c1d5e9c5f6ab7596) preceding the merge of [PR #1435](https://github.com/rust-lang/rustlings/pull/1435), [`Cargo.toml`](https://github.com/rust-lang/rustlings/blob/568767601427cccf11ef8ea0dd20ad1ac0dbee1e/Cargo.toml#L3) had rustlings at version 5.4.1 and [`Cargo.lock`](https://github.com/rust-lang/rustlings/blob/568767601427cccf11ef8ea0dd20ad1ac0dbee1e/Cargo.lock#L462) had rustlings at 5.4.0.

Under "normal" (non-`nix`) conditions, unless the `--frozen` or `--locked` flags are passed to `cargo` calls, `cargo` will silently update `Cargo.lock` before continuing; `nix` doesn't like this however, and complains when running `nix develop`:
```
> error: the lock file /private/tmp/nix-build-rustlings.drv-0/source/Cargo.lock needs to be updated but --frozen was passed to prevent this
> If you want to try to generate the lock file without accessing the network, remove the --frozen flag and use --offline instead.
```
This happens because `nix develop` tries to build the [`devShell`](https://github.com/rust-lang/rustlings/blob/main/flake.nix#L51), which requires first building [the `rustlings` package defined in `flake.nix`](https://github.com/rust-lang/rustlings/blob/main/flake.nix#L22), and `rustPlatform.buildRustPackage` calls `cargoBuildHook` which [passes the `--frozen` flag to `cargo build`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/hooks/cargo-build-hook.sh#L36).

This small patch adds a fetch step before the build and test steps, which runs `cargo fetch --locked`, updating `Cargo.lock` if necessary before pulling dependencies (but not building them). If `Cargo.lock` was updated, `cargo fetch --locked` exits with a 101 status, failing the workflow. If not (so, when the manifest and lock file are in sync), dependencies are already downloaded and available for the following `cargo build` step, so this doesn't add additional time to check.

Screenshot 2023-06-11 at 17 03 24

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - name: Fetch & maybe update Cargo.lock
+      run: cargo fetch --locked
     - name: Build
       run: cargo build --verbose
     - name: Run tests


shadows-withal (approved):
Thank you!

shadows-withal:
@all-contributors please add @proofconstruction for build

shadows-withal:
@all-contributors please add @proofconstruction for infra

PR #1551, opened 2023-06-11
fix: update threads2.rs to use Mutex

alikiki:
threads2.rs doesn't use Mutex, which I think is necessary here.


shadows-withal:
Correct, but doing this is part of the exercise, which is also pointed out in the hint.

PR #1565, opened 2023-06-16
fix(enums3): modify message string in test

pksadiq:
Otherwise it won't actually test the change of state.message and it would fail to check if the user missed changing state.message

This happened to me as I had a catch-all line in `match`


shadows-withal:
@all-contributors please add @pksadiq for content

PR #1572, opened 2023-06-22
fix: invalid function syntax in quiz1.rs

Creative-Difficulty:


shadows-withal:
I don't think this is needed, the user would need to pass function parameters anyways.

PR #1573, opened 2023-06-23
refactor: Update comment to use correct construct name 'Result' inste…

novanish:
Refactor the comment in the code to provide a more accurate description of the construct being used. Replace the mention of **Option** with **Result**.


shadows-withal:
@all-contributors please add @novanish for content

PR #1535, opened 2023-05-29
docs: cleanup the explanation paragraphs at the start of each exercise.

robertefry:
The paragraphs at the start of each exercise have been inconsistent in line length, and spacing. This pull request aims to fix that to make the exercises easier to read and understand.

PR #1587, opened 2023-07-02
chore: update line reference in strings2 hint

willhack:
Resolves a side-effect of PR #1535. In making the boilerplate consistent, [several new lines](https://github.com/rust-lang/rustlings/commit/7eef5d15eef780f93e22b1b4e0185f7708219ea0#diff-0eb1285ec18690b4824b04771d7b390176ef133a218d6b45c614e947fe2b3b0f) were added - misaligning the hint.

The new offending line is [line 12](https://github.com/rust-lang/rustlings/blob/main/exercises/strings/strings2.rs#L12). That being said, it might be more stable to just remove the line reference altogether.
> _There is a way to add one character ~~to line 12~~, though, that will coerce the `String` into a string slice._

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -438,8 +438,8 @@ path = "exercises/strings/strings2.rs"
 mode = "compile"
 hint = """
 Yes, it would be really easy to fix this by just changing the value bound to `word` to be a
-string slice instead of a `String`, wouldn't it?? There is a way to add one character to line
-12, though, that will coerce the `String` into a string slice.
+string slice instead of a `String`, wouldn't it?? There is a way to add one character to the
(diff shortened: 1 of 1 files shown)


shadows-withal:
Let's go with your suggestion of removing the line reference altogether. Maybe to make it generic, something like this: "There is a way to add one character to the if statement, though, that will coerce the String into a string slice."

willhack:
> Let's go with your suggestion of removing the line reference altogether. Maybe to make it generic, something like this: "There is a way to add one character to the if statement, though, that will coerce the String into a string slice."

You got it, updated!

shadows-withal (approved):

PR #1588, opened 2023-07-03
added if3 based on: `Using if in a let Statement`

jrcarl624:
This highlights on the let assignment syntax with if statements. The error is related to inconsistent primitive types in the arms of the if tree.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @jrcarl624 for content

PR #1591, opened 2023-07-06
vecs1: add hint for to_vec()

RogerJL:
It actually only case 2 that fulfils the excersise instruction
to use macro...

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -246,11 +246,13 @@ name = "vecs1"
 path = "exercises/vecs/vecs1.rs"
 mode = "test"
 hint = """
-In Rust, there are two ways to define a Vector.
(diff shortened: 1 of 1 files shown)


shadows-withal:
hmm, I disagree, `to_vec` doesn't _define_ a vec, it _copies_ into a vec. I don't think that's the solution we should hint at here, but it could be something we encourage users to try out in addition (something like "By the way, there's also a way to convert the array into a Vec directly, without having to specify its values again. Can you find out what it is?"

RogerJL:
But `to_vec` also define a vector...
The vector did not exist before this method is called.

mo8it:
Closing because I don't think we should recommend `.to_vec()` in this case :)

PR #1592, opened 2023-07-06
add array code exerciese to exerises

leikang123:
Code learning case of adding arrays and vectors to the project。


shadows-withal:
We already have vec exercises in-tree.

PR #1598, opened 2023-07-17
fix(primitives-4.rs): update hint so it's less confusing

yamila-moreno:
Hi there, first day at rustlings, enjoying the learning system. Thanks!!

This is a small PR to fix what (in my opinion) is a bit confusing hint.

Current hint:
>  use the starting and ending indices of the items in the Array

Problem:
In the documentation it's explained that is ending `PLUS ONE` (which is also expected I guess). I tried the hint literally in the excercise and it wasn't working.

I hope this is useful :)


shadows-withal (approved):

shadows-withal:
@all-contributors please add @yamila-moreno for content

PR #1599, opened 2023-07-20
add test-case to if/if1 to check equal values

szabgab:


shadows-withal (approved):

shadows-withal:
@all-contributors please add @szabgab for content

PR #1607, opened 2023-07-28
chore(errors4): improved comment

softarn:
I was a bit confused about this comment since I thought there was supposed to be something more than the Ok value there. But the problem is that the Ok value is always returned. Switching only -> always to signify that better.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @softarn for content

PR #1610, opened 2023-08-01
Updated if3.rs: Added a note that states that the test does not need to be changed.

jrcarl624:


exercises/if/if3.rs
```diff
@@ -28,7 +28,7 @@ pub fn animal_habitat(animal: &str) -> &'static str {
 
     habitat
 }
-
+// No test changes needed.
```
shadows-withal, line 31: nit, maybe one extra empty line above this?

PR #1624, opened 2023-08-14
Fix comment in errors2

shirts:
Fixes a trivial comment that references `parse` function, but that function does not exist. It exists as `total_cost`


shadows-withal (approved):

shadows-withal:
@all-contributors please add @shirts for content

PR #1631, opened 2023-08-24
my attempt at the rustlings quizes to learn rust, kept for reference

mgh100:


mgh100:
sorry I pushed to the wrong place by accident :see_no_evil:, amazing resource to learn with though, thank you!

PR #1632, opened 2023-08-25
fix(errors1): change Result to Option in exersise description

eLVas:
The way this description war written implied that similar but else than `Result` should be used. Which is not correct according to hint and provided tests. I have updated it to say that something like `Option` can be used to solve the exercise. 

If the intention was to mention `Result` explicitly to make the exercise easier, then I would phrase it like this:
> Rust has a similar construct to `Option` - `Result`, that can be used to express error conditions.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @eLVas for content

PR #1633, opened 2023-08-25
Port to Clap

mo8it:
The arguments from the PR https://github.com/rust-lang/rustlings/pull/715 that replaced Clap with argh are not valid anymore.

### Clap is heavy

Currently, this branch has only 7 more dependencies. The difference in compilation time is not significant.

Compiling the `main` branch:
```console
$ hyperfine -- "rm -r target && cargo b -r"
Benchmark 1: rm -r target && cargo b -r
  Time (mean ± σ):     11.496 s ±  0.206 s    [User: 89.523 s, System: 5.787 s]
  Range (min … max):   11.161 s … 11.822 s    10 runs
```

Compiling this `clap` branch:
```console
$ hyperfine -- "rm -r target && cargo b -r"
Benchmark 1: rm -r target && cargo b -r
  Time (mean ± σ):     13.415 s ±  0.290 s    [User: 115.880 s, System: 6.849 s]
  Range (min … max):   12.859 s … 13.865 s    10 runs
```

The benchmarks show that the compilation time increased by only 15% on my system. But remember, we are talking about just 2 additional seconds!

If you really want to save an additional second of compilation time, then you could disable the default features `color` and `suggestions` for example. But I don't think that it is worth it since the compilation is done only once!

### argh makes the code easier to read

This argument is only valid for version 2 of Clap. In this PR, version  4 has been used.

Currently, the opposite is true. As you can see in the git diff, the code has less boilerplate now since the Derive API of Clap has been used.

## Arguments for Clap

Clap is used by most Rust CLI projects and the community is used to its API. There is a huge ecosystem around it.

The argument that the code has less boilerplate with Clap's Derive API was mentioned above.

Clap has additional features like suggestions and colors:

image

It has better highlighting (above: argh, below: Clap):

image

## Clap vs bpaf?

Please see the another PR for `bpaf` first: https://github.com/rust-lang/rustlings/pull/1634

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,25 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.4"
+version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6748e8def348ed4d14996fa801f4122cd763fff530258cdc03f64b25f89d3a5a"
+checksum = "0c378d78423fdad8089616f827526ee33c19f2fddbd5de1629152c9593ba4783"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "anstream"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f58811cfac344940f1a400b6e6231ce35171f614f26439e80f8c1465c5cc0c"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "utf8parse",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


shadows-withal:
Thanks! I agree with your rationale, I've also been using modern Clap version recently and they're a _lot_ better than they were 2 years ago. I'll review this properly tomorrow, when I'm also planning on pushing a release.

mo8it:
I am working on a temporary fork for my Rust course next week: https://codeberg.org/mo8it/collective-rustlings

I have changes that I want to push upstream. Hint fixes, dependencies upgrades, etc. I will make those tiny PRs until tomorrow then.

You will receive more upstream love at the end of my course ❤️

shadows-withal (approved):
Hey! Sorry for the delay, I had some life stuff pop up. Happy to land this now!

PR #1634, opened 2023-08-26
Port to Bpaf

mo8it:
After porting to Clap in https://github.com/rust-lang/rustlings/pull/1633, @pacak asked me on Mastodon to try `bpaf`.

Here is the result:

image

## Compile time

Currently, `bpaf` has two dependencies less than Clap. The difference is not significant.

The benchmark:

```console
$ hyperfine -- "rm -r target && cargo b -r"
Benchmark 1: rm -r target && cargo b -r
  Time (mean ± σ):     12.431 s ±  0.164 s    [User: 97.754 s, System: 6.167 s]
  Range (min … max):   12.085 s … 12.616 s    10 runs
```

`bpaf` has 8% additional compile time instead of the 15% of Clap. But remember, we are talking about one additional second instead of two of a compilation that is done only once!

## Which to choose?

Personally, I like the output of `bpaf` more since it has colors, but if you don't like the colors and want only bold and underline highlightings, you can use the `bpaf` feature `dull-color` instead of `bright-color`.

The code quality is very similar to Clap. It doesn't have the boilerplate of `argh`.

I don't like that the `Subcommand` enum has to implement `Clone`. @pacak said that it is required to be able to run the parser multiple times, but I still think that it could be improved.

I will let you decide which one to merge :)

Personally, I will go with `bpaf` for my temporary fork for my Rust course. Just for the nice colors 🌈

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,34 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "argh"
-version = "0.1.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab257697eb9496bf75526f0217b5ed64636a9cfafa78b8365c71bd283fcef93e"
-dependencies = [
- "argh_derive",
- "argh_shared",
-]
-
-[[package]]
-name = "argh_derive"
-version = "0.1.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b382dbd3288e053331f03399e1db106c9fb0d8562ad62cb04859ae926f324fa6"
-dependencies = [
- "argh_shared",
- "proc-macro2",
- "quote",
- "syn 1.0.109",
-]
-
-[[package]]
-name = "argh_shared"
-version = "0.1.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "64cb94155d965e3d37ffbbe7cc5b82c3dd79dd33bd48e536f73d2cfb8d85506f"
-
-[[package]]
 name = "assert_cmd"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


pacak:
Outside of the scope of porting specifically but there are a few things that can be improved:
1. normalize data the app gets a bit more. Consider `--solved` and `--unsolved` flags in list. They are mutually exclusive and passing both doesn't really make much sense. From the app point of view it might make sense to represent this filter as something like this:
   ```rust
    #[derive(Debug, Clone, Bpaf)]
    #[bpaf(fallback(SolvedFilter::Everything))]
    enum SolvedFilter {
        /// Show solevd only
        Solved,
        /// Show unsolved only
        Unsolved,
        #[bpaf(skip)]
        Everything,
    }
    ```
    User will still get to pass either of `--solved` or `--unsolved`, but not both, the app will get a nice enum.
    Or `--nocapture` feels strange sitting at the top level.

2. Generate and include command reference - something like here https://crates.io/crates/cargo-show-asm - basically help for all the commands accessible without having to run anything
3. Generate and include shell completion at least with nix. And add instructions for users how to add them if they are not using nix. Mostly for conveniece, too lazy to type :)

I can help with some or all those items.

epage:
imo bpaf isn't ready for introductory material (yet), see https://github.com/nicoburns/blessed-rs/issues/90

I'm hoping we can work together to get it there.

epage:
imo bpaf isn't ready for introductory material (yet), see https://github.com/nicoburns/blessed-rs/issues/90

I'm hoping we can work together to get it there though

mo8it:
After publishing this second PR, I started a vote on Mastodon about which version people prefer:
https://fosstodon.org/@mo8it/110955677826536002

Looks like Clap is winning after at least 16 votes. But the decision is left to the maintainer @shadows-withal

shadows-withal:
I would also tend more towards using Clap, for the reasons discussed, but also I believe that it's easier to reason about, which improves the contributor experience.

shadows-withal:
Closing as the clap PR is landed

PR #1635, opened 2023-08-26
Formatted readmes so as not to exceed a line length of 120 characters

linguini1:
Although some IDEs come with markdown renderer, other's don't. I used Neovim to complete the rustlings exercises and preferred to just look at the raw markdown file in my editor alongside the exercise, instead of rendering it in the browser.

I noticed that some files had short line lengths that could fit on screen, and others were very long. This PR just reformats the readmes to 120 character long lines so that it can fit on a reasonably sized screen without having to use a markdown renderer.

Hopefully others who use text editors or don't want to render the markdown can find this useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,21 +4,25 @@
 
 </div>
 
-Greetings and welcome to `rustlings`. This project contains small exercises to get you used to reading and writing Rust code. This includes reading and responding to compiler messages!
+Greetings and welcome to `rustlings`. This project contains small exercises to get you used to reading and writing Rust
+code. This includes reading and responding to compiler messages!
 
 _...looking for the old, web-based version of Rustlings? Try [here](https://github.com/rust-lang/rustlings/tree/rustlings-1)_
 
 Alternatively, for a first-time Rust learner, there are several other resources:
 
-- [The Book](https://doc.rust-lang.org/book/index.html) - The most comprehensive resource for learning Rust, but a bit theoretical sometimes. You will be using this along with Rustlings!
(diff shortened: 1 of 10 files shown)


shadows-withal:
This is a good effort, I think, it just places a non-insignificant maintainer burden by requiring me to keep the lines properly wrapped in the future (lest we have to open more PRs like this). I'm 50/50 on whether I want to merge this as a result of this, I think it might be more "reliable" to rely on the user having soft wrapping enabled in their editor. What do you think?

x10an14:
I also support soft-wrap. Don't let one person's choice of tooling affect others rustlings experience.

Also, a different compromise would be to always start sentences on a new line. Only popular markdown parser which does not keep paragraphs together correctly is Github (afaik).

And how often do any of us intend to read the Readmes of this specific repo on the website?  Very few, I wager.

Finally, an edit as proposed does complicate git history. It'd be nice to be able to easily find author of some documentation, so as to keeping the possibility of contacting them for clarification more easily accessible.

linguini1:
> This is a good effort, I think, it just places a non-insignificant maintainer burden by requiring me to keep the lines properly wrapped in the future (lest we have to open more PRs like this). I'm 50/50 on whether I want to merge this as a result of this, I think it might be more "reliable" to rely on the user having soft wrapping enabled in their editor. What do you think?

Absolutely valid concern that I failed to consider. Users like myself who are viewing the READMEs in a text editor with no line-wrap/renderer are likely a very small minority, and keeping up with the line length wrapping is definitely more work than it's worth unless it was automated. I think I would agree with the decision not to merge this PR.

PR #1639, opened 2023-08-26
Update deps

mo8it:
Nothing breaking, only small changes from `indicatif`.

`notify` is a hard one though since it removed debounced events and one has to use an additional crate etc.
Therefore, I left it on version 4. I opened an issue for it: https://github.com/rust-lang/rustlings/issues/1640#issue-1868275205

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,24 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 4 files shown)


shadows-withal (approved):

shadows-withal:
@all-contributors please add @mo8it for code

PR #1641, opened 2023-08-26
Convert exercises with assertions to tests

mo8it:
Converted all exercises having assertions to test mode.

I am not really sure about all implications of this conversion. Please review it first. You might think that some of them should have both a main function and a test.

Should the function names be changed? `main` is not a good test function name. I can use other names if you want.

### DIFF
--- a/exercises/iterators/iterators1.rs
+++ b/exercises/iterators/iterators1.rs
@@ -11,6 +11,7 @@
 
 // I AM NOT DONE
 
+#[test]
 fn main() {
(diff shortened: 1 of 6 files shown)


shadows-withal (approved):
yep, I agree with this rationale

shadows-withal:
that being said, I think it's fine to keep `main` as a test name

mo8it:
Alright, the PR would be ready for merging then after your review, right? :D

shadows-withal:
Oh yeah sure! Let me do that.

PR #1647, opened 2023-08-29
Add an array exercise

leikang123:
**add array3.rs and array4.rs exercise**


shadows-withal:
I don't think these are a good fit for the project, sorry!

PR #1650, opened 2023-08-30
feat(flake): Add defaults to commands in flake

x10an14:
So that:
- `nix build .#`, and
- `nix run .#` both work.


shadows-withal (approved):
(i'm rubber-stamping all of these, i trust that you know what you're doing :wink:)

PR #1651, opened 2023-08-30
improvement(ci): Now github should also check if the nix dev-env/buil…

x10an14:
…d breaks

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,24 @@
+name: Ensure nix-dev environment still builds correctly
+
+on:
+  # On any push or PR
+  push:
+  pull_request:
+
+jobs:
+  nix-dev-shell:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: nixbuild/nix-quick-install-action@v25
+    - run: nix build --version
+    - run: nix build .#devShell
+
+  nix-build-and-run-rustlings:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: nixbuild/nix-quick-install-action@v25
+    - run: nix build --version
+    - name: Build & Run
+      run: nix run .#rustlings -- --help
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,4 +1,4 @@
-name: Ensure nix-dev environment still builds correctly
+name: Verify nix environment
 
 on:
   # On any push or PR


.github/workflows/nix.yml
```diff
@@ -0,0 +1,24 @@
+name: Ensure nix-dev environment still builds correctly
```
shadows-withal, line 1: maybe shorten the name so that it doesn't overflow in the interface?

x10an14: Sure, got a suggestion I can click "add" on here in review comment? I'm currently away from pc (on smartphone device you see).

shadows-withal:
```suggestion
name: Verify nix environment
```

x10an14:
Needs some debugging/fixing as to why the x86_64-linux experience on multiple private machines, break on the Ubuntu Github Actions build envs.

Any help from others to test the nix commands that fail is very welcome.

markgreene74:
hi @x10an14 I gave your workflow a try, and this is what I found:

## nix-build-and-run-rustlings

The [`nix-build-and-run-rustlings`](https://github.com/rust-lang/rustlings/actions/runs/6073126082/job/16474440782?pr=1651#logs) job seems to fail because of the `rustc` version.

Error:

```
> error: package `clap_derive v4.4.2` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.64.0
```

The error does not happen if I try a new build. I guess the `rustc` version was maybe bumped up in the Nix build.

Here is the link to a successful run with the `--debug` flag: [`nix-build-and-run-rustlings`](https://github.com/markgreene74/rustlings/actions/runs/6597960093/job/17925458202#logs)

```
copying path '/nix/store/66a3ip9miaimi7gyhl6jdzbjchp76ccv-rustc-1.72.0' from 'https://cache.nixos.org/'...
locking path '/nix/store/66a3ip9miaimi7gyhl6jdzbjchp76ccv-rustc-1.72.0'
lock acquired on '/nix/store/66a3ip9miaimi7gyhl6jdzbjchp76ccv-rustc-1.72.0.lock'
```

So this seems OK now.

## nix-dev-shell

The [nix-dev-shell](https://github.com/rust-lang/rustlings/actions/runs/6073126082/job/16474440394?pr=1651#logs) job fails with this error:

```
error: flake output attribute 'devShell' is not a derivation or path
```

I am not very good with Nix but looking at the `flake.nix` file and the Flake docs [here](https://nixos.wiki/wiki/Flakes#Output_schema) and [here](https://nixos.wiki/wiki/Flakes#Importing_packages_from_multiple_channels) the error makes sense, as it doesn't look like `devShell` is defined as `packages."<system>"."<name>"`.

If I replace `nix build .#devShell` with `nix build .#rustlings-app` instead I get this different error:

```
error: flake 'git+file:///home/runner/work/rustlings/rustlings?shallow=1' does not provide attribute 'packages.x86_64-linux.rustlings-app', 'legacyPackages.x86_64-linux.rustlings-app' or 'rustlings-app'
```

I did some additional testing in this branch if anybody is interested in seeing the results: https://github.com/markgreene74/rustlings/tree/test/x10an14-fork-add-nix-github-workflow

mo8it:
Closing because of missing response. Things will change in v6 anyway so we probably need something else then.

x10an14:
Apologies for missing the message you expected a response to. What are the plans for v6? Where/what issue/PR can I read up on them?

Issue #1596, opened 2023-07-16
hint (or directions?) for move_semantics2 are confusing/wrong?


solarshado:
Working my way through 5.5.1, and the hint for `move_semantics2` suggests multiple solutions. However, unless I've misunderstood something, neither of the first two suggestions produce the "expected output" described in the comments above `main()`; they both copy `vec0` before it's populated, leaving it empty when printed.

As far as I can tell, the only suggestion that does work is the third (mutably borrow a reference to `vec0`), which, contradictory to what the hint says, is the only option which *doesn't* affect the output of the first `println!`.

EnderWolf50:
I'm currently doing this and get extemely confused too. 😕

themainframe:
Adding another confused voice to this. Hint 1 - "Make another, separate version of the data that's in `vec0` and pass that to `fill_vec` instead." directly results in the stated expected output at the top of the kata not being satisfied (since `vec0` will remain empty).

This one needs a rethink. Too much going on, needs to focus on the concept of moving more.

jameskr97:
Hello, First time rust write, long time programmer. I thought I'd add add some code references for context, as I also found myself to be a little lost with this problem as-well.

For reference, here is the original problem
https://github.com/rust-lang/rustlings/blob/464cb5566d9e36bf6f8ccf83a105097bce91ad4d/exercises/move_semantics/move_semantics2.rs#L1-L32

Here are the hints associated:

https://github.com/rust-lang/rustlings/blob/464cb5566d9e36bf6f8ccf83a105097bce91ad4d/info.toml#L294-L314

If I understand the problem correctly, the author wants the reader to consider how the problem could be solved with and without changing the signature of the original `fill_vec`. I found 1 (somewhat) and 3 to be the only hints that (for me, a new rust developer) makes sense.

For 1: Point 1 says to "Make another, separate version of the data that's in `vec0` and pass that to `fill_vec` instead.", though vec0 starts with no data, and data is only added in the `fill_vec` function. The two solutions I could think of that fell in line with making "another separate version of the data" was the following: 
```rust
// Possibility 1: two disconnected variables (I think this is what the author intended?)
 fn main() {
     let vec0 = fill_vec(Vec::new()); 
     let mut vec1 = fill_vec(Vec::new());  // the rest of the file remains unchanged
}
// Possibility 2: slicing
// This doesn't fit - `vec0` shadows itself and depends on `vec1` for it's values. I believe
[...]

PR #1660, opened 2023-09-04
fix: refactor move semantics 1-4 into tests

### LINKED ISSUE #1197: Add "Expected output:" comment
First thing first Amazing project has been fun so far.

Now I'm currently in the move_semantics2 and I'm unclear what the output should be.
I Solved it, to the point where it is compiling, as well as read the hints and tried all 3 solutions.

But I'm expecting the output to be `Vec0: [22,44,66]` and for `Vec1: [22,44,66,88]` but im always getting:
`Vec0 []  Vec1[22,44,66,88]` is this what it should be ? 

It seems like having "is able to compile" as the only win condition is a bit confusing (this obviously doesn't apply to the parts with Tests but I don't think its necessary to apply test rules to every part a simple `Expected Output: ....` section would suffice." IMO.

any thoughts ?

shadows-withal:
Fixes #1197, supersedes #1589. Also references #1596.


shadows-withal:
Force pushed to simplify all exercises, so this should now hopefully close #1596 as well.

PR #1667, opened 2023-09-08
Make `primitive_types3` require at least 100 elements

husjon:
The current implementation of `primitive_types3` allows an array of any length to compile, hence invalidating the requirement in the top comment.
This PR adds a panic to the else clause so that it won't compile if the array is too small.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @husjon for content

PR #1678, opened 2023-09-15
remove hint comments when no hint exists

krmpotic:
three files still point to a hint that isn't there.


shadows-withal (approved):

shadows-withal:
@all-contributors please add @krmpotic for content

Issue #1689, opened 2023-09-24
rustlings panicking when updating exercise files


thenorili:
I worked through many exercises over the course of three sessions without trouble using VSCode on Windows. When I got to struct 2 and saved the file, rustlings panicked with the following error message.

```
thread 'main' panicked at 'We were unable to open the exercise file!: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }', src\exercise.rs:205:36
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace */
```

This now happens consistently, including after restarting the terminal and VSCode. Since a new 'rustlings watch' checks all previous exercises, it's now very slow going.

Here's the full backtrace.

```
stack backtrace:
   0:     0x7ff7523e81cc - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\std\src\sys_common\backtrace.rs:44
   1:     0x7ff75240b2cb - core::fmt::rt::Argument::fmt
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\core\src\fmt\rt.rs:138
   2:     0x7ff75240b2cb - core::fmt::write
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\core\src\fmt\mod.rs:1094
   3:     0x7ff7523e2dbf - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\std\src\io\mod.rs:1714
   4:     0x7ff7523e7f7b - std::sys_common::backtrace::_print
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\std\src\sys_common\backtrace.rs:47
   5:     0x7ff7523e7f7b - std::sys_common::backtrace::print
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\std\src\sys_common\backtrace.rs:34
   6:     0x7ff7523eaa4a - std::panicking::default_hook::closure$1
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library\std\src\panicking.rs:269
   7:     0x7ff7523ea69f - std::panicking::default_hook
[...]

PR #1690, opened 2023-09-25
Give a more helpful error when a file is missing

jyn514:
Previously, this would just say "missing file". Now it shows the path of the file that was missing.

This would have made https://github.com/rust-lang/rustlings/issues/1689 easier to debug.


jyn514:
I think ideally this would do something like run `git status` to see if it can detect where the file went, but this seems like a good  simple stopgap.

shadows-withal (approved):

PR #1700, opened 2023-09-28
chore: make hints proper markdown

### LINKED ISSUE #1698: Is making all hints proper Markdown a good idea?
I'm working on a VSCode extension to help with Rustlings: https://marketplace.visualstudio.com/items?itemName=drwilco.rustlings-helper

One of the features is showing the hints when the user wants to. A lot of hints use Markdown, but not all of them do it well. Would it be welcome if I went through all hints in `info.toml` and update them all to use proper Markdown?

Examples:

- `variables3` doesn't have an empty line between the two paragraphs.
- `variables5` has the same problem.
- `variables6` uses ticks instead of backticks around `let` and `const`.
- `primitive_types3` should probably use triple backticks.

And many many more.

I'm quite willing to go through them all and make sure they render nicely, but figured it'd be a good idea to ask before getting started.

docwilco:
Also rewrapped some hints to 80 columns so that they also look good in a terminal.

closes #1698

### DIFF
--- a/info.toml
+++ b/info.toml
@@ -5,7 +5,7 @@ name = "intro1"
 path = "exercises/intro/intro1.rs"
 mode = "compile"
 hint = """
-Remove the I AM NOT DONE comment in the exercises/intro/intro1.rs file
+Remove the `I AM NOT DONE` comment in the `exercises/intro/intro1.rs` file
 to move on to the next exercise."""
 
 [[exercises]]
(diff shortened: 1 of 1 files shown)


shadows-withal (approved):
I can't guarantee that I'll be able to keep the strict wrapping in the future, but this is a good start! Thanks!

docwilco:
I don't care about the wrapping for my extension, it ignores that, so I don't mind at all! It was very much a "while I'm at it" kinda thing. 😅

PR #1704, opened 2023-10-04
Add message if run next success

Cript:
Nothing happens when execute `rustlings run next` command if `// I AM NOT DONE` was not removed from exercise and exercise compiles without errors and tests are passed.


mo8it:
Your contribution is much appreciated, but I think this is best done in https://github.com/rust-lang/rustlings/pull/1931 because it contains many changes now and will be the base for version 6. I will make sure that this is done in v6 then :)

PR #1728, opened 2023-10-16
fix(intro1.rs): typo in the exercise body

markgreene74:
Very minor typo, but still worth a fix :)


dbofmmbt:
Definitely worth it, on VS Code you can ctrl + click in a path-like text in the terminal and go directly to the file. As the path is incorrect, you can't use this feature.

shadows-withal (commented):
Thanks! Good catch!

PR #1729, opened 2023-10-17
docs: fix windows installation instructions

VeeDeltaVee:
Currently, the windows installation instructions download a script from the URL `ps1.rustlings.cool`. This URL isn't detected as a URL in some cases, which means that PowerShell tries to load the data from a local file called `ps1.rustlings.cool`.

This was breaking my install, and adding the `https://` fixed it.


shadows-withal (approved):
Good catch, thanks!

shadows-withal:
@all-contributors please add @VeeDeltaVee for docs

PR #1735, opened 2023-10-20
fix(intro2): Changed intro2 to be a name error, not a format string error

0atman:
I think intro2 is too difficult to be the very first real exercise, and we're scaring away some students unnecessarily. 

**TL;DR: I've made a tweak to intro2 to cause a much friendlier compiler error.**

## My Case Against The Existing Exercise

The existing error for `intro2` is more difficult to understand compared with subsequent exercises. That's not right, the Rustlings exercises are otherwise in ascending order of difficulty. The first one should be the easiest!

The existing intro2 error is:

```rust
error: 1 positional argument in format string, but no arguments were given
  --> exercises/intro/intro2.rs:11:21
     println!("Hello {}!");
                     ^^
```

This is *terrifying* as an error! There's not even a hint or compiler help to aid us! Even existing programmers coming from other languages might be stumped:
- How do they know what a format string is? 
- How are they supposed to know that `{}` is called a 'positional argument'?
- There IS an argument there, it's the string argument to `println!` (they might reasonably think).

I don't think there's a need to teach the student about format strings right here, we're about to use them multiple times in a much friendlier way in the very next exercise: `variables`.

If this exercise were named `formatstring1` I would understand, but it's not, it's named `intro2`:
**This is a poor intro to Rustlings, but one we can fix easily.**

## My Case For The Proposed Change

I propose the first exercise should both showcase Rust's fantastic errors AND show that the compiler can often tell you how to solve your mistakes with help text. 

I propose changing `intro2` to the following, but I would agree with any proposal that has a clear error with correct compiler help text.

```rust
printline!("Hello there!");
```

The compiler error for this is much simpler, AND comes with a correct help suggestion. The error is:

```rust
error: cannot find macro `printline` in this scope
   --> exercises/00_intro/intro2.rs:11:5
    |
11  |     printline!("Hello there!")
    |     ^^^^^^^^^ help: a macro with a similar name exists: `println`
    |
   ::: /home/oatman/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/macros.rs:138:1
    |
138 | macro_rules! println {
    | -------------------- similarly named macro `println` defined here

```

The student only needs to read the help line to know how to fix the error. :tada:

### DIFF
--- a/exercises/00_intro/intro2.rs
+++ b/exercises/00_intro/intro2.rs
@@ -8,5 +8,5 @@
 // I AM NOT DONE
 
 fn main() {
-    println!("Hello {}!");
+    printline!("Hello there!")
 }
--- a/info.toml
+++ b/info.toml
@@ -13,7 +13,7 @@ name = "intro2"
 path = "exercises/00_intro/intro2.rs"
 mode = "compile"
 hint = """
-Add an argument after the format string."""
+The compiler is informing us that we've got the name of the print macro wrong, and has suggested an alternative."""
 
 # VARIABLES
 


0atman:
As a background to this, I tutor students 1:1 in Rustlings, and though they do Rustlings in their own time, I often help them set up rustup and rustlings on their computer, and to show them how everything works, ask them to complete `intro1` and `intro2`.
I have watch students struggle with `intro2` a lot. If it were named `formatstring1` I would understand, but it's not.

0atman:
Though I have read CONTRIBUTION.MD, this is my first PR here, and so I may need help to do things by the books. I have updated the exercise and the hint, and would welcome any suggestions that adheres to the the goal of making this first exercise as simple and friendly as possible.

Thanks very much to everyone in the Rustlings project, it's the best code kata system I've ever used :smiley:

chrischtel:
Hello everyone, 
I think this pull request is actually a very good Idea. I personally had a problem fixing the second Intro when I started out learning rust and it scared me a little bit. The updated version by @0atman seems to be a lot easier for new rustaceans.

shadows-withal (approved):
Agreed, sorry for the delay!

shadows-withal:
@all-contributors please add @0atman for content

Issue #1746, opened 2023-10-29
Would you be willing to include lessons on capturing (closures / async blocks) and the `Fn*` traits?


junderw:
One topic I see new people hit pretty early on is the capturing semantics and the `Fn*` traits.

I was thinking of adding a few lessons to show:

1. Capturing semantics without `move`
2. Capturing semantics with `move` (including ie. binding a reference to a variable before using it in the closure and a reference is moved)
3. How the `Fn*` traits are auto-implemented.
4. Briefly mention that async block capturing is similar to closure capturing, but that there is no `Fn*` traits for async-block-futures.

I wanted to read the room a bit before actually putting time into this. Do we think this might be a bit too advanced for rustlings? Would it depend on the actual content? Would anyone be willing to help review my lessons to try and get it over the finish line?

It's an issue I get asked about constantly, especially when dealing with GUI frameworks like egui that make use of lots of nested closures... it might help people intuit these issues more clearly.

Let me know what you think!

divyaranjan1905:
> I think closures should be explained after traits and tests. This is in line with the Rust book as well.

Indeed, as I proposed in the issue, it would replace the currently 18th set (iterators), so would be after traits and tests.

mo8it:
You are right, I forgot about this PR: https://github.com/rust-lang/rustlings/pull/1748

I agree @junderw, that PR only shows the basics. After talking about traits, I would like to have 3 additional exercises where you need to write your own functions that take a closure, one with `Fn`, one with `FnMut` and one with `FnOnce`. What do you think?

divyaranjan1905:
> After talking about traits, I would like to have 3 additional exercises where you need to write your own functions that take a closure, one with `Fn`, one with `FnMut` and one with `FnOnce`. What do you think?

Yep, @mo8it I think that'll be in line with the chapter from The Book. I'm considering to keep the number of exercises to 4, or at most 5.

rikettsie:
@divyaranjan1905, @junderw , when I wrote that couple of exercises last year (#1748), I considered them a gentle introduction
to closures.

I still think it's a good idea to have a taste of closures early in those chapters (_functions_ and _move_semantics_),
and definitely a very good thing to add a dedicated chapter later to build a deeper understanding (probably between _traits_ and _iterators_).

@mo8it, I updated the PR #1748 to compile to the new rustlings standard, if you think it's valuable to merge.

junderw:
I rebased and added a closures2.rs to the PR in #2099

PR #1748, opened 2023-11-01
feat: add functions6.rs and move_semantics7.rs exercises about closures

rikettsie:
This PR contains a couple of exercises about closures, one in the `02_functions` group, the other in the `06_move_semantics`. Hints and documentation references are added to correctly complete the exercises.

### DIFF
--- a/exercises/02_functions/README.md
+++ b/exercises/02_functions/README.md
@@ -1,8 +1,9 @@
 # Functions
 
 Here, you'll learn how to write functions and how the Rust compiler can help you debug errors even
-in more complex code.
+in more complex code. You will also learn what is the difference with closures.
 
 ## Further information
 
 - [How Functions Work](https://doc.rust-lang.org/book/ch03-03-how-functions-work.html)
+- [Closures](https://doc.rust-lang.org/book/ch13-01-closures.html)
--- a/exercises/02_functions/functions6.rs
+++ b/exercises/02_functions/functions6.rs
@@ -1,0 +1,20 @@
+// functions6.rs
+//
+// Here you can practice special functions called `closures`, that can capture
+// variables of their parent context.
+// Fix the code below to make it compile, without changing the two closure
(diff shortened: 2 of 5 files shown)


chrestomanci:
I am a student learning Rust, and while I support adding exercises about closures to Rustlings, I think it would be unfair and needlessly difficult for students to add those exercises to the 02_functions and 06_move_semantics section of Rustlings, because closures are considered a more advanced feature of Rust, and they are not described till later in the Rust book.

A student who is a beginner with learning rust, and is following along the Rust book would find these exercises too hard for their level of learning, and would not be able to skip them easily.

Issue #1937 relates to how the order of some exercises is slightly different from the book, but for that issue it would be fairly easy for the student to read ahead and solve the problems. For these exercises you propose, I don't think a student could realistically read ahead because there are a great many concepts that would need to be understood to get that far into the book.

Could you create another section for these and other advanced exercises ?

rikettsie:
@chrestomanci, your point makes sense, but as you say your perspective comes from your personal learning experience, and I think we should always be careful when reasoning about what is a good learning path, as there's no standard one.

Very often, gently introducing harder concepts earlier, allows one to better take them on later for deeper understanding.

In this case, because rust heavily supports functional patterns, you may argue that it's not too early to bring the student a taste of those special functions, namely closures, suggesting how they behave differently (e.g. in _"passing"_ in parameters) right after studying function usage.

The idea here is this is only an introduction suggesting for more room later, i.e. a dedicated chapter probably before _iterators_ and _mappers_, as these components leverage functional rust with the full expressivity of closures.

A discussion about such a dedicated chapter is going on with the PR #1746 if you want to join.

senekor:
The order in which topics are taught in the Rust book is well thought-out, I don't think we should deviate from it. There is much value in users being able to go through the book and practice with Rustlings at the same time.

PR #1753, opened 2023-11-04
Fix progress bar

### LINKED ISSUE #1691: Progress bar is inconsistent between file watcher and initial startup
Steps to replicate:
1. Apply this diff:
```
$ git apply
diff --git a/exercises/intro/intro2.rs b/exercises/intro/intro2.rs
index 990b20f..8161d1d 100644
--- a/exercises/intro/intro2.rs
+++ b/exercises/intro/intro2.rs
@@ -5,8 +5,6 @@
 // Execute `rustlings hint intro2` or use the `hint` watch subcommand for a
 // hint.

-// I AM NOT DONE
-
 fn main() {
-    println!("Hello {}!");
+    println!("Hello {}!", "world");
 }
```
2. Run `cargo watch`. The initial progress bar will show 0/92 exercises:
```
Progress: [------------------------------------------------------------] 0/96 (0.0 %) 
✓ Successfully ran exercises/intro/intro1.rs!
```
3. Save any exercise file (e.g. by adding whitespace). The progress bar will show 2/92 exercises:
```
Progress: [>-----------------------------------------------------------] 1/96 (1.0 %) 
Progress: [#>----------------------------------------------------------] 2/96 (2.1 %) 
✓ Successfully ran exercises/intro/intro1.rs!
```

---

Here is the way it's calculated on startup: https://github.com/rust-lang/rustlings/blob/4d04aad8902262d73d05b066ef0b8ffcd4fb634d/src/main.rs#L340-L345
and on changes: https://github.com/rust-lang/rustlings/blob/4d04aad8902262d73d05b066ef0b8ffcd4fb634d/src/main.rs#L356-L372

In particular, the first one only considers exercises *before* the first non-compiling exercises, while the latter includes exercises after.

I think the first one is probably the expected behavior? and it would have the advantage of raising errors like https://github.com/rust-lang/rustlings/issues/1689 lazily, only when the learner gets to that exercise. but either is probably fine as long as they're consistent.

danieltinazzi:
Fix #1691 
The saved exercise was counted twice, the second time was inside the verify function with `bar.inc(1)`.
Now the count after saving is consistent with the count at startup.
Please let me know if there's any doubt about this fix.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -362,7 +362,10 @@ fn watch(
                                     .iter()
                                     .filter(|e| !e.looks_done() && !filepath.ends_with(&e.path)),
                             );
-                        let num_done = exercises.iter().filter(|e| e.looks_done()).count();
+                        let num_done = exercises
+                            .iter()
+                            .filter(|e| e.looks_done() && !filepath.ends_with(&e.path))
+                            .count();
                         clear_screen();
                         match verify(
                             pending_exercises,


shadows-withal (approved):

shadows-withal:
@all-contributors please add @danieltinazzi for content

PR #1766, opened 2023-11-11
fix(installation): Optimize Rustlings Installation in Gitpod Workspaces

haraldreingruber:
Currently, when opening a Gitpod workspace, Rustlings is cloned into a sub-directory. This can be confusing and also leads to issues with the paths of the compile errors not opening the correct file when ctrl+clicking the paths in the terminal output.

To improve the user experience, this Pull Request proposes to skip the cloning process when the `install.sh` script is running in a Gitpod workspace. This change will ensure that the Rustlings exercises are directly accessible in the workspace root, thereby resolving the path issues in the terminal output.

At the moment, the latest main branch revision is used, which is probably fine too. Or would you prefer still checking out the latest tagged release?
If this might make also sense for GH Codespaces, please let me know.

Feedback and suggestions are welcome.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -135,34 +135,40 @@ else
     echo "SUCCESS: Rust is up to date"
 fi
 
-Path=${1:-rustlings/}
-echo "Cloning Rustlings at $Path..."
-git clone -q https://github.com/rust-lang/rustlings "$Path"
+# Check if GITPOD_WORKSPACE_ID environment variable exists
+if [ -z "$GITPOD_WORKSPACE_ID" ]; then
+    Path=${1:-rustlings/}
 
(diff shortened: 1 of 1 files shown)


shadows-withal:
I'm curious: Would the solution in #1780 also work for this?

haraldreingruber:
> I'm curious: Would the solution in #1780 also work for this?

Good thinking. 
I haven't tested it yet, but with the changes from #1780 it should not be a problem calling `install.sh --no-clone` from `.gitpod.yml` when the workspace is initialized on Gitpod.

mo8it:
Closing because the installation script is removed in the upcoming v6

PR #1774, opened 2023-11-14
Make enum variants more consistent between exercises

### LINKED ISSUE #1545: Enums3 rustling is unclear
I've been stuck on the enums3 rustling, specifically around matching on Move() and feel like compared to all prior exercises it is unclear. I think naturally you start with `"Message::Move() => {"` and iterate from there. The rust documentation doesn't detail how to do this kind of matching here: https://doc.rust-lang.org/book/ch06-02-match.html (maybe it's somewhere else that I missed?)

Regardless, iterating from there, you eventually can stumble into `Message::Move{x, y} => {` which compiles, at least. Which leads to frustration two - how to process the match. Iterating through the error messages, I eventually ran into `self.move_position(Point{x:x.try_into().unwrap(), y:y.try_into().unwrap()});`

Which leads to an error in the test at line 75, "not a value". This error message isn't super clear (I assume it's native rust, not part of the rustlings project, but still worth mentioning), and it's not clear what to do from here. The hints aren't super helpful, though I assume what I need to do is some kind of deconstruction/construction. However, I've brute-forced through each decon/con pairing I can think of and none of them have worked, without really any indication of what is the proper one either through documentation or error messages.

I realize that `self.move_position(Point{x:x.try_into().unwrap(), y:y.try_into().unwrap()});` contrasts with what the "Remember:" hint seems to be implying the answer should be - something like `self.move_position((x, y));`, but I could never get any variation of that to compile. Further, it's confusing to have the 'fn' characters there - this seems to be implying that when you're defining a function with a tuple parameter you need the extra paratheses, but for this exercise you're not defining any functions.

Regardless of how my tone may sound, I appreciate the immense amount of work that's gone into this project and am only trying to improve it.

matthri:
Make sure the `Point` struct is already introduced in `enums2` and can get reused in the following exercise.
I added an additional `Resize` variant since I think it is important to highlight the difference between the wrapped `Point` struct and an anonymous struct.

Additionally I changed the order of the listed variant types of the `enums2` exercise to match the order of the variants in the code to make the connection a bit clearer.

closes #1545


mo8it:
Thanks :D

I will adjust the solutions.

PR #1775, opened 2023-11-15
Update `notify` dependency to v6

### LINKED ISSUE #1640: Update notify to v6
Update `notify` to version 6. The jump to version 5 broke debounced events and at least one additional crate is needed ([notify-debouncer-mini](https://docs.rs/notify-debouncer-mini/latest/notify_debouncer_mini/) or [notify-debouncer-full](https://docs.rs/notify-debouncer-full/latest/notify_debouncer_full/)).

matthri:
This updates `notify` to v6 and uses `notify-debouncer-mini` for event debouncing. 

**Please Note:** `notify-debouncer-mini` does NOT distinguish different event kinds but also has fewer dependencies than `notify-debouncer-full`.
I don't think this is a problem in this case, but please let me know if you prefer the version where event kinds are explicitly checked, then I will happily update this PR.

PS: I think one could think about moving the event handling to a separate function, as the watch function is getting a bit complex. For easier review I tried not to do too much changes to the structure. 
So I can do the refactoring in a follow up PR or include it in this one if you prefer it. 

closes #1640

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,8 +4,8 @@ use crate::run::{reset, run};
 use crate::verify::verify;
 use clap::{Parser, Subcommand};
 use console::Emoji;
-use notify::DebouncedEvent;
-use notify::{RecommendedWatcher, RecursiveMode, Watcher};
+use notify_debouncer_mini::notify::{self, RecursiveMode};
+use notify_debouncer_mini::{new_debouncer, DebouncedEventKind};
 use std::ffi::OsStr;
 use std::fs;
 use std::io::{self, prelude::*};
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -350,38 +352,44 @@ fn watch(
...
-                    if b.extension() == Some(OsStr::new("rs")) && b.exists() {
-                        let filepath = b.as_path().canonicalize().unwrap();
-                        let pending_exercises = exercises
-                            .iter()
-                            .find(|e| filepath.ends_with(&e.path))
-                            .into_iter()
-                            .chain(
+                Ok(events) => {
+                    for event in events {
+                        let b = event.path;
```
shadows-withal, line 357: Could you turn this into a more descriptive variable name like `event_path` or something? Thanks!

matthri: That naming was adopted from the existing code, nevertheless I have updated the variable name as you suggested :)

shadows-withal (approved):

shadows-withal:
@all-contributors please add @matthri for code

PR #1779, opened 2023-11-18
feat: add refcell1 exercise

exdx:
Adds a basic exercise to introduce the concept of a `RefCell`. This API is not necessarily for beginners but is useful to have knowledge of.

### DIFF
--- a/exercises/19_smart_pointers/refcell1.rs
+++ b/exercises/19_smart_pointers/refcell1.rs
@@ -1,0 +1,52 @@
+// refcell1.rs
+//
+// Interior mutability is a design pattern in Rust that allows you to mutate 
+// data even when there are immutable references to that data; 
+// normally, this action is disallowed by the borrowing rules.  
+
+// The RefCell<T> type represents single ownership over the data it holds.
+// Recall the borrowing rules in Rust:
+//      1. At any given time, you can have either (but not both) one mutable 
+//      reference or any number of immutable references.
+//      2. References must always be valid.
+
+// With references and Box<T>, the borrowing rules’ invariants are enforced at 
+// compile time. With RefCell<T>, these invariants are enforced at runtime. 
+// With references, if you break these rules, you’ll get a compiler error. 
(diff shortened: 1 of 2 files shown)


shadows-withal:
I like the idea of teaching about RefCells, but I feel like this exercise would have to be somewhat more sophisticated... I'd like people to be able to see the difference of behaviour between the normal compile-time borrow check and what RefCell provides, maybe a more engaging exercise format could be converting something that's impossible with normal borrowing rules to using RefCell?

exdx:
Thanks for taking a look @shadows-withal. I agree the example could be more sophisticated and I will look into updating it. Potentially following the Rust Book more closely is a good direction to take.

exdx:
@shadows-withal I have updated the example to use the `compile` instead of `test` mode. I think this make the logic considerably simpler. The example is a straightforward example of using a RefCell to mutate via an immutable reference.

I think since this may be the first introduction to RefCell a user has, it's simplicity is a feature. Happy to adjust as necessary.

mo8it:
The exercise teaches the usage of `RefCell` very well, but not really _when to use it_. It might be just confusing why the user shouldn't use `&mut self` since a `RefCell` is not really required in this example at all.

exdx:
> The exercise teaches the usage of `RefCell` very well, but not really _when to use it_. It might be just confusing why the user shouldn't use `&mut self` since a `RefCell` is not really required in this example at all.

I see -- that's fair. At the risk of making the example more convoluted I can update the example to provide a scenario where refcell is a necessity.

mo8it:
I am a very new maintainer. Let's wait for the feedback from @shadows-withal first :D

Piyuuussshhh:
I think exercises on Cell, RefCell, RwLock and Mutex in sequence would be nice. I remember there was a video that explained when these smart pointers become useful quite nicely. It explains why and when interior mutability is required.
The dude took the construction of a graph and updating the value of a node & its neighbors as the running example throughout the vid. If you want, I could link it!

PR #1626, opened 2023-08-17
fix(hashmap2): fix all_fruit_types_in_basket passes when a fruit type is missing

danielsomerfield:

PR #1782, opened 2023-11-21
Fix all_fruits_types_in_basket to fail if all fruit kinds are not  included

danielsomerfield:
This is a redo of PR #1626 which I flaked on. This accounts for the new file structure and the formatting request.

### DIFF
--- a/exercises/11_hashmaps/hashmaps2.rs
+++ b/exercises/11_hashmaps/hashmaps2.rs
@@ -18,7 +18,7 @@
 
 use std::collections::HashMap;
 
-#[derive(Hash, PartialEq, Eq)]
+#[derive(Hash, PartialEq, Eq, Debug)]
 enum Fruit {
     Apple,
     Banana,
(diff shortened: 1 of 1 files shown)


exercises/11_hashmaps/hashmaps2.rs
```diff
@@ -27,6 +27,14 @@ enum Fruit {
     Pineapple,
 }
 
+const FRUIT_KINDS: [Fruit; 5] = [
```
shadows-withal, line 30: this can be in the test module, and there's no reason to have this be an array when it could be a `Vec<Fruit>` instead

mo8it:
@danielsomerfield Could you please apply the requested changes so that I can merge? :D

danielsomerfield:
Sorry, wasn't getting emails until recently. 

I can't move it into the test mod as both a const and a vec, but I could push it into the test function as a vec. Let me know if that is ok. Otherwise, I believe it has to be an array.

PR #1794, opened 2023-11-30
update hashmaps3 instructions to mention summing

diegomouradev:
We found it confusing that the scores were supposed to be cumulative over the whole input, so hopefully, this is a better description!


mo8it:
The PR https://github.com/rust-lang/rustlings/pull/1869 might be a better fit. What do you think @diegomouradev ?

diegomouradev:
@mo8it Thank you for considering my suggestion. PR #1869 does help clarify the exercise.

PR #1799, opened 2023-12-08
Remove confusing aside in 23_conversions/from_str.rs

NicolasRoelandt:
The advice tell us how to return as String error message. Unless I missed something, we can't even return a String error message in this exercise (we have to return a ParsePersonError), so this advice is more confusing than anything and should better be removed.


mo8it:
@all-contributors please add @NicolasRoelandt for doc

PR #1801, opened 2023-12-08
docs/better-base-url

radu-relantin:
Changed the base url to the better version of rust-lang docs

https://rust-book.cs.brown.edu/


shadows-withal:
I'm not really interested in deviating from the canonical book source.

PR #1809, opened 2023-12-27
fix(quiz2): incorrect value in test

yvf:
There's nowhere for "bar" to come from, as shown by the next line.

### DIFF
--- a/exercises/quiz2.rs
+++ b/exercises/quiz2.rs
@@ -58,7 +58,7 @@ mod tests {
         ]);
         assert_eq!(output[0], "HELLO");
         assert_eq!(output[1], "all roads lead to rome!");
-        assert_eq!(output[2], "foobar");
+        assert_eq!(output[2], "foofoo");
         assert_eq!(output[3], "barbarbarbarbarbar");
     }
 }


matthri:
Hi, 

as written in the comment at the beginning of the quiz the task for the `Append` command is to `Append "bar" to the string a specified amount of times`. 

Therefore you have to implement the code which appends (and provides) the string `bar` to the given string. 
So the current value in the test is correct and represents the desired behavior.

Just guessing here, but I think you maybe misunderstood the "append `bar` n times" and interpreted it instead as "append the input (given string) n times"

I hope this helps to clarify this and feel free to ask if you have any questions or if I have misunderstood your PR and you think the issue is still present.

matthri (commented):
Hi,

as written in the comment at the beginning of the quiz the task for the `Append` command is to `Append "bar" to the string a specified amount of times`.

Therefore you have to implement the code which appends (and provides) the string `bar` to the given string.
So the current value in the test is correct and represents the desired behavior.

Just guessing here, but I think you maybe misunderstood the "append `bar` n times" and interpreted it instead as "append the input (given string) n times"

I hope this helps to clarify this and feel free to ask if you have any questions or if I have misunderstood your PR and you think the issue is still present.

yvf:
Apologies. It appears I just hadn't quite read the instructions properly.

matthri:
Don't worry, it happens.

PR #1822, opened 2024-01-14
Next exercise

patrikkaprinay:
While doing the exercises I found myself wanting this small piece of information which file I'm going to have to edit next, so while the 1 sec timeout runs out in the rustlings watch, I can be already switched to that file.

### DIFF
--- a/src/exercise.rs
+++ b/src/exercise.rs
@@ -168,7 +168,7 @@ path = "{}.rs""#,
                     .output()
             }
         }
-            .expect("Failed to run 'compile' command.");
+        .expect("Failed to run 'compile' command.");
 
         if cmd.status.success() {
             Ok(CompiledExercise {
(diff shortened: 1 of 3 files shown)


matthri (commented):
Hi, 
thank you for helping to improve this project 👍.

> A little disclaimer: 
I am not a maintainer of this project, so I can't make any decisions here. These are just my personal observations and opinions.

I have added a few comments to your code with some thoughts that came to my mind while reading your PR. 
I hope those comments don't sound harsh or unfriendly, they are just meant as some ideas that could be worth to consider. 

While reading your PR it's a bit unclear to me, what problem you are trying to solve (this could also be an issue on my end  😉). 
While I think it's a good idea to provide the students with some guidance through the exercises, I'm not sure if displaying the next path in the success message is providing much benefit over the current situation. 

This is especially because in `rustlings watch` mode Rustlings tries to compile the next exercise automatically after you remove the `I AM NOT DONE` comment and displays the file in which the error for the next exercise occurs. 
In VSCode this path is even clickable, which will open the exercise for you. 

So maybe you can elaborate a bit more on why the current solution doesn't fit your needs and what should be improved.

PR #1829, opened 2024-01-18
fix(from_into): correcting test from commit 2691a35

charlielu05:
I don't think the commit: https://github.com/rust-lang/rustlings/commit/2691a35102b0eab111706da0c27fc39541b64e0c should of been merged into main, in the comments for this exercise it states:

// 5. Extract the other element from the split operation and parse it into a
//    `usize` as the age.
// If while parsing the age, something goes wrong, then return the default of
// Person Otherwise, then return an instantiated Person object with the results

This should mean that the original test code  were correct in that it returns the default Person since parsing `32,` and `32,man` respectively should result in an Error.

### DIFF
--- a/exercises/23_conversions/from_into.rs
+++ b/exercises/23_conversions/from_into.rs
@@ -127,14 +127,14 @@ mod tests {
     #[test]
     fn test_trailing_comma() {
         let p: Person = Person::from("Mike,32,");
-        assert_eq!(p.name, "Mike");
-        assert_eq!(p.age, 32);
+        assert_eq!(p.name, "John");
+        assert_eq!(p.age, 30);
     }
 
     #[test]
     fn test_trailing_comma_and_some_string() {
         let p: Person = Person::from("Mike,32,man");
-        assert_eq!(p.name, "Mike");
-        assert_eq!(p.age, 32);
+        assert_eq!(p.name, "John");
+        assert_eq!(p.age, 30);
     }
 }


matthri (commented):
Hi, 

it depends on how the steps in the comment are interpreted (I agree that they maybe should be described a bit clearer).

The comment tells you to `2. Split the given string on the commas present in it.` (note the plural form of commas).
Therefore I would assume one should use the `split()` method which splits the `&str` and returns an iterator over n elements. 
In the test cases the iterator would return the following elements (I formatted them as a list for readability):
`test_trailing_comma` -> `["Mike", "32", ""]`
`test_trailing_comma_and_some_string` -> `["Mike", "32", "man"]`
When one now extracts the first and second element the values are valid and the parsing succeeds.
In this case the old test cases represent the correct behavior.

If one assumes that the string should be splitted only once by using `split_once()` which returns a `Option<(&str, &str)>` then the second element would contain `32,` or `32,man` respectively. 
In that case the parsing would fail and your test cases represent the correct behavior.

As you can see the correct behavior heavily depends on the interpretation of the exercise description and I think the best way to solve this is making the description of the expected steps clearer.

charlielu05:
Hi @matthri , I think you've pointed out my blindspot. I was too focussed on step 5 in my comment where I assumed it should of been done with a `split_once()` which made me totally miss out on the plural `commas` in step 2.

PR #1830, opened 2024-01-18
optimized the UI code

lucaplian:
I did a refactoring on the 'warn' and 'succes' macros. And created a new macro called print_emojis which did what the previous macros did but more generalized. I used the expr for "$emoji" and "$sign", because it can accept any UTF8 characters. In the case of "$color", I used the ident because it allowed me for customizability for the colour, because the variable to be used as an identifier(in my case a function)
I did this in order to avoid the duplication from the original code


alexandruradovici:
Please describe your pull request.

AnonimAnonim2245:
I did a refactoring on the 'warn' and 'succes' macros. And created a new macro called print_emojis which did what the previous macros did but more generalized. I used the expr for "$emoji" and "$sign", because it can accept any UTF8 characters. In the case of "$color", I used the ident because it allowed me for customizability for the colour, because the variable to be used as an identifier(in my case a function)

AnonimAnonim2245:
I did this in order to avoid the duplication from the original code

alexandruradovici:
> I did a refactoring on the 'warn' and 'succes' macros. And created a new macro called print_emojis which did what the previous macros did but more generalized. I used the expr for "$emoji" and "$sign", because it can accept any UTF8 characters. In the case of "$color", I used the ident because it allowed me for customizability for the colour, because the variable to be used as an identifier(in my case a function)

Please add it to the beginning of the pull request.

shadows-withal (approved):

PR #1832, opened 2024-01-19
chore: update from_into.rs task description to fit the code

reifenrath-dev:
I noticed that the comment doesn't exactly fit the addressed line of code, so here is my humble text fix.


shadows-withal:
@all-contributors please add @reifenrath-dev for content

PR #1841, opened 2024-01-30
Update variables2.rs

omcodedthis:
- Improved phrasing of variables2.rs to emphasize that the error lies in the the declaration of the variable x.


omcodedthis:
Good spot, thanks. Updated this.

LeverImmy:
But I think this so-called 'emphasize' is **meaningless**. 'is' is the same as 'is equal to'. You read the symbol '=' to be 'is', too.

PR #1842, opened 2024-01-30
Update exercise.rs

omcodedthis:
- Updated metadata of version info and year to the current version of rustlings.


matthri (commented):
Hi,
the `edition` is a Rust feature which is used to allow non backwards compatible changes to the language (see [the edition guide](https://doc.rust-lang.org/edition-guide/editions/)).

Since there is no `edition 2023` unfortunately this isn't a valid change.

omcodedthis:
- Closed to o allow non backwards compatible changes.

PR #1877, opened 2024-03-09
Simplify and shrink development container

kylev:
This replaces a very large "universal" image with the typical rust base. I also trimmed down some non-functioning/excess config obviated by the presence of the container's toolchain and Dev Container lifecycle (no checkout needed).


shadows-withal (approved):

shadows-withal:
@all-contributors please add @kylev for infra

PR #1882, opened 2024-03-13
feat: stash all changes at once

db-keli:
added a function to run.rs that stashes all changes in exercises. This is to help reset all exercises

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,6 @@
 use crate::exercise::{Exercise, ExerciseList};
 use crate::project::RustAnalyzerProject;
-use crate::run::{reset, run};
+use crate::run::{reset, run, reset_all};
 use crate::verify::verify;
 use clap::{Parser, Subcommand};
 use console::Emoji;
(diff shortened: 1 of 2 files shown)


shadows-withal:
I don't think this current implementation is a good idea, running ~90 `git stash` processes in a row is going to cause some sort of race condition. If people want to reset their entire workspace, they could just run `git checkout .`.

db-keli:
Fine but that will only move head to latest committed change. To someone who commits their changes every single time or ever committed a change, that wouldn't revert back to default🙃

db-keli:
Fine but that will only move head to latest committed change. To someone who commits their changes every single time or ever committed a change, that wouldn't revert back to default🙃

shadows-withal:
That's true... I suppose reverting to the tag the user installed at would be the correct way to do it.

mo8it:
> I suppose reverting to the tag the user installed at would be the correct way to do it.

Either that or checking out the HEAD of the `origin` remote.

Personally, I don't see the benefit of this approach over just installing Rustlings again in a different directory. That way, you have the option of keeping your old solutions and could get improvements from a newer version of Rustlings.

In any way, we agree that this PR isn't the solution. Therefore, I will close it but you can still comment :)

PR #1908, opened 2024-03-21
feat: add lib

honeywest:


mo8it:
May I ask why? 😅
I might be missing some advantage here.

honeywest:
Code is easier to read and understand

honeywest:
Code is easier to read and understand.
This structure is better, and the purpose here is to teach people how to learn rust

mo8it:
I will leave the decision to @shadows-withal

shadows-withal:
I think for people wanting to work on Rustlings itself, learning about the lib/main split is more confusing than convenient, and we derive no significant benefit from this (e.g. Rustlings is pretty much never going to be a "library"), so I'm gonna close this.

PR #1923, opened 2024-03-27
chore: update the chapter of macros

wznmickey:
The chapter of macros is 19.5,not 19.6.
picture
The link of this chapter is ch19-06, but it seems that a chapter before is removed, so macros is now 19.5


mo8it:
@all-contributors please add @wznmickey for doc

PR #1944, opened 2024-04-08
docs: add more info in threads

Selflocking:
info.toml: 

```toml
[[exercises]]
name = "threads3"
path = "exercises/threads/threads3.rs"
mode = "test"
hint = """
An alternate way to handle concurrency between threads is to use
a mpsc (multiple producer, single consumer) channel to communicate.
With both a sending end and a receiving end, it's possible to
send values in one thread and receive them in another.
Multiple producers are possible by using clone() to create a duplicate
of the original sending end.
See https://doc.rust-lang.org/book/ch16-02-message-passing.html for more info.
"""
```

threads3'hint contains this link, so it should be placed in Further Information


shadows-withal (approved):

shadows-withal:
@all-contributors please add @Selflocking for content

PR #1961, opened 2024-04-28
docs : add a comma in structs3.rs file 

Allupeng:
when i use rustRover, and there is a grammar error says "Use a comma before 'and' if it connects two independent clauses (unless they are closely connected and short). ".
So add a comma in structs3.rs file will solve this error.


Allupeng:
and i find another grammar in hashmaps1.rs file which is e.g , the correct should be e.g.

mo8it:
I decided to remove the comma because I am not sure about it. But thank for adding the dot :D

PR #1980, opened 2024-05-29
Added missing return type

MukundSinghRajput:


mo8it:
This is intentional. I added a comment about it in version 6: https://github.com/rust-lang/rustlings/blob/990c68efcba8e1e2b7f2d8c5b6c16885d3920010/exercises/02_functions/functions4.rs#L10

PR #2005, opened 2024-06-27
Fixing links from powershell.

wkstar:
The trailing bang breaks the link.


mo8it:
Thanks for the contribution! I would normally merge it, but the installation script will be removed in the upcoming version 6 anyway.

PR #2006, opened 2024-06-27
Fix shell links not opening

wkstar:
The trailing bang breaks the link.


mo8it:
https://github.com/rust-lang/rustlings/pull/2005#issuecomment-2194308521

PR #2015, opened 2024-07-04
move_semantics5: change to fix misleading compiler error

### LINKED ISSUE #1963: move_semantics6 intended solution does not address the error
## Hinted solution
https://github.com/rust-lang/rustlings/blob/71053101c32d0fd374f8e122880b3d682519bac6/exercises/06_move_semantics/move_semantics6.rs#L23-L28
Based on hints and code comments, the expected solution here is to make `string_uppercase()` take ownership of `data`. When done, the compiler prompts to remove `&` from line 25 to match the type of variable `data`.

## Actual error
The actual error shown (before making any changes) is **temporary value dropped while borrowed**.
```
error[E0716]: temporary value dropped while borrowed
  --> exercises/06_move_semantics/move_semantics6.rs:23:13
   |
22 | fn string_uppercase(mut data: &String) {
   |                               - let's call the lifetime of this reference `'1`
23 |     data = &data.to_uppercase();
   |     --------^^^^^^^^^^^^^^^^^^^- temporary value is freed at the end of this statement
   |     |       |
   |     |       creates a temporary value which is freed while still in use
   |     assignment requires that borrow lasts for `'1`
```

## Addressing the compiler error
The real problem here is not that `string_uppercase` takes reference. It i s because in line 25,  a reference to temporary value is created. Since the temporary value goes out of scope after line 25, the reference out-lives the value being referred to.
If we were to follow this thread, we would end up with a solution that does not necessarily take ownership. Some examples below,
```rust
// Should take ownership
fn string_uppercase(mut data: &String) {
    let data_uc = data.to_uppercase();

    println!("{}", data_uc);
}
```
or without another variable
```rust
// Should take ownership
fn string_uppercase(mut data: &String) {
    let data = &data.to_uppercase();

    println!("{}", data);
}
```

I understand that in `main()`, `data` is not used after calling `string_uppercase()` and hence it is valid solution to take ownership. However, my concern is that it does not address the  root cause, rather fixes error as a side effect. As someone using rustlings to learn rust, this confused me.

## Environment
```
@ramenhost ➜ /workspaces/rustlings (main) $ rustc --version
rustc 1.77.2 (25ef9e3d8 2024-04-09)
@ramenhost ➜ /workspaces/rustlings (main) $ rustlings --version
rustlings 5.6.1
```

ramenhost:
Part of move_semantics5, `string_uppercase` function takes `&String` and intended solution is to change the argument to `String`. However, the compiler error was leading towards a different solution causing confusion. 
```
error[E0716]: temporary value dropped while borrowed
  --> ../exercises/06_move_semantics/move_semantics5.rs:21:13
   |
20 | fn string_uppercase(mut data: &String) {
   |                               - let's call the lifetime of this reference `'1`
21 |     data = &data.to_uppercase();
   |     --------^^^^^^^^^^^^^^^^^^^- temporary value is freed at the end of this statement
   |     |       |
   |     |       creates a temporary value which is freed while still in use
   |     assignment requires that borrow lasts for `'1`

```
This PR fixes #1963 such that the new compiler error leads to intended solution.
```
error[E0308]: mismatched types
  --> ../exercises/06_move_semantics/move_semantics5.rs:21:12
     |
20 | fn string_uppercase(mut data: &String) {
     |                               ------- expected due to this parameter type
21 |     data = data.to_uppercase();
     |            ^^^^^^^^^^^^^^^^^^^ expected `&String`, found `String`
     |
```

### DIFF
--- a/exercises/06_move_semantics/move_semantics5.rs
+++ b/exercises/06_move_semantics/move_semantics5.rs
@@ -18,7 +18,7 @@ fn get_char(data: String) -> char {
 
 // Should take ownership
 fn string_uppercase(mut data: &String) {
-    data = &data.to_uppercase();
+    data = data.to_uppercase();
 
     println!("{data}");
 }


mo8it:
Awesome, thanks :D

PR #2016, opened 2024-07-04
fix: un-solve clippy1

tveness:
I was working through these as a refresher with the newest release (thanks!), and noticed that clippy1 is already solved (the diff with the solution is empty, so it feels like this must be a mistake).

This small PR breaks the code in the way I imagine would be intended, where the clippy hint provides the expected solution.

### DIFF
--- a/exercises/22_clippy/clippy1.rs
+++ b/exercises/22_clippy/clippy1.rs
@@ -4,11 +4,9 @@
 // For these exercises, the code will fail to compile when there are Clippy
 // warnings. Check Clippy's suggestions from the output to solve the exercise.
 
-use std::f32::consts::PI;
-
 fn main() {
     // Use the more accurate `PI` constant.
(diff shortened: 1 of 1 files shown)


mo8it:
Oh, yes, this is a copy-pasta mistake. I will close this PR and fix the issue myself because I also want to add a TODO comment.
Thanks!

mo8it:
Fixed in https://github.com/rust-lang/rustlings/commit/0b220f9fffd02f77de1de79f510ae10c7c53b81f

mo8it:
With https://github.com/rust-lang/rustlings/commit/a3657188b63d20b33f6770552169a473f021228f and https://github.com/rust-lang/rustlings/commit/4bf0ddc0e1b6213b9ddd7b804c19326774cdbf61, this shouldn't happen again :D

PR #2019, opened 2024-07-04
Add alternative solution for iterators5

Nahor:
On one hand, nested iterators are somewhat common (container of containers, iterator of containers, container of iterators), so introducing `flatten` could be useful.

On the other hand, `flatten`, `flat_map`, etc. are not mentioned in the Rust book...


mo8it:
Could you please use `flat_map` to only iterate over the values by mapping with `.values()`?

PR #2031, opened 2024-07-07
Update iterator solution in quiz2.rs

NewtonChutney:
The iterator solution uses a for loop in its body, to the dismay of functional programmers.. 🫠
We could use the repeat method in this situation!

### DIFF
--- a/solutions/quizzes/quiz2.rs
+++ b/solutions/quizzes/quiz2.rs
@@ -52,12 +52,7 @@ mod my_module {
             .map(|(mut string, command)| match command {
                 Command::Uppercase => string.to_uppercase(),
                 Command::Trim => string.trim().to_string(),
-                Command::Append(n) => {
-                    for _ in 0..n {
-                        string += "bar";
-                    }
-                    string
-                }
+                Command::Append(n) => string + &"bar".repeat(n),
             })
             .collect()
     }


mo8it:
I was against `str::repeat` because it allocates a `String` first.

But I just benchmarked both methods and `repeat` seems to be more efficient. The checks while appending a `&str` in the loop seem to cost more than the temporary allocation.

mo8it:
Complete with https://github.com/rust-lang/rustlings/commit/0f4cb94cfe31745e8fee9efc7dc1d287e65e16cb

NewtonChutney:
> I was against `str::repeat` because it allocates a `String` first.
> 
> But I just benchmarked both methods and `repeat` seems to be more efficient. The checks while appending a `&str` in the loop seem to cost more than the temporary allocation.

Isn't this the case generally? I remember reading about this somewhere.. 🤔
That functional style programming in Rust at least as efficient as the elaborated for loop..

mo8it:
This is at least the case with iterators over loops.
But `str::repeat` has nothing to do with iterators. In fact, `str::repeat` is implemented over `slice::repeat` which itself is [implemented using a loop](https://github.com/rust-lang/rust/blob/7fdefb804ec300fb605039522a7c0dfc9e7dc366/library/alloc/src/slice.rs#L510-L566) in the standard library. This case is not about functional vs imperative.

NewtonChutney:
> This is at least the case with iterators over loops. But `str::repeat` has nothing to do with iterators. In fact, `str::repeat` is implemented over `slice::repeat` which itself is [implemented using a loop](https://github.com/rust-lang/rust/blob/7fdefb804ec300fb605039522a7c0dfc9e7dc366/library/alloc/src/slice.rs#L510-L566) in the standard library. This case is not about functional vs imperative.

Ahh.. thanks for pointing it out. TIL

I need to look into the standard library's implementations more.. 😅

PR #2043, opened 2024-07-11
Feature/generics3 updt from old pr

poneciak57:
Created a new pull request based on the previous one (#1470).

Added a new generic exercise that focuses on function reusability using generics.

### DIFF
--- a/exercises/14_generics/generics3.rs
+++ b/exercises/14_generics/generics3.rs
@@ -1,0 +1,35 @@
+// generics3.rs
+// Execute `rustlings hint generics3` or use the `hint` watch subcommand for a hint.
+
+// TODO Use your knowledge of generics to enchance the `stringify` function by only changing the signature.
+fn stringify(list: &[&str]) -> String {
+    let items_str = list
+        .iter()
+        .map(|e| e.to_string())
+        .collect::<Vec<String>>()
+        .join(", ");
+    format!("[{items_str}]")
+}
+
+fn main() {
+    // You can optionally experiment here.
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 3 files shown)


poneciak57:
One of the checks failed `Rustlings Tests / dev-check (pull_request)` 
It told me to run command `cargo run -- dev update` but this failed should i open an issue or should i just update this file manually ?

windows 11
rustup 1.27.1
cargo 1.79.0
Error message:
cargo run -- dev update
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running `target\debug\rustlings.exe dev update`
Error: Failed to update the file `dev/Cargo.toml`

Caused by:
    Failed to find the start of the `bin` list (`bin = [`)
error: process didn't exit successfully: `target\debug\rustlings.exe dev update` (exit code: 1)

mo8it:
I can take care of the CI test later. Probably a Windows issue.

We shouldn't use traits / trait bounds because the exercises about them are after the generics exercises.

How about a function taking `&[Option<T>]` and returning `Vec<T>` containing only the `Some` values?

mo8it:
> i tried to have &[Option<T>] as an argument but it requires T to implement Clone (and i couldn't find any suitable workaround)

Probably because of `into_iter` instead of `iter`. I will try to fix it with the CI fail.

mo8it:
> i tried to have &[Option<T>] as an argument but it requires T to implement Clone (and i couldn't find any suitable workaround)

Oh, yeah. We would need to return something like `Vec<&T>` otherwise. But this is fine how you did it with `Vec` as argument.

kh3rld (approved):

PR #2053, opened 2024-07-20
nit: Making panic description in tests3.rs more accurate

paramdeep-singh:
Per title


mo8it:
Fixed in https://github.com/rust-lang/rustlings/commit/5116a812fbcdf8662934ef35e5d6d0d0794a5d47 both the exercise and solution. Thanks :)

PR #2056, opened 2024-07-25
The Traditional Chinese version of Rustlings

TimLai666:
I have translated Rustlings into (Taiwan) Traditional Chinese.


mo8it:
We can't merge it into the official Rustlings. But you are welcome to adapt it to the version 6 of Rustlings and then publish it as third-party exercises in repository of your own.

Here is the guide: https://github.com/rust-lang/rustlings/blob/main/THIRD_PARTY_EXERCISES.md

PR #2062, opened 2024-08-01
update struct name in hashmap3

jimbo5922:
In the exercise `hashmap3`, the name of the struct is `Team`, but I thought `TeamScores` would be better since it relates to scores(There are several places in the same file where `score` is used.).
Similarly, I made changes to the solution.


mo8it:
Thanks :D

PR #2074, opened 2024-08-07
Remove dependency on hashbrown

senekor:
Hashbrown is already the standard implementation of Hash{Map,Set} in the standard library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -519,7 +519,6 @@ version = "6.1.0"
 dependencies = [
  "anyhow",
  "clap",
- "hashbrown",
  "notify-debouncer-mini",
  "os_pipe",
  "ratatui",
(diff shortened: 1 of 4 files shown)


mo8it:
I am aware of that and I appreciate trying to reduce the number of dependencies. But I use `hashbrown` mainly because it has `ahash` without a random state as the default hasher. This is not safe against hash attacks, but Rustlings doesn't need to worry about that and benefits of the improved performance.

`hashbrown` also has more features like [`entry_ref`](https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.entry_ref). I don't think that I use any of these additional features, but still nice to have if needed :)

If you like, you are welcome to replace `hashbrown` with `ahash` and use it as the hasher for the `std` collections in a new PR :D

PR #2075, opened 2024-08-08
Replace hashbrown with ahash

senekor:
hashbrown is already used in the standard library, but we want the improved performance of the different hash algorithm. Using ahash directly conveys this intent more clearly.


mo8it:
Thanks :D

PR #2076, opened 2024-08-08
Improve initialization in workspace

senekor:
- Detect if we are in a cargo project more reliably. (e.g. if `rustlings init` is run in the `src/` directory)

- Refuse to initialize rustlings in a non-workspace cargo project.

- Automatically populate the `workspace.members` field if `rustlings init` is run in a workspace.

  This may be considered risky, as there is no guarantee that's what the user wanted to do. However, it is consistent with the behavior of `cargo new`. Also, newcomers to Rust are unlikely to accidentally be in a cargo workspace, as they won't know how to create one in the first place.

  The use case for initialization in a workspace is when a workshop organizer wants to use rustlings alongside other exerices and provide a single repository with everything in one place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,6 +226,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
+name = "errno"
+version = "0.3.9"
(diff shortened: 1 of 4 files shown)


mo8it:
Thanks! I will merge and improve some things in a commit :)

mo8it:
My improvements: https://github.com/rust-lang/rustlings/commit/140c4e4812eff982cc3e0c9df6fa5076d7b56633

senekor:
Nice!

Issue #2071, opened 2024-08-06
`rustlings` takes lots of time just to detect file changes


GlowCheese:
## Description

This is the video that the issue happens:

https://github.com/user-attachments/assets/00adc400-8e44-4899-9646-a7cff18bd7aa

It can be shown that every time I update the code of an exercise, the `rust-analyzer` builds the whole `exercises` folder (it has like 156 exercises and the extension run the code check for every single of them). During this process, I cannot use any command (`h`, `n`, `l` or `q`). Only after it finishes, rustlings tells me if my solution is correct or not and directs me to the next exercise.

I don't think it's a bug, but rather a code design issue. If I turn of `rust-analyzer`, it seems the issue's gone but I still have to wait for like 1-2 seconds for `rustlings` to detect file changes.

Please look into this issue and fix it. Thanks!

## System Information

- OS: Windows 10 x64 (10.0.19045)
- VSCode: version 1.19.0
- rust-analyzer: 0.3.2062
- rustlings: 6.1.0

mo8it:
Could you please show what plugins you have installed in VS-Code and whether you have any special configuration for `rust-analyzer`?

Rustlings detects a change and attempts to rerun the exercise. But if `rust-analyzer` is rebuilding the project, then cargo is blocked. Noramally, it doesn't take longer than one second to run the exercise.

It seems like a `rust-analyzer` problem.

Are you using WSL?

GlowCheese:
I installed many plugins alongside with `rust-analyzer`. I tried disabling all of them and doing exercises but the issue still happens.

Do I have any special configuration? Nope, haven't opened `rust-analyzer`'s extension settings before.

I'm not sure it's due to `rust-analyzer` problem or not. Tried 4 recent versions of `rust-analyzer` and none of them fixes the issue.

I'm not using WSL.

mo8it:
I reported it in the Rust-Analyzer repo: https://github.com/rust-lang/rust-analyzer/issues/17829

Until this is fixed, I will add a line print showing that Rustlings is running an exercise (not just frozen) and then make a release.

GlowCheese:
> I reported it in the Rust-Analyzer repo: [rust-lang/rust-analyzer#17829](https://github.com/rust-lang/rust-analyzer/issues/17829)

Thanks you so much for handling this issue for me. I'll keep an eye on it to see if I can help with anything!

mo8it:
[...]

PR #2077, opened 2024-08-08
Add pending UI

senekor:
This is a POC of [this suggestion](https://github.com/rust-lang/rustlings/issues/2071#issuecomment-2274679885). It's a little hacky, but as far as I understand it will be removed again once the issue with rust-analyzer is fixed. I think removing it is a good idea, because rust-analyzer is working fine for me and the pending UI is only shown for a very short amount of time.

Also, I added the pending UI at the very bottom, because if rust-analyzer isn't running at all the UI becomes very janky if any of the other text changes or moves.

### DIFF
--- a/src/watch/state.rs
+++ b/src/watch/state.rs
@@ -27,6 +27,7 @@ pub struct WatchState<'a> {
     show_hint: bool,
     done_status: DoneStatus,
     manual_run: bool,
+    pending: bool,
 }
 
 impl<'a> WatchState<'a> {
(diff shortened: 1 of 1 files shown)


mo8it:
Thanks, but there is no need to keep the a pending state. I solved by only adding this:

https://github.com/rust-lang/rustlings/blob/4ce8667b9d878dc48fafb665699a5fc71c190972/src/watch/state.rs#L54-L58

I will also keep it even after the Rust-Analyzer issue is resolved. It gives the user instant feedback which can be relevant on older PCs or with exercises with a long runtime.

senekor:
Much better 👍

PR #2092, opened 2024-08-31
Fix example in 'primitive_types3' hint

wugalde19:
The example provided in the hint for exercise `primitive_types3` does not work

**ACTUAL HINT**
image

If you try what the hint suggests, it fails
image
image

However, if you update the value to be 100, it works (`10` -> `100`)
image
image


mo8it:
Thanks :D

wugalde19:
> Thanks :D

@mo8it Anytime!

Issue #2093, opened 2024-08-31
Make the list searchable


frroossst:
When pressing `l` to bring up a list of exercises, I think it'd be a neat idea to press `/`and start searching, similar to vim. 

If there is  interest in this feature, I'd like to take a crack at implementing it. Thanks!

mo8it:
Good idea! You are welcome to implement it in a PR :D

Here are some tips for getting started:
- Handle the input `l` here: https://github.com/rust-lang/rustlings/blob/ea52c99560119070cb2c82c65dd92017240c5071/src/list.rs#L35
- Add `Option<String>` to `ListState`: https://github.com/rust-lang/rustlings/blob/ea52c99560119070cb2c82c65dd92017240c5071/src/list/state.rs#L37. `None` means that there is no active search. `Some(s)` means that the user is searching for an exercise containing `s` in its name.
- Extend the exercise iterator filter before displaying the rows: https://github.com/rust-lang/rustlings/blob/ea52c99560119070cb2c82c65dd92017240c5071/src/list/state.rs#L198-L204
- Add that filter to the selection logic as well: https://github.com/rust-lang/rustlings/blob/ea52c99560119070cb2c82c65dd92017240c5071/src/list/state.rs#L332-L354

Don't add any additional crates. `O(N)` shouldn't be dramatic here. I don't think that Rustlings will be ever used on more than 150 exercises.

No fuzzy search, just use [`str::contains`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.contains).

If you need any help, feel free to ask me :)

mo8it:
(_I deleted my comment because I have a better idea of how to implement it now_)

Good idea! You are welcome to implement it in a PR :D

I added `<s>earch` to the help footer in https://github.com/rust-lang/rustlings/commit/75a38fa38b65c075f34233f4745eb6d1d7405a39. So we should trigger a search by `s` and `/`.

Here are some tips for getting started:
- Handle the input `s` and `/` here: https://github.com/rust-lang/rustlings/blob/ea52c99560119070cb2c82c65dd92017240c5071/src/list.rs#L35
- Add a `String` field to `ListState` which contains the search input: https://github.com/rust-lang/rustlings/blob/ea52c99560119070cb2c82c65dd92017240c5071/src/list/state.rs#L37
- During search, you need to handle any input char you get and add it to the search string.
[...]

PR #2098, opened 2024-09-02
Made the list of exercises searchable, ref #2093

frroossst:


src/list.rs
```diff
@@ -32,8 +33,47 @@ fn handle_list(app_state: &mut AppState, stdout: &mut StdoutLock) -> Result<()>
...
+                            continue;
+                        }
+                        _ => {
+                            continue;
+                        }
+                    }
+                }
+
                 match key.code {
-                    KeyCode::Char('q') => return Ok(()),
```
mo8it, line 36: Why?

src/list.rs
```diff
@@ -32,8 +33,47 @@ fn handle_list(app_state: &mut AppState, stdout: &mut StdoutLock) -> Result<()>
...
                 list_state.message.clear();
 
+                let curr_key = key.code;
+
+                if is_searching {
+                    match curr_key {
+                        KeyCode::Esc | KeyCode::Enter => {
+                            is_searching = false;
+                            list_state.search_query.clear();
+                            continue;
```
mo8it, line 43: We continue in each match arm. It would be better to use `continue` at the end of the `if` body instead.

src/list.rs
```diff
@@ -32,8 +33,47 @@ fn handle_list(app_state: &mut AppState, stdout: &mut StdoutLock) -> Result<()>
...
+                        }
+                        KeyCode::Char(k) => {
+                            list_state.search_query.push(k);
+                            list_state.message.push_str("search:");
+                            list_state.message.push_str(&list_state.search_query);
+                            list_state.message.push('|');
+
+                            list_state.select_if_matches_search_query();
```
mo8it, line 51: Updating the message and then calling `select_if_matches_search_query` is duplicated. It would be better to replace `select_if_matches_search_query` with something like `apply_search_query` which also updates the message.

src/list.rs
```diff
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{Context, Ok, Result};
```
mo8it, line 1: As far as I can see, this isn't needed here.

src/list.rs
```diff
@@ -66,6 +106,10 @@ fn handle_list(app_state: &mut AppState, stdout: &mut StdoutLock) -> Result<()>
                             return Ok(());
                         }
                     }
+                    KeyCode::Char('s') | KeyCode::Char('/') => {
```
mo8it, line 109: You can use `KeyCode::Char('s' | '/')`

src/list.rs
```diff
@@ -66,6 +106,10 @@ fn handle_list(app_state: &mut AppState, stdout: &mut StdoutLock) -> Result<()>
...
+                        list_state.message.push_str("search:|");
```
mo8it, line 110: This can be replaced with `apply_search_query`. It should return early when `search_query` is empty.

src/list/state.rs
```diff
@@ -345,6 +347,39 @@ impl<'a> ListState<'a> {
...
+                Filter::Pending => !exercise.done,
+            })
+            .enumerate()
+            .find_map(|(i, s)| {
+                if s.name.contains(self.search_query.as_str()) {
+                    Some(i)
+                } else {
+                    None
+                }
+            });
```
mo8it, line 367: Please use [`position`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.position) instead of `enumerate` with `find_map`.

I also try to avoid such short variable names. Better use `exercise` instead of `s`.

src/list/state.rs
```diff
@@ -345,6 +347,39 @@ impl<'a> ListState<'a> {
...
+        match idx {
+            Some(i) => {
+                let exercise_ind = i;
+                self.scroll_state.set_selected(exercise_ind);
+                self.update_rows();
+            }
+            None => {
+                let msg = String::from("[NOT FOUND]") + &self.message.clone();
+                self.message.clear();
+                self.message.push_str(&msg);
```
mo8it, line 378: Maybe we can just push `" (not found)"` to the existing message.

src/list/state.rs
```diff
@@ -345,6 +347,39 @@ impl<'a> ListState<'a> {
...
+                } else {
+                    None
+                }
+            });
+
+        match idx {
+            Some(i) => {
+                let exercise_ind = i;
+                self.scroll_state.set_selected(exercise_ind);
+                self.update_rows();
```
mo8it, line 373: This call isn't required.

mo8it (changes requested):
Very good implementation so far. Only small changes are required :)

src/list.rs
frroossst: This already existed in the source code, [ref](https://github.com/rust-lang/rustlings/blob/main/src/list.rs#L36), if you are asking why I made it NOT a single line and added the `{` then those are remnants from me experimenting around, will fic it

src/list/state.rs
frroossst: removed!

src/list/state.rs
frroossst: done!

src/list/state.rs
frroossst: made the refactor, out of curiosity, is there a reason why position() is preferrable?

src/list.rs
frroossst: done

src/list.rs
frroossst: made the refactor!

src/list/state.rs
mo8it: Shorter and easier to understand what one is computing.

src/list/state.rs
```diff
@@ -345,6 +347,37 @@ impl<'a> ListState<'a> {
...
+            })
+            .position(|exercise| exercise.name.contains(self.search_query.as_str()));
+
+        match idx {
+            Some(exercise_ind) => {
+                self.scroll_state.set_selected(exercise_ind);
+            }
+            None => {
+                let msg = String::from(" (not found)");
+                self.message.push_str(&msg);
```
mo8it, line 376: No need to create a `String` here (it allocates). `.push_str(" (not found)")` is enough.

src/list.rs
```diff
@@ -32,6 +33,29 @@ fn handle_list(app_state: &mut AppState, stdout: &mut StdoutLock) -> Result<()>
 
                 list_state.message.clear();
 
+                let curr_key = key.code;
+
+                if is_searching {
+                    match curr_key {
+                        KeyCode::Esc | KeyCode::Enter => {
+                            is_searching = false;
+                            list_state.search_query.clear();
```
mo8it, line 42: We probably need to draw here to remove the search query. I will test it.

mo8it (approved):

mo8it:
I will take care of the two comments myself. Merging for now. Very well implemented. Thank you! :D

mo8it:
I made the final touches in https://github.com/rust-lang/rustlings/commit/da8b3d143a5b7462baf912c58cc768f7cd210ab2, mostly subjective changes :)

PR #2103, opened 2024-09-11
Remove redundant enum definition task

senekor:
The exercise enums2.rs already contains a task where an identical enum has to be defined.


mo8it (approved):

mo8it:
Thanks :)

Issue #2104, opened 2024-09-12
#[warn(dead_code)] showing unnecessary x is never used warnings


huss4in:
Since testing in some exercises is done using tests modules, annoying `function 'x' is never used warning` shows while and after solving the exercise...
![image](https://github.com/user-attachments/assets/c0d91e80-f251-4174-a299-f19002de02e4)

Which could sometimes be annoying or even come before/hide more relevant code lints related to the main exercise issue:

Before: 
![before](https://github.com/user-attachments/assets/d40e4418-e6ff-4543-9ce8-c6859e4f9002)
After: 
![after](https://github.com/user-attachments/assets/8de1a810-dadf-4f39-b1aa-a39fba419076)

A simple way to make it look cleaner is by adding `#![allow(dead_code)]` to all files which have `mod tests {`

```rust
| #![allow(dead_code)]

  fn fill_vec(vec: Vec<i32>) -> Vec<i32> {
      let mut vec = vec;

      vec.push(88);

      vec
  }
  ...
```

Or in a more generic/centralized way in `Cargo.toml`

```toml
[lints.rust]
usued = "allow"
```
```[tasklist]
### Solutions
- [ ] https://github.com/rust-lang/rustlings/pull/2105
- [ ] https://github.com/rust-lang/rustlings/pull/2106
```

mo8it:
Allowing dead code is just a workaround. The actual solution would be to tell Rust-Analyzer to consider the usage in tests.

Rustlings itself does so by passing `--profile test` to the used Cargo commands. I will see if I can pass this option to Rust-Analyzer.

mo8it:
@alibektas is the `rust-analyzer.toml` ready for usage? This issue can be resolved by setting `check.extraArgs` to `["--profile", "test"]`.

alibektas:
To answer your question : It is honestly not completely ready as we keep changing config classes a lot but I think in your case it is safe to say that we are not going to make any changes to config keys `check.*`.

huss4in:
Thanks for looking into the issue!

I would just like to shed some lights on the fact that using `--profile test` doesn't solve all cases
Example in `enums2`:
![image](https://github.com/user-attachments/assets/cfd02e17-0c1d-4c93-a9e1-b9c962d6ad16)

Compared to `allow(dead_code)`:
![image](https://github.com/user-attachments/assets/bb9872aa-2cfa-4968-9f5f-7df9378798c3)

[...]

PR #2105, opened 2024-09-12
(Solution 1) fix: showing unnecessary `x is never used` warnings

huss4in:
This is the first approach to solve #2104

If accepted the second PR #2106  shall be dropped as is solves the same issue in a different way..


mo8it:
The lint is called `dead_code`, not `unused`

PR #2107, opened 2024-09-12
Add rust-analyzer.toml file

### LINKED ISSUE #2104: #[warn(dead_code)] showing unnecessary x is never used warnings

alibektas:
Fixes #2104.

### DIFF
--- a/dev-rust-analyzer.toml
+++ b/dev-rust-analyzer.toml
@@ -1,0 +1,4 @@
+# rust-analyzer configuration file
+# DO NOT edit what is already defined.
+# You may add new configurations as needed.
+check.extraArgs = ["--profile", "test"]
--- a/dev-rust-analyzer.toml
+++ b/dev-rust-analyzer.toml
@@ -1,4 +1,0 @@
-# rust-analyzer configuration file
-# DO NOT edit what is already defined.
-# You may add new configurations as needed.
-check.extraArgs = ["--profile", "test"]
--- a/src/init.rs
+++ b/src/init.rs
@@ -130,8 +130,7 @@ pub fn init() -> Result<()> {
     fs::write("Cargo.toml", updated_cargo_toml)
         .context("Failed to create the file `rustlings/Cargo.toml`")?;
 
-    let ra_toml = include_str!("../dev-rust-analyzer.toml");
-    fs::write("rust-analyzer.toml", ra_toml)
+    fs::write("rust-analyzer.toml", RATOML)
         .context("Failed to create the file `rustlings/rust-analyzer.toml`")?;
 
     fs::write(".gitignore", GITIGNORE)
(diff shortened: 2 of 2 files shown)


alibektas:
Just my amateur solution to this problem. Let me know if you have any stylistic changes you want me to make :wave:

mo8it:
Thanks!

We need the small file in the repo. We can define the content in a `&[u8]` constant and use it like `GITIGNORE`: https://github.com/rust-lang/rustlings/blob/1f624d4c2a4ee0441ffce842591a8e37e329d309/src/init.rs#L176

I wonder if `extraArgs` will stay as is or get renamed to `extra_args` since camel case is common in TOML files.

alibektas:
> Thanks!
> 
> We don't need the small file in the repo. We can define the content in a `&[u8]` constant and use it like `GITIGNORE`:
> 
> https://github.com/rust-lang/rustlings/blob/1f624d4c2a4ee0441ffce842591a8e37e329d309/src/init.rs#L176
> 
> I wonder if `extraArgs` will stay as is or get renamed to `extra_args` since camel case is common in TOML files.

I don't think a renaming is among our plans. In either case we never remove old config names and check for them [too](https://github.com/rust-lang/rust-analyzer/blob/fd243cd0fbad8297e7a3c86f18aa5f0a110bb85f/crates/rust-analyzer/src/config.rs#L484).

mo8it:
Tested the config in a `rust-analyzer.toml` file with the Rust-Analyzer extention in VS-Code (pre-release). It doesn't seem to work (yet?). See the screenshot where the function `bigger` is underlined. The warning is that this function isn't used. I restarted Rust-Analyzer after creating the config file and even restarted the whole editor. It doesn't seems to use that config.

Screenshot from 2024-09-13 13-48-54

alibektas:
That's odd. I made the warning go away when I tried it in my local. I will have another look at it in a few hours.

mo8it:
@alibektas Is `.rust-analyzer.toml` (with dot) also supported? It would be nice to be able to hide it from beginners.

alibektas:
not yet and we didn't really plan such thing but I can see people asking for it so I will implement it in a few days. I am making an issue of it an pinging you.

Issue #2108, opened 2024-09-13
The if2 exercise is confusing


senekor:
I'm teaching a Rust course right now and I noticed that several people got hung up on this exercise. I know that the "fizz" and "foo" stuff isn't supposed to matter, but it's human nature to search for meaning in things. People were struggling to "turn off their brain" and just code against the tests.

I think it would be nice if the mapping from input to output values made at least some sort of sense. It doesn't have to be an elaborate story, just enough so people can focus on the Rust syntax as they are supposed to.

I'll try to come up with something suitable.

mo8it:
I agree. Feel free to open a PR :D

I might review PRs related to exercises later, but they are welcome, especially if they are related to feedback from your course :)

PR #2110, opened 2024-09-14
Make if2 less confusing

senekor:
Some people would get stuck on this exercise, trying to understand the meaning behind foo, fuzz, baz etc. Making the theme of the code make a little more sense to humans should hopefully prevent people from getting confused by abstract and non-sensical tests.


senekor:
fixes #2108

mo8it (approved):

mo8it:
Thanks :)

PR #2114, opened 2024-09-22
Do not use `.as_bytes().len()` on strings

samueltardieu:
`.len()` already returns the length in bytes when used on a `String` or a `str`. As this might be counter-intuitive for beginners, I've also added an extra comment. This has the benefit of preparing them to a proper use of `.len()` on a string.


mo8it (approved):

mo8it:
Thanks :)

PR #2122, opened 2024-10-02
Improvement to "check all exercises"

Nahor:
* Update the state of all the exercises instead of stopping with the first failure (especially given the old code was running all of them anyway, it was just ignoring the result of the later exercises). This allows the user to see all the exercises that needs to be redone instead of having to "fix/see one, run check_all, fix/see another, run check_all, ..."
* Fixed the progress display frozen for a long time if/when the first threads were run after later ones
* Limit the number of threads to the number of CPU cores (avoids spawning 94 cargo instances at the same time, possibly more in the future or in custom exercise sets)
* Show a progress bar instead of an unappealing "Progress: xxx/yyy".
* Add a command to run check all the exercises immediately (useful if they are solved without running `rustlings` in between)

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -13,4 +13,6 @@ disallowed-methods = [
   # Use `thread::Builder::spawn` instead and handle the error.
   "std::thread::spawn",
   "std::thread::Scope::spawn",
+  # Return `ExitCode` instead.
+  "std::process::exit",
 ]
--- a/src/app_state.rs
+++ b/src/app_state.rs
@@ -1,4 +1,9 @@
 use anyhow::{bail, Context, Result};
+use crossterm::{
+    queue,
+    style::{Print, ResetColor, SetForegroundColor},
+    terminal,
+};
 use std::{
     env,
     fs::{File, OpenOptions},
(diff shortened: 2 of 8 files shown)


mo8it (approved):

mo8it:
@Nahor Thanks! This is the biggest PR I reviewed since I became a maintainer. Thank you very much for implementing all of this :D

I added my changes and merged the PR. The most important change is the visualization of the progress. Check out the animation, I find it very fancy :D

I really appreciate all your work on this PR, but next time, please **discuss non-trivial changes in an issue first**. This applies to contributing to any open source project, not only Rustlings.
I had to learn it the hard way. You put so much effort in a PR and then it gets rejected by the maintainer because it doesn't align with their vision.
Don't force me or any open source maintainer to do that and cause what I had to experience 🤗

Nahor:
@mo8it It started trivial and mostly for me to practice Rust. Then I figured it looked cool enough to submit 😊.
Personally I don't mind rejections... unless the reason is lame, in which case a rejection through discussion wouldn't be any better anyway 😝. But sure, I'll try to keep that in mind.

PR #2125, opened 2024-10-07
Restriction by trait `Grade` so that the `ReportCard` has and ONLY ha…

1vk3y:
…s two types: numeric and alphabetic.

### DIFF
--- a/solutions/quizzes/quiz3.rs
+++ b/solutions/quizzes/quiz3.rs
@@ -9,9 +9,21 @@
 
 use std::fmt::Display;
 
-// Make the struct generic over `T`.
-struct ReportCard<T> {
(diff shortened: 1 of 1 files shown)


mo8it:
Using a marker trait is a good idea. But I think that it complicates the exercise and could leave the user confused because we only restrict the type. If we want take the direction of restricting what a grade can be, we should also add some form of validation.

1vk3y:
> Using a marker trait is a good idea. But I think that it complicates the exercise and could leave the user confused because we only restrict the type. If we want take the direction of restricting what a grade can be, we should also add some form of validation.

Yes, it's right and inspiring. THANKS

Issue #1935, opened 2024-03-30
Use `Cargo.toml` instead of `rust-project.json`


mo8it:
We can generate a `Cargo.toml` file containing a `[[bin]]` entry for every exercise.

This should make things much smoother and less hacky, especially because we have many issues related to the language server.

This should also enable us to use Cargo instead of rustc to run the exercises. That would especially make Clippy exercises less of a trouble. This approach also allows us to easily add dependencies to the exercises in case we want to do so in the future.

mo8it:
Done in the `v6` branch :D

antoineco:
@mo8it I just tried the `v6` branch and love the improvements from this PR, so first of all thank you for the impressive work!

I have a question regarding this statement:
>  we have many issues related to the language server

If anything, the current state of `v6` seem to actually _degrade_ the language server experience. Inside a fresh Git clone of the branch on Linux, `rust-analyzer` (latest stable) no longer emits diagnostics while editing an exercise.
Are there still any upcoming changes related to language server for `v6`, or am I doing something wrong?

mo8it:
Thanks for your feedback 🥰

Yes, you are right, if you just clone the repository and try edit the exercises, the language server will not work. This is one downside of the current approach. But this only affects developing exercises.

The new method of doing Rustlings is to install Rustlings using `cargo install rustlings` (not published yet), then running `rustlings init`. No repo cloning happens. Instead, the directory `rustlings/` will be created where you find the exercises. The language server works there out of the box :)

antoineco:
Indeed `Cargo.toml` looks a lot more like what I'd expect now, appreciate the tip about `rustlings init` 👍

I spotted one issue though; out of the box, the language server is still showing the following diagnostic for every exercise file:
```
file not included in crate hierarchy [unlinked-file]
```

This is solved by editing the exercises' paths inside `Cargo.toml`:
```diff
 bin = [
- { name = "intro1", path = "../exercises/00_intro/intro1.rs" },
+ { name = "intro2", path = "exercises/00_intro/intro2.rs" },
- { name = "intro1", path = "../exercises/00_intro/intro1.rs" },
+ { name = "intro2", path = "exercises/00_intro/intro2.rs" },
 ...
```

mo8it:
Good catch! Thank you very much :D

It was without the prefix `../`, but I messed it up at some point 😅
[...]

PR #1026, opened 2022-06-16
Add lsp command to fix rust-analyzer

jackos:
This is continuing from #911, as requested I've moved functionality into a subcommand that can be run like `rustings lsp` and added documentation plus added it to the initial welcome message.

This will ensure a good developer experience for newcomers to the Rust ecosystem.

PR #2131, opened 2024-10-14
Add warning about `rust-analyzer` not working if you clone and use the repo directly

MadLittleMods:
Add warning about `rust-analyzer` not working if you clone and use the repo directly.

> Yes, you are right, if you just clone the repository and try to edit the exercises, the language server will not work. This is one downside of the current approach. But this only affects developing exercises.
>
> The new method of doing Rustlings is to install Rustlings using `cargo install rustlings` (not published yet), then running `rustlings init`. No repo cloning happens. Instead, the directory `rustlings/` will be created where you find the exercises. The language server works there out of the box :)
>
> I need to add a warning when people try to work on the exercises from the repository. Thanks pointing this out.
>
> -- @mo8it, https://github.com/rust-lang/rustlings/issues/1935#issuecomment-2067664066

### Reproduction steps

Personally, I also fell into this trap since I prefer to just run the thing from source than install something on my system.

 1. Clone the repo: `git clone git@github.com:rust-lang/rustlings.git`
 1. Start Rustlings: `cargo run` 
 1. Open VSCode and notice that `rust-analyzer` and lsp hints don't pop up
 1. Google and find [StackOverflow answers](https://stackoverflow.com/questions/75272113/rust-analyzer-in-vscode-does-not-work-in-rustlings-workspace) that mention `rustlings lsp` to setup lsp support
 1. But `rustlings lsp` doesn't exist anymore and the [changelog mentions that it should be supported out of the box](https://github.com/rust-lang/rustlings/blob/0432e07864d5ee4d2bac1954c965b2077c0447c6/CHANGELOG.md#lsp-support-out-of-the-box). It's also confusing because the `rustlings` repo has it's own `Cargo.toml` with `include = [... "/exercises/", ...]` which to my newbie Rust eyes, seems like it might cover it and makes me think, this is supposed to work.
 1. Eventually, I saw https://github.com/rust-lang/rustlings/issues/1935 linked somewhere and it mentions this exact problem

### Dev notes

 - Previous `rustlings lsp` command: https://github.com/rust-lang/rustlings/pull/1026
 - The changelog says "LSP support out of the box", https://github.com/rust-lang/rustlings/blob/main/CHANGELOG.md#lsp-support-out-of-the-box

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,15 @@ rustlings
 Our general recommendation is [VS Code](https://code.visualstudio.com/) with the [rust-analyzer plugin](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer).
 But any editor that supports [rust-analyzer](https://rust-analyzer.github.io/) should be enough for working on the exercises.
 
+<details>
+<summary><strong>If you cloned the repository and <code>rust-analyzer</code> isn't working…</strong> (<em>click to expand</em>)</summary>
+
+The intended way to run Rustlings is to install the binary and run `rustlings init` as described above. This generates a `Cargo.toml` (different than what you see in the repository) that includes the each excersise as a separate binary target which is enough for `rust-analyzer` to work.
+
+If you just clone the repository and try to run and edit the exercises directly, the language server will not work. This is one downside of the current approach. But this only affects developing exercises.
+
+</details>
+
 ### Terminal
 
 While working with Rustlings, please use a modern terminal for the best user experience.


mo8it:
I agree that we should add something to the README for this, but the addition should tell people to not clone the repository and instead follow the instructions above. A simple warning should be enough. Something like this:

"Don't try to clone the repository to do the exercises! Rust-Analyzer won't work in that case. Please follow the instructions above instead."

README.md
```diff
@@ -45,6 +45,18 @@ cargo install rustlings
 
 </details>
 
+> [!CAUTION]
```
MadLittleMods, line 48: I've used caution but feel free to nit to warning or [any other](https://github.com/orgs/community/discussions/16925).

README.md
```diff
@@ -45,6 +45,18 @@ cargo install rustlings
...
+> Don't try to clone the repository to do the exercises! `rust-analyzer` won't work in that case. Please follow the instructions above instead.
```
MadLittleMods, line 49: Updated to your exact wording :white_check_mark: 

I've kept the extra "Why?" background in a collapsible section since it feelslike adds a little bit of something that makes it more understandable and convincing. Happy to remove to get this shipped if you prefer.

mo8it:
I don't think that this is needed anymore. Cloning is deprecated for a while now.

PR #2138, opened 2024-10-19
Add a method in solutions/18_iterators/iterators4.rs

1vk3y:
Add a method that solves the problem: using recursion combined with match to implement factorial.


mo8it:
This isn't Haskell :P
In most cases, iterative solutions are more effecient than recursive ones. I don't want to recommend recursion where it isn't needed.

1vk3y:
Touché. :D

PR #2144, opened 2024-10-29
grammatical fix on move_semantics2.rs

elian1780:
// TODO: Make both vectors `vec0` and `vec1` accessible at the same time to
changed to 
// TODO: Make both vectors `vec0` and `vec1` are accessible at the same time to


mo8it:
The sentence is correct. It already has the verb "Make".

Issue #2157, opened 2024-11-13
errors6 feedback


mnshdw:
errors6 suggests adding a function to map the error, that has then to be called with `map_err`.

``` 
    // TODO: Add another error conversion function here.
    // fn from_parse_int(???) -> Self { ??? }
``` 

```
        let x: i64 = s.parse().map_err(ParsePosNonzeroError::from_parse_int)?;
        //                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It seems to me that it would be more idiomatic to implement `From<ParseIntError>`, so we could get away with a simpler `s.parse()?`.

Is there a specific reason this is not the suggested solution here?

mo8it:
> It seems to me that it would be more idiomatic to implement `From<ParseIntError>`, so we could get away with a simpler `s.parse()?`

True. But `errors6` is before the exercises about traits and conversions with `From`.

I think that we should add the implementation of `From<ParseIntError>` as an additional and preferred way to the solution. A comment should mention that traits like `From` are dealt with in later exercises.

Do you want to create a PR?

mnshdw:
> But errors6 is before the exercises about traits and conversions with From.
Ah! Good point.

> Do you want to create a PR?
Done. https://github.com/rust-lang/rustlings/pull/2158

PR #2158, opened 2024-11-13
errors6: Add alternative solution using From trait

mnshdw:
Discussed in https://github.com/rust-lang/rustlings/issues/2157

Let me know if you think this should be reworded or something else should be added.

Thanks!


mo8it:
Awesome, thanks!

Could you please replace `///` with `//` for the comments? Then I will merge :)

PR #2166, opened 2024-12-01
Fix rust-book links

joaolsouzajr:
Update rust-book links to use `https://doc.rust-lang.org/stable/book` instead of `https://dosc.rust-lang.org/book` as URL


joaolsouzajr:
Sorry It is not a real  issue, the wrong links is just generated by `rustlings init` on version installed by cargo `rustlings 6.4.0` .

PR #2167, opened 2024-12-02
Do not suggest vec macro for vecs1.rs

GrimaceOfDespair:
The solution uses the to_vec function rather than the vec macro

See also https://stackoverflow.com/a/75230582/281084

### DIFF
--- a/exercises/05_vecs/vecs1.rs
+++ b/exercises/05_vecs/vecs1.rs
@@ -2,7 +2,6 @@ fn array_and_vec() -> ([i32; 4], Vec<i32>) {
     let a = [10, 20, 30, 40]; // Array
 
     // TODO: Create a vector called `v` which contains the exact same elements as in the array `a`.
-    // Use the vector macro.
     // let v = ???;
 
     (a, v)


exercises/05_vecs/vecs1.rs
```diff
@@ -2,7 +2,6 @@ fn array_and_vec() -> ([i32; 4], Vec<i32>) {
     let a = [10, 20, 30, 40]; // Array
 
     // TODO: Create a vector called `v` which contains the exact same elements as in the array `a`.
-    // Use the vector macro.
```
foxfromworld, line 5: Thanks for the PR! I understand the goal is to align the hints with the solution. However, as a learner, I think vec! is such a foundational tool that it's worth keeping in the first Vector exercise. Maybe we could introduce .to_vec() as a separate example later instead? This way, beginners learn the most common macro first while still getting to practice array-to-Vec conversion later. 🙂

senekor: The book teaches the `vec!` macro, so I agree we should keep it.

senekor:
> The solution uses the to_vec function rather than the vec macro

This is not true? The solution uses `vec!`:
https://github.com/rust-lang/rustlings/blob/main/solutions/05_vecs/vecs1.rs#L5

PR #2181, opened 2024-12-26
Update structs3.rs

br8bit:
Refactored the code to encapsulate the cents_per_gram field within the `Package` struct, improving architectural clarity and maintainability. Additionally, updated the related functions and tests to align with this restructuring


mo8it:
I prefer the old way. These exercises are tiny, it is not about maintainability.

PR #2188, opened 2025-01-01
A clearer description of the original problem

jbethune:
The original excercise instructs the learner to use the vector macro. It is easy to assume that the code needs to read from `a` which is not what is intended.

The updated excercise introduces the syntax for initial array and vector contents and lets the learner figure out how to tweak initial values of a vector. This learning seems to be the original intent of the excercise.

### DIFF
--- a/exercises/05_vecs/vecs1.rs
+++ b/exercises/05_vecs/vecs1.rs
@@ -1,9 +1,13 @@
 fn array_and_vec() -> ([i32; 4], Vec<i32>) {
-    let a = [10, 20, 30, 40]; // Array
+    // You can define an array with the intitial values 10, 20, 30 and 40 like
+    // this:
+    let a = [10, 20, 30, 40]; // Array. Do not change!
 
-    // TODO: Create a vector called `v` which contains the exact same elements as in the array `a`.
-    // Use the vector macro.
-    // let v = ???;
+    // There is a similar way you can define a vector with initial values:
+    let v = vec![20, 30, 40]; // Vector. Needs to be fixed
+
+    // TODO: Adjust the vector definition above so that `a` and `v` have the
+    // same contents.
 
     (a, v)
 }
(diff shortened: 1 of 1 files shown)


exercises/05_vecs/vecs1.rs
```diff
@@ -1,9 +1,13 @@
 fn array_and_vec() -> ([i32; 4], Vec<i32>) {
-    let a = [10, 20, 30, 40]; // Array
+    // You can define an array with the intitial values 10, 20, 30 and 40 like
+    // this:
+    let a = [10, 20, 30, 40]; // Array. Do not change!
```
foxfromworld, line 4: The original example already shows it’s just array initialization, so it works well. Still, it might be good to see the maintainer’s opinion.

exercises/05_vecs/vecs1.rs
```diff
@@ -1,9 +1,13 @@
...
 
-    // TODO: Create a vector called `v` which contains the exact same elements as in the array `a`.
-    // Use the vector macro.
-    // let v = ???;
+    // There is a similar way you can define a vector with initial values:
+    // let v = vec![10, 20, ???]; // Vector. Needs to be fixed
```
foxfromworld, line 7: Both to_vec() and Vec::from compile, but the author just wanted vec![10, 20, 30, 40]. So I think the original is fine — let’s see the maintainer’s opinion.

foxfromworld (commented):
This is related to #2167.

senekor:
I think this PR is over explaining stuff, but I can see how the current state is confusing to some. Duplicating the literals is not something people would do in a real code base, so the mind doesn't jump to that. Maybe it's a good idea to redesign the exercise a little, so the `vec!` macro is the more intuitive solution.

mo8it:
Resolved by https://github.com/rust-lang/rustlings/pull/2397

Issue #2163, opened 2024-11-24
[FR] provide previous key to go back to previous exercise


thkoch2001:
First please reopen issue https://github.com/rust-lang/rustlings/issues/2111 and once this is solved, please also provide a way to quickly go back to the previous exercise without the detour to the exercises list.

This might be a nice feature request for a new contributor?

mo8it:
Good idea!

Yeah, it would be a good first issue. You are welcome to create a pull request :)

doublevoid:
> Good idea!
> 
> Yeah, it would be a good first issue. You are welcome to create a pull request :)

Would it be ok if I worked on this? I have a version of this working locally, but not sure if this issue is open to new PRs from everyone or just from the issue OP.

doublevoid:
Got invested in it, polished it up and created a PR for this:
https://github.com/rust-lang/rustlings/pull/2189

senekor:
(Copied from my comment on the PR:) I don't think users will often want to go back to the previous exercise, right? And there's a pretty easy way to do it right now. `l` for list, ⬆️ for the previous exercise, `c` to continue from there.

mo8it:
After thinking about it again, I think that it might be very unintuitive. We skip done exercises when going to the next one. What are we supposed to do when going back? Go back to the latest one? What if the users reopens Rustlings after a break of some days. Should it still go back to the latest or just to the actual previous one?
Adding this reason to the comment by @senekor, I think we should not add this option.

PR #2189, opened 2025-01-02
Add command to go back to previous exercise, ref #2163

doublevoid:
Adds `p` command/hotkey for going back to a previous exercise.

Might not be fully ready to merge, this is my first Rust PR, will be happy to work/iterate where needed.


senekor:
I don't think users will often want to go back to the previous exercise, right? And there's a pretty easy way to do it right now. `l` for list, ⬆️ for the previous exercise, `c` to continue from there.

mo8it:
See the comment on the issue

PR #2191, opened 2025-01-02
adding file IO exericses

### LINKED ISSUE #482: Stdio, Read, Write and Files

frroossst:
Once approved and merged this should close #482

### DIFF
--- a/exercises/24_fileIO/fileIO1.rs
+++ b/exercises/24_fileIO/fileIO1.rs
@@ -1,0 +1,57 @@
+use std::fs::File;
+use std::io::Read;
+
+fn my_file_reader() -> String {
+
+    // TODO: Open the file in read-only mode 
+    let mut file = File::???("ferris.txt").unwrap();
+    let mut content = String::new();
+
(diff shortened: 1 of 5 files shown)


frroossst:
Sorry, for being a tad bit late to open this PR, December was finals month so super busy.

I know this implementation isn't quite ready to be merged just yet. Looking for pointers on improving this to make it merge-worthy.

irfanghat:
@frroossst I'm really forward to these. I hope they progressively get hard so people like me can enforce the concepts. I constantly have to look up reading multiple lines from standard input because I get it wrong for some reason each time.

foxfromworld (commented):
In addition to open and create, the maintainer previously suggested including an exercise for 'Appending to an existing (log) file.' It might be worth adding this case as well, since it would be a great way to introduce OpenOptions. What do you think?

frroossst:
Been inactive, but looking to pick this back up soon.

PR #2192, opened 2025-01-02
Fix 404 in Lifetimes README.md

Akaito:
Link to Lifetimes (in Rust By Example) leads to a 404 page.  Removing the "/stable" slug fixes this.


Akaito:
Disregard.  When using the "/stable" version of the link, I didn't notice the last two characters had been truncated when using a mouse to follow the link in a wrapped line, due to Windows Terminal behavior (not having awareness of the underlying text).

PR #2195, opened 2025-01-12
Add errors7 exercise to show full potential of try operator combined with From trait

gdziadkiewicz:
Thanks for creating the Rustlings! I loved it!

While doing them I felt that the user might miss the full potential of the try operator due to the lack of exercise that shows how well custom error types work with it and From trait. This PR adds an exercise focused on that. I placed it after the first trait exercise to be sure that they understand what they are doing.

Only after creating the thing did I find out that I should have created an issue first - sorry for that!

If it goes in let's squash it.

### DIFF
--- a/exercises/13_error_handling/errors7.rs
+++ b/exercises/13_error_handling/errors7.rs
@@ -1,0 +1,97 @@
+// TODOUsing catch-all error types like `Box<dyn Error>` isn't recommended for
+// library code where callers might want to make decisions based on the error
+// content instead of printing it out or propagating it further. Here, we define
+// a custom error type to make it possible for callers to decide what to do next
+// when our function returns an error.
+
+use std::num::ParseIntError;
+
+#[derive(PartialEq, Debug)]
+enum CreationError {
+    Negative,
+    Zero,
+}
+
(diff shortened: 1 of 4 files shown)


mamalieaz:
salam

exercises/13_error_handling/errors7.rs
```diff
@@ -0,0 +1,100 @@
...
+
+// TODO Implement From trait for ParseIntError
+// impl From<ParseIntError> for ParsePosNonzeroError {
+//     fn from(err: ParseIntError) -> Self {
+//        ???
+//     }
+// }
+
+// TODO Implement From trait for CreationError
+// ...
```
foxfromworld, line 40: To help beginners, perhaps we could also provide a commented-out template for this implementation, similar to the one used for ParseIntError, if you think it would be useful.

senekor:
I don't think we should add this. It's too advanced, the book doesn't even cover it in this detail. It's also strange to go back to error handling after the traits chapter.

gdziadkiewicz:
Hi Senkor, thanks for the feedback! Your point about being too advanced and weak book coverage are valid. I will close the PR.

PR #2202, opened 2025-01-21
fix: clippy error message not inline with the intent.

### LINKED ISSUE #2201: Clippy3: error message from clippy not inline with the intent
In this part of code clippy error message not inline with the intent.

https://github.com/rust-lang/rustlings/blob/fbfd4f25e7e715007f5f3f6678f5d336d24d3660/exercises/22_clippy/clippy3.rs#L20-L21

Image

Because the solution stated that this is about using `clear` instead of `resize`.

https://github.com/rust-lang/rustlings/blob/fbfd4f25e7e715007f5f3f6678f5d336d24d3660/solutions/22_clippy/clippy3.rs#L20-L23

sugab:
Fix #2201

Screenshot 2025-01-21 at 11 34 23

### DIFF
--- a/exercises/22_clippy/clippy3.rs
+++ b/exercises/22_clippy/clippy3.rs
@@ -17,7 +17,8 @@ fn main() {
     ];
     println!("My array! Here it is: {my_arr:?}");
 
-    let my_empty_vec = vec![1, 2, 3, 4, 5].resize(0, 5);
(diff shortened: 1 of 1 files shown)


melvinius (approved):

senekor:
reason for closing:
https://github.com/rust-lang/rustlings/issues/2201#issuecomment-4418354087

PR #2204, opened 2025-01-29
feat: open current exercie via `--editor`

pacexy:


src/watch/state.rs
```diff
@@ -199,6 +199,7 @@ impl<'a> WatchState<'a> {
         show_key(b'l', b":list / ")?;
         show_key(b'c', b":check all / ")?;
         show_key(b'x', b":reset / ")?;
+        show_key(b'e', b":edit / ")?;
```
mo8it: This should only be shown if `--editor` is passed

PR #2212, opened 2025-02-24
Add proper testing for cases where divisions fail in iterators3

### LINKED ISSUE #2207: iterators3 tests dont check if errors are returned correctly
Im currently doing rustlings and struggled with iterators3 and searched a solution, which i thought i found [here](https://github.com/wh5a/rustlings-solutions/blob/main/exercises/18_iterators/iterators3.rs).

It implements one of the functions to edit like this:

 ```
fn result_with_list() -> Result<Vec<i64>, DivisionError> { 
let numbers = vec![27, 297, 38502, 81];
    let division_results = numbers.into_iter().map(|n| divide(n, 27));
    division_results.filter(|x| x.is_ok()).collect()
}
```

which compiles and passes the implemented test but this didnt make any sense to me, as this can never return a correct Err variant, at "best" this would return and empty Vec<i64>, if every division returns an Err variant, which is, as far as i understand, not the intendet solution.

To remedy this, I would suggest adding test cases to cover this wrong solution, but since the functions that get tested here dont take any parameters, but instead have hardcoded values inside them, this would require changing their signature and passing in the vector of numbers to be divided.

Im new to rust, so please let me know if this isnt a real issue.

If this is indeed a real issue, im happy to suggest an updated version of this exercise.

Lars-Schumann:
Fix #2207

### DIFF
--- a/exercises/18_iterators/iterators3.rs
+++ b/exercises/18_iterators/iterators3.rs
@@ -15,17 +15,17 @@ fn divide(a: i64, b: i64) -> Result<i64, DivisionError> {
 }
 
 // TODO: Add the correct return type and complete the function body.
-// Desired output: `Ok([1, 11, 1426, 3])`
-fn result_with_list() {
-    let numbers = [27, 297, 38502, 81];
-    let division_results = numbers.into_iter().map(|n| divide(n, 27));
+// For numbers = [27, 297, 38502, 81] and divisor = 27,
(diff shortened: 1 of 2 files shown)


Halloloid (commented):
LGTM! The changes correctly test both Result<Vec<_>, E> and Vec<Result<_, E>> collection behavior, including the error cases.

PR #2216, opened 2025-03-08
Async updates and reformatting for new exercise

Polycarbohydrate:
- Renamed/ordered the directories for macros, clippy, and conversions, shifting them all one up in numerical order to fit async as number 21
- Added the 21_async directories in exercises and solutions
- Added async1.rs in the 21_async directory in exercises and solutions
- Added async to the info.toml in rustlings-macros and reordered the rest of the exercises in there as well
- Created an exercise for async1 and its solution

The first exercise demonstrates the concept of adding the `async` keyword to the function signature to declare the function as asynchronous. It introduces the `tokio` crate with its `sleep` function and uses futures with `await`.


Polycarbohydrate:
I saw the file in dev\Cargo.toml and knew the file would be updated by running `cargo run -- dev update` but checking the Contributing.md (which is extremely outdated), it did not state to run that command then pushing those files into the repository.

mo8it:
Closing because it is not clear why async is needed in this exercise. We are working on a set of async exercises.

PR #2231, opened 2025-04-06
clarified the comment in exercise

### LINKED ISSUE #2230: Enhancement (09_strings/strings2.rs): confusing comment
### When I was doing rustlings, I got confused in the comment of this code. I will put here the whole rustling since it is not big:

```rust
// TODO: Fix the compiler error in the `main` function without changing this function.
fn is_a_color_word(attempt: &str) -> bool {
    attempt == "green" || attempt == "blue" || attempt == "red"
}

fn main() {
    let word = String::from("green"); // Don't change this line.

    if is_a_color_word(word) {
        println!("That is a color word I know!");
    } else {
        println!("That is not a color word I know.");
    }
}

```

`Main confusion cause` is the word `this`. At first I thought that this might relate to main function because it goes right after it being mentioned in the sentence, then I thought that it might mean function that is directly below.

I believe it is possible to completely eliminate the confusion. 

---

`Suggestion:` 
 - replace `this` with the actual name of the funciton `is_a_color_word`
> the whole comment would then look like: 
```rust
// TODO: Fix the compiler error in the `main` function without changing `is_a_color_word` function.
```

akshatcoder-hash:
This Pull request fixes #2230 
Replaced `this` with the actual name of the function `is_a_color_word` which was causing confusion for the users

### DIFF
--- a/exercises/09_strings/strings2.rs
+++ b/exercises/09_strings/strings2.rs
@@ -1,4 +1,5 @@
 // TODO: Fix the compiler error in the `main` function without changing this function.
+
 fn is_a_color_word(attempt: &str) -> bool {
     attempt == "green" || attempt == "blue" || attempt == "red"
 }


rahmatnazali:
Hi, maybe what the issue suggests is more like this:

```rust
// TODO: Fix the compiler error in the `main` function without changing `is_a_color_word` function.
fn is_a_color_word(attempt: &str) -> bool {
    attempt == "green" || attempt == "blue" || attempt == "red"
}
```

mo8it:
AI reported…

akshatcoder-hash:
> AI reported…

i'm sorry, could you please clarify

PR #2243, opened 2025-05-16
feat: add async exercises

manyinsects:
Adds two new exercises about async/await and handling concurrency with Tokio. I shoved these inbetween threads and macros, as that's where they are in the book (and having done threads is useful for `async2` and general understanding of concurrency in Rust).

`async1`:

- Uses minimal Tokio functionality
- Learn how to declare an async function
- Learn how to call and await an async function
- Learn about function coloring (just not under that term)

`async2`:

- Uses `tokio::task::JoinSet`
- Learn how to handle a set of Tokio tasks
- Learn about async closures

Happy to change anything at all about these exercises, this is just my two cents on what would be useful to learn during the course, without being _too_ complex.


senekor:
I've added my own proposal for a first async exercise in #2382. My aim was to use a little story to hopefully provide an intuition for why async is needed, or at least makes sense. I recall from my own learning that I struggled with abstract concepts like "worker", "tasks" and so on.

PR #2244, opened 2025-05-16
fix typo in structs3.rs

DrEverr:


mo8it:
We are referring to the [type signature](https://en.wikipedia.org/wiki/Type_signature) of the function :)

PR #2247, opened 2025-05-17
Website

mo8it:
TODO:

- [x] New website CI
- [x] Final edits, especially to the custom exercises page
- [x] Update links, especially in changelog


README.md
```diff
@@ -1,165 +1 @@
...
-```bash
-cargo uninstall rustlings
-```
-
-## Contributing
-
-See [CONTRIBUTING.md](https://github.com/rust-lang/rustlings/blob/main/CONTRIBUTING.md) 🔗
-
-Thanks to [all the wonderful contributors](https://github.com/rust-lang/rustlings/graphs/contributors) ✨
+# ➡️ [rustlings.rust-lang.org](https://rustlings.rust-lang.org) ⬅️
```
manyinsects, line 1: I would _at least_ keep a normal header, and a short description blurb, and maybe the one-liner installer. I feel like removing everything kind of messes with people who arrive here through direct link to GitHub (which is most links to Rustlings)

website/justfile
```diff
@@ -0,0 +1,5 @@
+zola:
+	zola serve --open
+
+tailwind:
+	fnm exec --using latest npx @tailwindcss/cli -w -i input.css -o static/main.css
```
manyinsects, line 5: nit: could this be changed to respect a `.node-version` or the `engines` field of `package.json`, or something like that?

manyinsects: just not sure if the recommended workflow should be having to use `fnm`, I feel like using normal `npx` and then guaranteeing a consistent version via some other means is better

manyinsects (changes requested):
couple of comments, but looks good otherwise!

README.md
mo8it: You are right. I added more context before the link :)

website/justfile
mo8it: `fnm` is just what I personally use. I removed it. It uses `npx` directly now.

Issue #2241, opened 2025-04-30
Incomplete Solution for quiz3


SAVYAGOEL:
In quiz3 solution, I think it is incomplete, since ideally we shouldn't even allow to create a ReportCard with the grade attribute having value of type other than those that implement the Display trait, so in my opinion in the complete solution we should also put a Trait Bound in the definition of the struct ReportCard and modify it to struct ReportCard<T: Display>

thely314:
(I used AI for translation because my native language is not English, and I hope it won't cause any misunderstandings)
You're correct from a technical perspective -- adding the trait bound at the struct level (`struct ReportCard<T: Display>`) would be a completely valid approach and in many cases would represent excellent practice.

In my opinion, keeping the struct definition unconstrained in the solution may be for specific tutorial reasons: 

Focus on method-level constraints: This quiz may primarily aims to demonstrate how trait bounds can be applied to implementations rather than struct definitions;
Showcasing Rust's flexibility: This quiz may also wants to highlight how Rust allows different constraint granularity (struct-level vs method-level)
Maybe other contributers can cover struct-level trait bounds in subsequent new exercises? I guess.

Also I would like to submit a PR that only adds explanatory comments to the solution :)

Vergesst:
Well, it will be better to add a Trait Constraint for `ReportCard`, I think. When it comes to ReportCard, most of us will regard it as a symbol of identity, just like id card... So it is sensible to change it into `ReportCard<T: Display>` or `ReportCard<T> where T: Display`, since that is the most common case when we use ReportCard...
So if this quiz is about flexbility of rust, a wrapper or something else used to wrap structs will be better, I think.

senekor:
I don't think there's a need to change anything here. Adding the trait bound could also be confusing to other people. "Why is there a trait bound here that's not necessary?" The current state is also not problematic, at worst it causes people to think about the difference between adding and omitting the bound.

PR #2253, opened 2025-06-03
Add additional explain for quiz3 solution

thely314:
Adds explanatory comments for `quiz3 solution` to avoid potential confusion for beginners on trait bound design (struct-level vs method-level).
#2241 

I used AI for translation because my native language is not English, and I hope it won't cause any misunderstandings, which means that the comments I added need to be check.


mo8it:
Sorry, I think this is too much text for a hint.

PR #2256, opened 2025-06-13
Added fold method to solution of iterators5

vallops99:
Another implementation of count_collection_iterator that could be useful for beginners in order to explore different ways of achieving the same thing.
This just avoids the "sum" method and directly sum each value returned from `count_iterator` with an accumulator


mo8it:
This is how `sum` is implemented. `fold` is more powerful, but also harder to understand. Therefore I will not add it to this solution.

PR #2262, opened 2025-07-18
Fix incorrect book chapter number

aCactusOrSomething:
Makes the README within the exercises file reference the correct section of the book for macros (20.5, corresponding to [https://doc.rust-lang.org/book/ch20-05-macros.html](https://doc.rust-lang.org/book/ch20-05-macros.html) instead of 19.5).

### DIFF
--- a/exercises/README.md
+++ b/exercises/README.md
@@ -22,6 +22,6 @@
 | iterators              | §13.2-4             |
 | smart_pointers         | §15, §16.3          |
 | threads                | §16.1-3             |
-| macros                 | §19.5               |
+| macros                 | §20.5               |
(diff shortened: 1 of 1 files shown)


aCactusOrSomething:
Upon further inspection, this table also has an additional potentially incorrect reference under clippy: it points to section 21.4 (which doesn't exist in the current version of the rust book). 

My guess would be that this should point to the appendix in 22.4, which has a section about clippy [here](https://doc.rust-lang.org/book/appendix-04-useful-development-tools.html#more-lints-with-clippy), but I haven't gotten to the material in this section of rustlings yet, so I'm not sure.

Issue #2174, opened 2024-12-17
VSCode: File links confusingly cause their files to be opened in OS default editor


laundmo:
Opening a terminal file link in VSCode terminal causes the file to be opened in the OS default application for the file type.

This leads to confusing behaviour where ctrl+click on the paths in a Rust error message files opens in VSCode but clicking the solution link opens in a different editor.

Possible solution could be detecting editors (like VSCode) which have their ow in path handling? sounds less than ideal...

PR #2267, opened 2025-08-01
feat: vscode protocol for vscode users

manmartgarc:
Solves #2174

The protocol change is not necessary for Windows users; I just tested it on VSCode on Windows 11 and it works fine, i.e. the files are opened on VSCode.

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -5,7 +5,7 @@ use crossterm::{
     terminal::{Clear, ClearType},
 };
 use std::{
-    fmt, fs,
+    env, fmt, fs,
     io::{self, BufRead, StdoutLock, Write},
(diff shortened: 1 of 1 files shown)


mo8it:
It didn't on VS Code on Linux. Therefore, I ended up with another solution which disables file links in VS Code since it has its own (weird and picky) file link handling: https://github.com/rust-lang/rustlings/commit/2d1d531550afaac36dbf4d15c383bc0e1cbf248d

manmartgarc:
I tested this on Ubuntu 22.04 and it worked fine, I *also* tested on Windows - what didn't work for you? Did it also ask you to select which program to open the file with?

PR #2270, opened 2025-08-07
Improve comment clarity for enums1 exercise

E-Geraet:
This PR updates the comments in enums1.rs to better explain the enum variants and make the exercise instructions clearer for beginners.


mo8it:
I am sorry, I don't think that the added comment is required. Even if it means that the user is stuck for a minute, I think it is worth finding out how enums work on yourself.

PR #2287, opened 2025-09-13
Add tooling for GitHub Codespaces

kaladron:
With this, a new user can fork Rustlings and start a Github Codespace.  The Container will already be correctly set up with everything needed, and VSCode tasks are set up for starting rustlings.


mo8it:
We want people to install Rust and run Rustlings on their machine.

kaladron:
I wanted to provide this because it's an easy way for someone to get started to get someone to try it easily.

PR #2297, opened 2025-10-02
Update from_into exercise to use error handling instead of default value

### LINKED ISSUE #2225: Incorrect answer passed in from_into.rs
I used `split_once` to split string by the comma into two parts. I didn't check the situation where N ≥ 3. The tests are surprisingly passed! 

I think [the tests](https://github.com/rust-lang/rustlings/blob/7c0d269279aaed64d6b8240ed5c1e9f6a981181e/exercises/23_conversions/from_into.rs#L125) aren't good enough and not thorough to test the condition that N ≥ 3.

iaadi4:
closes #2225

### DIFF
--- a/exercises/23_conversions/from_into.rs
+++ b/exercises/23_conversions/from_into.rs
@@ -3,24 +3,27 @@
 // You can read more about it in the documentation:
 // https://doc.rust-lang.org/std/convert/trait.From.html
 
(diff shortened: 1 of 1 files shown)


senekor:
This is completely misguided. There is a separate exercise to teach the `TryFrom` trait.

PR #2300, opened 2025-10-10
Docs/add instructions

ParsaBordbar:
Hi this pr adds instructions from https://rustlings.rust-lang.org/setup/
and a simple instructions for getting Back To Practice, that also address the issue [#2274].


mo8it:
I will add instructions to a separate README generated on initialization.

PR #2315, opened 2025-11-25
Update `13_error_handling/error6`: Remove redundant functions, use enum constructors instead

Lev200501:


senekor:
Using the enum variant as a constructor function is neat. But sometimes the error conversion function needs to do additional stuff. I think it's reasonable here to use an explicit, normal function. Even though it's not technically necessary.

mo8it (approved):

PR #2317, opened 2025-11-29
try_from_into.rs: Improve slice implementation

EugenDueck:
Improves readability. In terms of performance, it probably doesn't matter, because the multiple bounds checks in the original are likely to get optimized away.


senekor (approved):
I like this 👍

senekor:
I considered if maybe the pattern matching could confuse users, but this is one of the very last exercises. Users doing rustlings together with the book will have already read chapter 19 on pattern matching.

Issue #2233, opened 2025-04-09
Feature Request: File I/O Exercises in Rustlings


bluntbrain:
## Description
Add a new category (`24_file_io/`) focused on file operations, a crucial skill for Rust developers currently missing from Rustlings.

## Proposed Exercises

1. **Basic File Reading (file_io1.rs)**
   - Teaches using `fs::read_to_string` and handling file not found errors
   - Includes input file with text content to read
   - Tests verifying correct content parsing

2. **Buffered Reading & Writing (file_io2.rs)**
   - Demonstrates efficient reading with `BufReader` and writing with `BufWriter`
   - Processes a large input file line by line
   - Writes transformed content to an output file
   - Tests verifying correct transformation

3. **Path Manipulation & Metadata (file_io3.rs)**
   - Works with `Path` and `PathBuf` to handle file paths safely
   - Extracts and manipulates file metadata (creation time, size, permissions)
   - Tests verifying correct metadata extraction

## Implementation Structure
- Each exercise includes a Rust file and associated input/output files
- Files are embedded and available during testing
- All exercises follow Rustlings patterns with TODO comments and progressive difficulty

## Questions
1. Is handling extra files (inputs/outputs) compatible with current Rustlings structure?
2. Any specific file I/O patterns you'd prefer to prioritize?

I'm happy to draft sample implementations if this proposal is of interest.

cc @mo8it

bluntbrain:
Hi @mo8it ! 
I learned Rust a few months ago using Rustlings and it was incredibly helpful. I'm a big fan of project-based learning and would love to contribute to this repository. 

Let me know your thoughts on this proposal or if there are other features you'd like me to work on instead, I have a few hours daily to contribute. thanks!

senekor:
I'm not a fan of adding file IO exercises to Rustlings. The concepts for doing file IO are the same as in any other language. People can just check the `std::fs` documentation. I think it should be considered out-of-scope for Rustlings to teach the entire surface area of the standard library.

PR #2321, opened 2025-12-11
File IO Exercises added and tested

kivancgnlp:
Hi,  

I've added the exercises as described in the issue 2233

Please feel free to give feedbacks, thanks.

https://github.com/rust-lang/rustlings/issues/2233

### DIFF
--- a/dev/Cargo.toml
+++ b/dev/Cargo.toml
@@ -188,6 +188,12 @@ bin = [
   { name = "conversions4_sol", path = "../solutions/23_conversions/conversions4.rs" },
   { name = "conversions5", path = "../exercises/23_conversions/conversions5.rs" },
   { name = "conversions5_sol", path = "../solutions/23_conversions/conversions5.rs" },
+  { name = "file_io1", path = "../exercises/24_file_io/file_io1.rs" },
+  { name = "file_io1_sol", path = "../solutions/24_file_io/file_io1.rs" },
+  { name = "file_io2", path = "../exercises/24_file_io/file_io2.rs" },
+  { name = "file_io2_sol", path = "../solutions/24_file_io/file_io2.rs" },
+  { name = "file_io3", path = "../exercises/24_file_io/file_io3.rs" },
+  { name = "file_io3_sol", path = "../solutions/24_file_io/file_io3.rs" },
 ]
 
 [package]
--- a/exercises/24_file_io/README.md
+++ b/exercises/24_file_io/README.md
@@ -1,0 +1,30 @@
+# File IO
+
+Rust provides several file I/O functions in the standard library. Buffered
+reads and writes provide better performance by reducing the number of
+underlying system calls.
+
+## Learning goals
+
+- **file_io1** — Read an entire file into memory with `fs::read_to_string`
+  and handle the `Result` it returns instead of unwrapping blindly, so you
+  get comfortable with the basic "open, read, handle the error" pattern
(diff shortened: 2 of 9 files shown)


kivancgnlp:
Hi @mo8it and @senekor,

I wanted to check in on this PR's status. It's been open since December 2025, all CI checks are passing, and it's mergeable, but there's been no maintainer review or feedback since the initial notification discussion.

I also noticed @senekor's comment on the tracking issue (#2233) expressing that file I/O exercises might be out of scope for Rustlings, since the concepts overlap with general std::fs documentation rather than Rust-specific language concepts. That's a fair concern, and I'd rather get clarity now than leave this in limbo.

Could you let me know:

Whether the maintainers have reached a consensus on file I/O exercises being in/out of scope for Rustlings, and

If there's appetite for a scaled-down version of this PR, what that might look like — or if it makes more sense to close this out (along with the related duplicate, #2191).

Happy to adjust or close if that's the direction. Just want to avoid this sitting indefinitely. Thanks for your time.

senekor:
I don't have time to review right now, August should be better.

I still have that opinion about file IO exercises in general. We discussed that using file IO for the async exerices, which we do want, would be good. So I'm fine at least with file IO exercises that prepare for the async ones. If Mo thinks more file IO exercises are good on their own, I'm not strongly opposed to adding them either.

So, I personally will first work on the async exercises again before deciding what they need in terms of file IO.

It would be helpful for future reviews if you could describe the learning goal of each exercise. Then we can first think about if the learning goals themselves are good and well-ordered. In a second step, the individual exercises can be reviewed in terms of how well they actually teach the intended goal.

(Side note on LLM usage: There's no official policiy, but I think Mo is not a big fan just like myself. I would prefer you drop the Claude co-author out of your commits. The tools you use don't matter, only you are responsible for the code you submit. Maybe cleanup the commit history in general, while you're at it.)

kivancgnlp:
Hi @senekor,

Just a quick follow-up. I've cleaned up the branch history (down to 2 commits, no LLM co-author lines) and added learning-goal descriptions for each exercise to the README, per your feedback:

- **file_io1** : read a whole file with fs::read_to_string and handle the Result it returns
- **file_io2** : read/write efficiently line-by-line with BufReader/BufWriter
- **file_io3** : build a path with PathBuf and inspect it via .metadata()

I also expanded the hints in info.toml to point at the relevant concept/API rather than just restating the exercise title.

No rush given your August timeline, just wanted this ready and easy to review whenever you get to the async exercises and revisit file I/O scope. Thanks again for the detailed feedback.

kivancgnlp:
Hi @senekor,
Just checking in on this one during your August window. 
The branch is still mergeable, CI green, and the cleanup we discussed is in (README learning goals, 2-commit history, expanded hints). 
Happy to make any further tweaks if you want to trim scope to just the async-preparation exercises.

senekor:
The current state of the async exercise doesn't require file IO exercises to come beforehand, IMO. So, I'm not planning to go forward with this unless Mo sees independent value in IO exercises.

kivancgnlp:
Hi @mo8it, senekor mentioned the decision rests with whether you see independent value in file I/O exercises (outside of async preparation). If you have time to share a quick opinion, I'd appreciate it, and if it's not a fit for rustlings, I'm happy to close.

kivancgnlp:
Thanks both. Closing since there's no clear standalone case for these exercises in rustlings.

PR #2327, opened 2025-12-26
Fix wording in quiz1 to match test condition

### LINKED ISSUE #2326: Please fix the wording in quiz1
The text says "However, if Mary buys more than 40 apples".  But there is a test for specifically 40 apples.  Please change the wording to 40 or more.  It changes the if statement from "if apple_qty < 40" to "if apple_qty <= 40".

Took me a few minutes to realize why it was showing failed in the background check.

muskan-773:
This change updates the wording in quiz1 to match the test logic. The condition now correctly states "40 or more apples" instead of "more than 40".
Fixes #2326

### DIFF
--- a/exercises/quizzes/quiz1.rs
+++ b/exercises/quizzes/quiz1.rs
@@ -5,7 +5,7 @@
 //
 // Mary is buying apples. The price of an apple is calculated as follows:
 // - An apple costs 2 rustbucks.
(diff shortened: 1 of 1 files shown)


senekor:
The discussion on #2326 concluded that the exercise is fine the way it is. This PR should be closed.

PR #2333, opened 2026-01-17
improve iterators3 to demonstrate collect() type-driven behavior

leochen4891:
- No change to the existing solution
  - Use simpler numbers for easier mental math (9, 12, 18, 6 ÷ 3)
  - Add function parameters to allow testing different inputs
  - Add error case tests to show fail-fast vs process-all behavior
  - Enhance comments explaining Result<Vec<T>> vs Vec<Result<T>>

### DIFF
--- a/exercises/18_iterators/iterators3.rs
+++ b/exercises/18_iterators/iterators3.rs
@@ -15,17 +15,21 @@ fn divide(a: i64, b: i64) -> Result<i64, DivisionError> {
 }
 
 // TODO: Add the correct return type and complete the function body.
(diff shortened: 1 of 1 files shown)


senekor:
I don't think this adds any value.

- Users don't have to do the math, those who want to might enjoy the challenge.
- The error case test doesn't actually demonstrate fail-fast behavior. It only shows that the first error is returned, not that the subsequent values are never processed.
- The additional comments remove the last mystery for users to figure out on their own. The hint already contains all the information necessary for people who can't figure it out on their own.

PR #2336, opened 2026-01-23
Clarify TODO comments in options2 exercise

### LINKED ISSUE #1990: options2: unclear what to do and why
Hello Rustlings Team,

First of all thank you for this project. It was really fun solving it so far. I got really stuck and scratching my head at this one and I decided to ask you for an improvement.

```rust
    fn simple_option() {
        let target = "rustlings";
        let optional_target = Some(target);

        // TODO: Make this an if let statement whose value is "Some" type
        word = optional_target {
            assert_eq!(word, target);
        }
    }
```

Looks like this excercise is trying to show how to use `Option<T>` type in Rust. Currently, the exercise seems somewhat forced and does not effectively demonstrate the real-world use cases of the Option<T> type.

We start with a "rustlings" string literal and wrap it in optional_target. Then the comment asks us to `Make this an if let statement whose value is "Some" type` (which is a riddle by itself) providing us with some convoluted assertion. I am just learning Rust and have never seen this type before. If that task had some real-world scenario showing us why Option type is useful (for example safe parsing of data) that would really help.

jjstrnad:
Fixes #1990

## Summary
Clarified TODO comments in options2 exercise to indicate that the code needs to be completed rather than replaced.

## Problem
The current TODO comments say "Make this an if-let statement" and "Make this a while-let statement", which implies the existing code is already a valid statement that needs to be transformed into something else. This caused confusion for many learners who expected to replace valid code with different valid code.

In reality, the code is syntactically incomplete and needs keywords added (`if let` and `while let`) to become valid statements.

## Solution
Updated the TODO comments to explicitly state that learners need to complete the incomplete statements by adding the missing keywords and patterns.

### DIFF
--- a/exercises/12_options/options2.rs
+++ b/exercises/12_options/options2.rs
@@ -9,7 +9,8 @@ mod tests {
         let target = "rustlings";
         let optional_target = Some(target);
 
-        // TODO: Make this an if-let statement whose value is `Some`.
+        // TODO: Complete this if-let statement by adding the missing keywords.
+        // The pattern should match oon 'Some'.
         word = optional_target {
             assert_eq!(word, target);
         }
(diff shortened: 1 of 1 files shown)


senekor:
I don't think this is an improvement. The current comments seem unterstandable to me. Saying "by adding the missing keywords" is arguably worse, because users have to do more than just add keywords.

jjstrnad:
Fair enough, appreciate you taking a look!

Issue #1204, opened 2022-09-19
Only mention of anonymous structs is enums2 hint


Brian-K-Smith:
"Anonymous struct" isn't a term or concept introduced in the structs exercises or Ch. 5 of the Rust book.  It does appear in the hint for enums2. The way the Move struct is created in the enums2 exercise example I found after cheating on the internet seems like it probably is an anonymous struct.
`enum Message {
    Move{x: i32, y: i32,}, Echo(String), ChangeColor(i32, i32, i32), Quit,
}`
I don't know if it makes sense to add the concept in the structs exercises or to insert an enums exercise that covers this. From what I can see as a complete novice randomly searching on the internet it looks like the primary use case for anonymous structs is as part of an enum.

Brian-K-Smith:
Ch 6 of the Rust book provides more clues.

PR #2353, opened 2026-02-12
Mention struct-like variants in enums2 hint instead of anonymous structs

orzechow:
The current hint of the enums2 exercise uses an outdated term for struct-like enum variants, which is confusing to Rust newbies:

> You can create enumerations that have different variants with different types such as anonymous structs, structs, a single string, tuples, no data, etc.

The term `anonymous struct` seems outdated. It's neither mentioned in the Rust book on [structs](https://doc.rust-lang.org/stable/book/ch05-01-defining-structs.html) or [enums](https://doc.rust-lang.org/stable/book/ch06-01-defining-an-enum.html#listing-6-2), nor the references on [structs](https://doc.rust-lang.org/stable/reference/items/structs.html) or [enums](https://doc.rust-lang.org/stable/reference/items/enumerations.html#r-items.enum.constructor).
A web search rather points to [a Rust Forum thread](https://users.rust-lang.org/t/is-the-anonymous-struct-unnamed-struct-still-in-progress/134416) hinting
> Don’t assume that there is such a thing as “the anonymous struct feature” at all

The only place I found this term, is in the Rust docs on [enums](https://doc.rust-lang.org/std/keyword.enum.html), but not in the docs on [structs](https://doc.rust-lang.org/std/keyword.struct.html).
It seems like Rust used this term once, but not anymore. See also #1204 and the 2018 edition [Rust book on enums](https://doc.rust-lang.org/1.30.0/book/2018-edition/ch06-01-defining-an-enum.html).

Instead the [book on enums](https://doc.rust-lang.org/stable/book/ch06-01-defining-an-enum.html#listing-6-2) now calls it `named fields, like a struct`.
The [reference on enums](https://doc.rust-lang.org/stable/reference/items/enumerations.html#r-items.enum.constructor) uses the term `struct-like enum variant`.

Therefore, I'd suggest updating this hint to `struct-like enum variants`.

### DIFF
--- a/rustlings-macros/info.toml
+++ b/rustlings-macros/info.toml
@@ -440,7 +440,7 @@ dir = "08_enums"
 test = false
 hint = """
 You can create enumerations that have different variants with different types
-such as struct-like variants, regular structs, a single string, tuples, no data, etc."""
+such as struct-like, tuple-like and unit-only variants."""
 
 [[exercises]]
 name = "enums3"
--- a/rustlings-macros/info.toml
+++ b/rustlings-macros/info.toml
@@ -440,7 +440,7 @@ dir = "08_enums"
 test = false
 hint = """
 You can create enumerations that have different variants with different types
-such as anonymous structs, structs, a single string, tuples, no data, etc."""
+such as struct-like variants, regular structs, a single string, tuples, no data, etc."""
 
 [[exercises]]
 name = "enums3"


MatrixFrog (approved):

senekor:
I think the list of possible variants is still not correct. There are three possibilities, syntactically:
- struct-like: `enum E { Variant { field: i32 } }`
- tuple: `enum E { Variant(i32) }`
- unit: `enum E { Variant }`

Not sure about the names, but these are the three syntactic possibilities that should be reflected.

orzechow:
Thanks for the feedback!

I agree, let's cleanup the list as well.
The [reference](https://doc.rust-lang.org/stable/reference/items/enumerations.html) uses terms like
- `struct-like [enum] variant`
- `tuple-like variant`
- `unit-like variant`, `unit-only enum`

See f9f8a37 

@senekor  Shall I also add those code snippets, you suggested? Like this:

```
hint = """
You can create enumerations that have different variants with different types
such as struct-like `enum E { Variant { field: i32 } }`,
tuple-like `enum E { Variant(i32) }` and unit-only `enum E { Variant }` variants."""
```

senekor:
Good idea to check the reference for what terms it uses. If the reference uses both, I would lean toward using "unit-like" instead of "unit-only", for uniformity. But both are fine.

Reading over that hint again, the expression "variants with different types" also bugs me a little. The three possibilities "struct-like", "tuple-like" and "unit-like" aren't "types" in the context of a programming language... They are rather syntactic elements used to define a type.

I took a stab at rewriting the hint. What do you think, is it too much?

```
Enum variants can be defined using three different forms: "struct-like",
"tuple-like" and "unit-like". Struct-like variants contain named fields, whereas
tuple-like variants have unnamed fields. Unit-like variants don't have any
fields. Here's an example enum definition, which uses all three forms:

enum EnumUsingAllVariantForms {
    StructLike { named_field: bool },
    TupleLike(bool),
    UnitLike
}
```

@orzechow your commits show up as unverified btw., you might want to check your signing config :)

orzechow:
Agree, let's be precise.

`type` could be confusing – we're talking about variants of one struct.

Also, after re-reading the reference, an enum is called `unit-only` if it `only contains unit variants`. So, using `unit-like` for the variant is much better, yes.

I'd keep the code example and condense the description to:

```
Enum variants can be defined using three different forms: struct-, tuple- and 
unit-like. Here's an example enum definition, which uses all three forms:

enum EnumUsingAllVariantForms {
    StructLike { named_field: bool },
    TupleLike(bool),
    UnitLike
}
```

@senekor What do you think?

PS: Thanks for the signing hint – that shouldn't have happened…

senekor:
I like this version a lot. @mo8it what do you think?

mo8it (approved):

PR #2368, opened 2026-03-16
docs: clarify stale comment in cow1 owned_mutation test

### LINKED ISSUE #2303: [19_smart_pointers/cow1.rs] Stale comment in  `owned_mutation`  test is confusing
Hi,

I found a confusing comment in the `cow1.rs` exercise, inside the `#[cfg(test)]` module.

**File:** `exercises/19_smart_pointers/cow1.rs`
**Test:** `owned_mutation`

The test contains this comment:

```rust
    // Of course this is also the case if a mutation does occur (not all
    // numbers are absolute). In this case, the call to `to_mut()` in the
    // `abs_all` function returns a reference to the same data as before.
```

The last sentence, "...the `abs_all` function returns a reference...", is incorrect. The `abs_all` function provided in the exercise has this signature and returns unit `()`:

```rust
fn abs_all(input: &mut Cow<[i32]>) {
    // ...
}
```

This comment is very confusing for learners, as it describes behavior that doesn't match the code.

It looks like this comment is a "stale" leftover from a previous version of the exercise where `abs_all` *did* return a reference, like in this screenshot:

Image

To avoid confusion, this stale comment should probably be updated to remove the incorrect part about the return value, or be re-written entirely.

Thanks\!

DeoJin:
## Summary
- update the stale comment in exercises/19_smart_pointers/cow1.rs
- clarify that 	o_mut() returns a mutable reference to already-owned data here
- remove the misleading implication that bs_all itself returns a reference

Closes #2303

## Validation
- unable to run cargo test owned_mutation --bin rustlings in this environment because Rust tooling (cargo, 
ustc) is not installed
- verified the change is comment-only and scoped to the referenced exercise text

### DIFF
--- a/exercises/19_smart_pointers/cow1.rs
+++ b/exercises/19_smart_pointers/cow1.rs
@@ -57,9 +57,10 @@ mod tests {
 
     #[test]
     fn owned_mutation() {
-        // Of course this is also the case if a mutation does occur (not all
-        // numbers are absolute). In this case, the call to `to_mut()` in the
-        // `abs_all` function returns a reference to the same data as before.
+        // This is also the case if a mutation does occur (not all numbers are
+        // absolute). Since `input` is already owned here, the call to
+        // `to_mut()` in `abs_all` returns a mutable reference to the existing
+        // owned data instead of cloning it.
         let vec = vec![-1, 0, 1];
         let mut input = Cow::from(vec);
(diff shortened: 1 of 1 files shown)


senekor:
I don't see the improvement here. The two comments seem equally clear to me.

If an exercise file is changed, the solution file should be kept in sync.

PR #2369, opened 2026-03-18
fix: Better test for exercises/10_modules/modules1.rs

### LINKED ISSUE #2339: Better test for exercises/10_modules/modules1.rs
In this exercise, if we comment out the function call in the `main` function, the test passes, 

```Rust
fn main() {
    // sausage_factory::make_sausage();
}
```

I'm suggesting to add a better test for this exercise. So that on commenting the testcase we don't pass the test case.

JiwaniZakir:
Fixes #2339

The `modules1` exercise could be passed by commenting out the call in `main`, since no test enforced that `make_sausage()` must be public. Adds a `#[cfg(test)]` block to `exercises/10_modules/modules1.rs` that calls `sausage_factory::make_sausage()` directly, and sets `test = true` in `rustlings-macros/info.toml` so rustlings actually runs `cargo test` against the exercise.

### DIFF
--- a/exercises/10_modules/modules1.rs
+++ b/exercises/10_modules/modules1.rs
@@ -14,3 +14,13 @@ mod sausage_factory {
 fn main() {
     sausage_factory::make_sausage();
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_make_sausage() {
+        sausage_factory::make_sausage();
+    }
+}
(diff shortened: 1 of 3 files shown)


JiwaniZakir:
Closing this PR as it hasn't received maintainer review. Happy to resubmit if there's interest.

senekor:
I'm not sure what this would improve. The new tests can also just be commented out. We don't need to prevent cheating, since we assume users do these exercises voluntarily.

We could possibly add a comment above the main function, telling users not to change it, in case it's unclear. But that seems pretty obvious to me from the context.

PR #2377, opened 2026-04-01
Add explicit ownership check to iterators1.rs

### LINKED ISSUE #2261: Using to_iter instead of iter leads to unhelpful errors in iterators1
The hint for iterators1 leads to [the trait.Iterator documentation](https://doc.rust-lang.org/std/iter/trait.Iterator.html), where many of the examples use the to_iter method for obtaining the iterator. The compiler errors when using this method are unfriendly, and it is not clear that the iter method exists.

The simplest solution in my mind is to add a use of my_fav_fruits after the assert statements, to compel the compiler to throw an error about the array being consumed if using to_iter instead of iter.

penguinawesome1:
Closes #2261

This PR adds an explicit check to the array after iterator creation to ensure it was not consumed.
I noticed that while that assert technically passes using `into_iter()` because `&str` implements `Copy`, adding this `len()` check reinforces the intent.


senekor:
As you say, the assert doesn't actually enforce what it's supposed to. So, I think there's potential here to create even more confusion. If the only point is to express intent, improving the comments might be a better approach.

PR #2378, opened 2026-04-07
Remove unnecessary tmp.txt file

k7a-tomohiro:
`tmp.txt` was maybe pushed by mistake in https://github.com/rust-lang/rustlings/commit/4d97c31c0f748acda001e6f0f9c63261ec1d7afe.

So remove it.


mo8it:
You are right. Thanks

PR #2380, opened 2026-04-17
Add a note about offline stdlib docs

### LINKED ISSUE #2275: Offline Mode
Somewhere during initialization or after, give an to download all the docs referenced and have the links point to the offline references instead of going to the website.

MatrixFrog:
Fixes https://github.com/rust-lang/rustlings/issues/2275

### DIFF
--- a/website/content/setup/index.md
+++ b/website/content/setup/index.md
@@ -73,6 +73,10 @@ While working with Rustlings, please use a modern terminal for the best user exp
 The default terminal on Linux and Mac should be sufficient.
 On Windows, we recommend the [Windows Terminal](https://aka.ms/terminal).
 
+### Offline documentation (optional)
+
+If you are going to be working on Rustlings while offline, you may want to run `rustup doc --std` to make the standard library documentation available on your machine.
+
 ## Usage
 
 After being done with the setup, visit the [**usage**](@/usage/index.md) page for some info about using Rustlings 🚀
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Feel free to edit if there's a better command to run to download the docs

website/content/setup/index.md
```diff
@@ -73,6 +73,10 @@ While working with Rustlings, please use a modern terminal for the best user exp
 The default terminal on Linux and Mac should be sufficient.
 On Windows, we recommend the [Windows Terminal](https://aka.ms/terminal).
 
+### Offline documentation (optional)
+
+If you are going to be working on Rustlings while offline, you may want to run `rustup doc --std` to make the standard library documentation available on your machine.
```
senekor: This sounds a little like it's a necessary preparation step that needs to be run _before_ going offline. As if `rustup doc --std` downloads the documentation. How about something like:

> Whenever you're working on Rustlings offline, you can access a local copy of the standard library documentation by running `rustup doc --std`.

MatrixFrog: Yes I was under the impression you had to run it at least once to make sure the docs get downloaded. It looks like I was wrong and docs are installed by default along with the compiler and everything else.

senekor (approved):
I guess it's not technically a setup step, but it seems like a good time to provide that tip. Moving it to the "usage" page could also be appropriate. LGTM.

PR #2381, opened 2026-04-17
Emphasize hotkeys in footer with color

senekor:
I find this nicer to read than the angle bracket emphasis:

image

---

This is how it looks with a filter activated:

image

### DIFF
--- a/src/list/state.rs
+++ b/src/list/state.rs
@@ -233,15 +233,38 @@ impl<'a> ListState<'a> {
             )?;
             next_ln(stdout)?;
 
+            let hotkey = |writer: &mut MaxLenWriter, hotkey| -> io::Result<()> {
+                writer
+                    .stdout
+                    .queue(SetForegroundColor(Color::Yellow))?
+                    .queue(SetAttribute(Attribute::Bold))?;
+                writer.write_ascii(hotkey)?;
+                writer.stdout.queue(ResetColor)?;
+                Ok(())
+            };
+
             let mut writer = MaxLenWriter::new(stdout, self.term_width as usize);
             if self.message.is_empty() {
                 // Help footer message
                 if self.scroll_state.selected().is_some() {
-                    writer.write_str("↓/j ↑/k home/g end/G | <c>ontinue at | <r>eset exercise")?;
+                    writer.write_str("↓/")?;
+                    hotkey(&mut writer, b"j")?;
+                    writer.write_str(" ↑/")?;
+                    hotkey(&mut writer, b"k")?;
+                    writer.write_ascii(b" home/")?;
+                    hotkey(&mut writer, b"g")?;
+                    writer.write_ascii(b" end/")?;
+                    hotkey(&mut writer, b"G")?;
+                    writer.write_ascii(b" | ")?;
+                    hotkey(&mut writer, b"c")?;
+                    writer.write_ascii(b"ontinue at | ")?;
+                    hotkey(&mut writer, b"r")?;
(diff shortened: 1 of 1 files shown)


mo8it:
I love it! Since the colors are essential here, did you check whether they are fine for color-blindness?

mo8it:
Maybe also make it bold?

senekor:
> did you check whether they are fine for color-blindness?

Not really. It's the same color is the "PENDING" marker in the list of exercises. Also, the user controls the actual colors that are being rendered via their terminal theme / configuration. So, if somebody can't see yellow, presumably they already configured their terminal to render something that's better visible to them.

> Maybe also make it bold?

Good idea! Done.

Btw. do you prefer clean history or never force-pushing for PRs? My default is to amend and force-push, but I can do what's more pleasant for you to review.

senekor:
updated screenshots with boldness:

image

---

image

src/list/state.rs
```diff
@@ -233,43 +233,80 @@ impl<'a> ListState<'a> {
             )?;
             next_ln(stdout)?;
 
+            let hotkey = |writer: &mut MaxLenWriter, hotkey| -> io::Result<()> {
```
mo8it: This should be a normal function. Nothing is being used from the environment to justify a closure.

src/list/state.rs
```diff
@@ -233,43 +233,80 @@ impl<'a> ListState<'a> {
...
                 }
 
                 match self.filter {
                     Filter::Done => {
+                        writer.stdout.queue(SetAttribute(Attribute::Underlined))?;
+                        hotkey(&mut writer, b"d")?;
                         writer
                             .stdout
                             .queue(SetForegroundColor(Color::Magenta))?
                             .queue(SetAttribute(Attribute::Underlined))?;
```
mo8it, line 280: This can be removed.

src/list/state.rs
```diff
@@ -233,43 +233,80 @@ impl<'a> ListState<'a> {
...
                     Filter::Pending => {
-                        writer.write_ascii(b"<d>one/")?;
+                        hotkey(&mut writer, b"d")?;
+                        writer.write_ascii(b"one/")?;
+                        writer.stdout.queue(SetAttribute(Attribute::Underlined))?;
+                        hotkey(&mut writer, b"p")?;
                         writer
                             .stdout
                             .queue(SetForegroundColor(Color::Magenta))?
                             .queue(SetAttribute(Attribute::Underlined))?;
```
mo8it, line 295: This too

mo8it (changes requested):

mo8it:
> Btw. do you prefer clean history or never force-pushing for PRs? My default is to amend and force-push, but I can do what's more pleasant for you to review.

Force-pushing in PRs is fine. But only when the whole direction changes. The goal is not to end up with one commit. Especially changes after a review shouldn't be amended.

src/list/state.rs
senekor: I use `writer.stdout.queue(ResetColor)?;` in the `hotkey` function, which also resets the underline attribute (a little confusing, I think). So just removing that will lead to only the first letter being underlined. I can also reset color and boldness individually in the `hotkey` function, if that's better.

mo8it: Ah, you are right

mo8it:
Can you please pull out the function `hotkey`? Defining a function in a function should only be done when the function is rather small and it has to happen at the top.
I will merge after that nitpick :)

PR #2382, opened 2026-04-18
Add exercise async1

senekor:
The goal here was to get the first bit of "muscle memory" for using the async and await keywords. The little story should make it more intuitive for users why asynchronous programming is needed in the first place.

This exercise can be moved to the location corresponding to the book in a later PR, to keep the diff of this one clean.


dev/Cargo.toml
```diff
@@ -196,6 +198,9 @@ edition = "2024"
 # Don't publish the exercises on crates.io!
 publish = false
 
+[dependencies]
+tokio = { version = "1.52.1", features = ["rt"] }
```
mo8it: `version = "1" ` should be enough

exercises/24_async/async1.rs
```diff
@@ -0,0 +1,55 @@
...
+// finish the chores earlier and have more time left to play soccer.
+//
+// Let's simulate this using asynchronous programming. Each boy is represented
+// as an asynchronous task, which can be executed concurrently (they can be
+// working at the same time).
+
+use std::sync::atomic::{AtomicU8, Ordering};
+
+// Used by "mom" to check that all chores are done before Tim plays soccer :-)
+static CHORES_DONE: AtomicU8 = AtomicU8::new(0);
```
mo8it: Atmoics are not introduced in Rustlings, right? So some explanation is needed or a comment to tell users to not worry about it.

exercises/24_async/async1.rs
```diff
@@ -0,0 +1,55 @@
...
+
+fn main() {
+    // Async tasks need to be executed by a "runtime", which is not provided by
+    // Rust's standard library. We use the popular "tokio" runtime here.
```
mo8it: Popular doesn't necessarily mean that it is mature and well established. Maybe "mainstream"?

```suggestion
    // Rust's standard library. Here, we use the mainstream runtime `tokio`.
```

exercises/24_async/async1.rs
```diff
@@ -0,0 +1,55 @@
+// Tim has to complete a few chores today, before he's allowed to play soccer
+// with his friends. His friends decide to help him. Working together, they
+// finish the chores earlier and have more time left to play soccer.
+//
+// Let's simulate this using asynchronous programming. Each boy is represented
+// as an asynchronous task, which can be executed concurrently (they can be
```
mo8it: Here, you say that they can be working at the same time. But in `main`, you create a single threaded runtime.

exercises/24_async/async1.rs
```diff
@@ -0,0 +1,55 @@
...
+    // Block the runtime on a task that waits for all boys to finish the chores.
+    // TODO: "await" all three tasks to fix the compiler errors.
+    rt.block_on(async {
+        task_tim;
+        task_carl;
+        task_nick;
+    });
+
+    assert_eq!(
+        CHORES_DONE.load(Ordering::SeqCst),
```
mo8it: `Relaxed` everywhere is enough

mo8it (changes requested):
I like
- The requirement of adding `async` and `.await`
- Spawning tasks, awaiting them and then checking that they are done
- The strory about splitting tasks among different workers

I don't like
- The usage of atomics
- The work done in each task. Printing locks under the hood so the tasks will run mostly sequentially even if the runtime was multi-threaded
- The story details about the boys and soccer. Some might find it a bit childish for tasks meant mainly for adults

What about letting the tasks do some calculation and return the result? Then all three results could be checked. This way, we don't need atomics or printing.

mo8it:
What about using `tokio::fs` to open three files asynchronously, reading their content by parsing each line as a number and summing them up?

Something like that does actual work and is a valid usage for async.

mo8it:
The story could be something like teachers want to calculate the mean grade for three different classes. Instead of only one teacher doing all the work or doing it sequentially, they can do it async.

exercises/24_async/async1.rs
senekor: "concurrently" is the more precise word. I think it's good for teaching to use a single-threaded runtime, to show that async can make a difference even without multi-threading. Though, as you mentioned, just printing to the console might not be best for that.

senekor:
Agree with most of these points. I redid the exercise to hopefully address them. I'm not sure yet about using a multi-threaded runtime and using file IO. Yes, it would be more realistic, but I'm concerned it might be more complicated and distract from the intended lesson here - building a little async/await muscle memory.

exercises/24_async/async1.rs
```diff
@@ -1,55 +1,42 @@
...
-    println!("Cleaning my room...");
-    CHORES_DONE.fetch_add(1, Ordering::SeqCst);
-}
-
-fn carl() {
-    println!("Washing the dishes...");
-    CHORES_DONE.fetch_add(1, Ordering::SeqCst);
+    assert_eq!(mean_score_a, 84);
+    assert_eq!(mean_score_b, 89);
+    assert_eq!(mean_score_c, 76);
```
mo8it: The assertions can be moved to the body of `block_on` to avoid the indirection by the array.

exercises/24_async/async1.rs
```diff
@@ -1,55 +1,42 @@
-// Tim has to complete a few chores today, before he's allowed to play soccer
-// with his friends. His friends decide to help him. Working together, they
-// finish the chores earlier and have more time left to play soccer.
+// Alice is an elementary school teacher who needs to calculate the mean test
+// score for three classes she teaches. Instead of calculating them one after
+// the other, she decides to ask her friends Bob and Catherine for help. Working
+// together, they can finish the job much faster.
```
mo8it, line 4: Using a single thread without actual async work means that they won't finish much faster, in fact, they will be slightly slower.

exercises/24_async/async1.rs
```diff
@@ -1,55 +1,42 @@
...
-// as an asynchronous task, which can be executed concurrently (they can be
-// working at the same time).
-
-use std::sync::atomic::{AtomicU8, Ordering};
-
-// Used by "mom" to check that all chores are done before Tim plays soccer :-)
-static CHORES_DONE: AtomicU8 = AtomicU8::new(0);
+// Let's simulate this using asynchronous programming. Each person is
+// represented as an asynchronous task, which can be executed concurrently (i.e.
+// they can be doing the calculations at the same time).
```
mo8it: "at the same time" is misleading

mo8it (changes requested):
I like the overall direction, but I think that trying to keep the exercise simple results in misleading the users. By avoiding async work, they won't get the difference to multi-threading and will be confused because the solution is slower on a single thread.

mo8it:
Maybe somthing like this?

```rust
fn main() {
    // Async tasks need to be executed by a "runtime", which is not provided by
    // Rust's standard library. Here, we use the mainstream runtime `tokio`.
    let rt = tokio::runtime::Builder::new_current_thread()
        .build()
        .unwrap();

    let alice = rt.spawn(calculate_mean_score("a.txt"));
    let bob = rt.spawn(calculate_mean_score("b.txt"));
    let catherine = rt.spawn(calculate_mean_score("c.txt"));

    // Block the runtime on a task that awaits all three calculations.
    rt.block_on(async {
        // TODO: "await" all three tasks.
        assert_eq!(alice, 84);
        assert_eq!(bob, 89);
        assert_eq!(catherine, 76);
    });
}

// TODO: Fix the compiler errors by making the spawned function async.
fn calculate_mean_score(scores_file: &str) -> usize {
    // Read the file asynchronously
    let file = tokio::fs::read_to_string(scores_file).await.unwrap();

    // Initialize the sum and the number of scores
    let mut sum = 0;
    let mut n = 0;
    for line in file.lines() {
        // Parse every line as a score
        let score = line.parse::<usize>().unwrap();
        sum += score;
        n += 1;
    }

    sum / n
}
```

It is not much more complicated. Remember that this exercise is one towards the end, so the users aren't complete beginners at this point.

If we take this direction, we should add the IO tasks before.

senekor:
One thing I'm not sure about with the file IO approach is how to show users that it's actually running concurrently? Presumably the work is still small enough that users won't actually notice the performance difference between sync and async versions. Should we run some benchmarks and display the results to users or something?

mo8it:
They don't need to feel the difference. It is enough to know that this approach is theoretically more efficient if the files were bigger.

senekor:
FYI I'm pretty busy now until the end of July. I made a reminder to get back to this in August :)

dev/Cargo.toml
```diff
@@ -196,6 +198,9 @@ edition = "2024"
 # Don't publish the exercises on crates.io!
 publish = false
 
+[dependencies]
+tokio = { version = "1", features = ["fs", "macros", "rt", "rt-multi-thread"] }
```
mo8it:
```suggestion
tokio = { version = "1", features = ["fs", "macros", "rt-multi-thread"] }
```

`rt-multi-thread` activates `rt`

exercises/24_async/async1.rs
```diff
@@ -0,0 +1,41 @@
...
+//
+// Let's simulate this using asynchronous programming. Each person is
+// represented as an asynchronous task, which can be executed concurrently.
+
+// Async tasks need to be executed by a "runtime", which is not provided by
+// Rust's standard library. Here, we use the mainstream runtime `tokio`.
+// The macro `tokio::main` wraps the entire main function in a runtime.
+#[tokio::main]
+async fn main() {
+    // TODO: Fix the compiler errors by making the spawned function async.
```
mo8it: This TODO is redundant. It is duplicated above the function signature. I would prefer to keep it there since the actual required change is there.

rustlings-macros/info.toml
```diff
@@ -1211,3 +1211,26 @@ name = "conversions5"
...
+hint = """
+Asynchronous runtimes like tokio can only spawn tasks that are defined as async
+functions, not regular ones. Add the "async" keyword before the "fn" keyword of
+the functions "tim", "carl" and "nick".
+
+An async task can wait for another one to complete by "awaiting" it. Add
+".await" after the three "task_name" variables in the "block_on" call."""
+
+[[input_files]]
+name = "scores_class_a.txt"
```
mo8it: Why not just `input_files = ["…", "…", …]`? Do you expect more fields than `name` for input files?

mo8it (changes requested):

mo8it:
I like the async1 exercise!

This PR introduces a feature planned for the next version: https://github.com/rust-lang/rustlings/discussions/2170
We should think about the design before committing to one. Having everything in `input_files` can get chaotic pretty quickly. As a user, I would expect related files to be in the same directory where `main.rs` of the exercise is.

Another issue: If the user edits the input file, they won't be able to complete the exercise. Currently, we offer no way to reset input files. A better approach would be to check the hash of the input files of the current exercise before running it. If they were altered, we should reset them automatically from the embedded version.

rustlings-macros/info.toml
senekor: Agreed, just the name should be enough. That particular syntax will attach the input files to a specific exercise. (Whereas now, I had input_files as a global list.) That's probably a good thing, right? That automatically ensures the input files are in the same directory as the exercise file itself.

senekor:
I haven't yet added a check if the user changed an input file, I'll get back to that later.

Another thought: The paths are a little long right now. What do you think about executing exercises with their exercise directory as the CWD? That would shorten the path for reading input files. Also, it might be easier for text editors to recognize the input files as relative paths to the currently openend file.

src/dev/check.rs
```diff
@@ -133,19 +133,25 @@ fn check_info_file_exercises(info_file: &InfoFile) -> Result<HashSet<PathBuf>> {
 
         file_buf.clear();
 
-        paths.insert(PathBuf::from(path));
+        let path = PathBuf::from(path);
+
+        for input_file in &exercise_info.input_files {
+            paths.insert(path.parent().unwrap().join(input_file));
```
mo8it: Getting the parent should be done outside the loop

mo8it:
Looks good so far!
Definitely set CWD for running the binary. It should be enough to add it to the command here: https://github.com/rust-lang/rustlings/blob/9ed849c7aaea8cfc2e096de81d574f99eacc7e80/src/cmd.rs#L144

After thinking about it again, maybe checking if the file is not changed is less efficient than just dumping it every time before executing the binary (not before check and build). This way, we don't have to open the file, read it and calculate its hash.

mo8it:
Force pushing forces me to rereview everything 😅 It's easier to push new commits after a review. Don't worry too much about the history. I will review later.

senekor:
Sorry, there was a trivial merge conflict. The existing commits are unchanged. I guess you would've preferred I merge main into my branch?

tshepang:
> Force pushing forces me to rereview everything 😅 It's easier to push new commits after a review.

triagebot can help: https://forge.rust-lang.org/triagebot/range-diff.html

Chuhan-Mateo:
I wish this could be merge soon! 🚀

src/exercise.rs
```diff
@@ -123,6 +140,14 @@ pub trait RunnableExercise {
             output.clear();
         }
 
+        let cwd_buf;
```
mo8it, line 143: I have never seen this before. A nice way to extend the lifetime :D

mo8it (approved):

mo8it:
Thanks @senekor, the PR took a while (sorry for the delayed reviews), but it is really good now :D

For an incremental review in future PRs: If a PR already has a review, I think it would be the easiest to not fix merge conflicts until the PR is approved. If the fixes are trivial, just rebase and merge directly after the approval. In doubt, better just merge main.

senekor:
Understood, I'll try to remember in the future. It's possible I forget... I also contribute to projects that have the [opposite philosophy](https://github.com/jj-vcs/jj?tab=contributing-ov-file#code-reviews) 😅 What do you think about enabling [range-diff from triagebot](https://forge.rust-lang.org/triagebot/range-diff.html), as suggested above? If only as a safety net, in case I do forget.

As it happens, I'm actually working on [my own solution](https://blog.buenzli.dev/announcing-development-on-flirt/) to the "must review everything from scratch after rebase" problem, but that one's not production ready yet.

mo8it:
Hot take: A clean commit history is overrated as long as you keep your commits as small as possible and give them a good description.

Flirt sounds promising! I like the idea :D

PR #2383, opened 2026-04-18
Add exercise async2

senekor:
The lesson of this exercise is that blocking primitives from the standard library are often unsuitable for asynchronous programming. The runtime usually provides async-compatible replacements.


senekor:
It's been a while, I need to rethink where we might want to go with the second async exercise.

mo8it:
Request a review when ready :D

PR #2384, opened 2026-04-18
Allow selecting next exercise with enter key

senekor:


mo8it (changes requested):
I agree that `Enter` is more intuitive. But then I would remove `c`. It should be added to the footer.

mo8it:
I remember now why I didn't choose Enter: I didn't know how to show it in the footer while staying consistent with the other keys. Maybe just `Enter: continue at`?

senekor:
I think `c` is actually fine too. I'd guess there are two types of people: Some check the footer to find out what to do. Those will use `c`, as indicated. Others will hit enter blindly, assuming it'll work, because it's intuitive (that's me). Both are fine, and I don't think it would bother anyone that only one option shows up in the footer.

mo8it:
I don't like it when an option is not defined. If I hit enter, I want to be able to look up what exactly it does.
Could you then just add `Enter/` to the footer?

senekor:
Sure. There's also a unicode symbol, just like for the up and down arrows:
image

Although I'm not sure if all terminal fonts render this well. What do you prefer, symbol or just "Enter" ?

This will cause an (easy to fix) merge conflict with #2381, just to keep in mind.

mo8it (approved):

PR #2386, opened 2026-05-03
Use infallible conversion to teach From trait

### LINKED ISSUE #2225: Incorrect answer passed in from_into.rs

senekor:
The From trait should only be implemented if the conversion is infallible. The previous exercise design used a fallible conversion with a fallback to a default value, with is bad practice.

closes #2225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - `structs3`: Rewrote the exercise to make users type method syntax themselves.
 - Rename the exercises for smart pointers and conversions so they're sorted alphabetically. [@foxfromworld](https://github.com/foxfromworld)
 - `vecs1`: Remove array literal. Some learners assumed their task is to convert it to a vector.
+- `conversions2`: Redesign the context such that infallible conversion makes sense.
 
 ## 6.5.0 (2025-08-21)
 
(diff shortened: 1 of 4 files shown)


mo8it (commented):
I like the direction of using `From` for infallible conversions only.

The story contains many words which a non-native speaker might not know. It overcomplicates the exercise which only requires multiplication by 100. I would suggest simplifying the story and making the exercise a bit harder.

senekor:
Had to think a little to find a decent example. I came up with Celsius <-> Fahrenheit conversion. No story needed to explain and the task is slighly more difficult than multiplying by 100.

mo8it (changes requested):
I love the exerice now!

Can you please add the formula for the backwards conversion to the hint and briefly explain its derivation? I know that it makes the exercise more interesting and it is simple math, but our task is not to teach math. So hiding it in the hint would be a good compromise.

You can merge afterwards :D

PR #2397, opened 2026-05-14
Redesign vec1 to avoid confusing array literal

### LINKED ISSUE #2194: exercises/05_vecs/vecs1.rs: TODO clarification
## exercises/05_vecs/vecs1.rs: 
```rust
let a = [10, 20, 30, 40];
// TODO: Create a vector called `v` which contains the exact same elements as in the array `a`
// Use the vector macro
```
vec! macro can only be used like this: `vec![10, 20, 30, 40]`.
Is it intended to be like this? I guess, it would make more sense to use `Vec::from(a)`

senekor:
Learners were sometimes confused by the literal array. Their assumption was that they are supposed to convert the array to a vector. Duplicating the literal elements was not an intuitive solution.

This redesign avoids putting an identical array literal near the place where learners are supposed to use the vec! macro.

### DIFF
--- a/exercises/05_vecs/vecs1.rs
+++ b/exercises/05_vecs/vecs1.rs
@@ -1,11 +1,6 @@
-fn array_and_vec() -> ([i32; 4], Vec<i32>) {
-    let a = [10, 20, 30, 40]; // Array
-
-    // TODO: Create a vector called `v` which contains the exact same elements as in the array `a`.
-    // Use the vector macro.
-    // let v = ???;
-
-    (a, v)
+fn elems_to_vec(a: i32, b: i32, c: i32) -> Vec<i32> {
+    // TODO: Return a vector containing the elements a, b and c.
+    // Use the "vec!" macro.
 }
 
 fn main() {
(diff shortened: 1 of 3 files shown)


exercises/05_vecs/vecs1.rs
```diff
@@ -1,11 +1,6 @@
-fn array_and_vec() -> ([i32; 4], Vec<i32>) {
-    let a = [10, 20, 30, 40]; // Array
-
-    // TODO: Create a vector called `v` which contains the exact same elements as in the array `a`.
-    // Use the vector macro.
-    // let v = ???;
-
-    (a, v)
+fn elems_to_vec(a: i32, b: i32, c: i32) -> Vec<i32> {
+    // TODO: Return a vector containing the elements a, b and c.
```
mo8it:
```suggestion
    // TODO: Return a vector containing the elements a, b and c (in this order).
```

mo8it (changes requested):

mo8it:
I can confirm that some users tried to covert it. It is much better now. You can merge after clarifying the required order.

Issue #1346, opened 2023-01-23
Instructions not clear for: `modules/modules2.rs`


archcorsair:
```rust
// TODO: Fix these use statements
use self::fruits::PEAR as ???
use self::veggies::CUCUMBER as ???

// ... snip

fn main() {
    println!(
        "favorite snacks: {} and {}",
        delicious_snacks::fruit,
        delicious_snacks::veggie
    );
}
```
At first glance it seems as though the solution would simply be to replace the `???` with respective `fruit` and `veggie`;
However after much struggling and leaning on the rust discord community for assistance it was pointed out that the `use` statement itself should also be changed to `pub use`. This was not clear in the original problem, as the concept of making an import public is not one I'm accustomed to coming from other languages. The compiler doesn't event suggest this change, making it even more difficult to come to the correct answer without external assistance.

diannasoreil:
Yeah, good point, I'd imagine we can probably rewrite this exercise as a whole anyways, it's kind of clunky.

justin-thurman:
Including a link to the Rust Book documentation on [re-exporting names](https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#re-exporting-names-with-pub-use) in the hint might make this one clear enough. I also fumbled with this exercise a bit, but I knew I had read something in the Rust Book about this. I just couldn't remember where it was.

dlloyd1:
In addition to the above, could this exercise benefit from extending the TODO comment directive to include a subtle nod to the fact that each use statement requires two distinct changes or perhaps that fixing the compiler error is a two-stage process here?

Step one: identify and correct the alias in the use statement
Step two: correct the access level of the constant import

The use of ??? here was helpful but made me zone in on it a bit too much, and I think that prevented me understanding what was needed properly.

senekor:
> Including a link to the Rust Book documentation on [re-exporting names](https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#re-exporting-names-with-pub-use) in the hint might make this one clear enough.

This was done in #1490. I don't think more help in the inline comment is benefitial. A redesign might be good though, so I'll keep this issue open for now.

PR #2398, opened 2026-05-14
Clarify instructions for modules2.rs (issue #1346)

foxfromworld:
This PR updates the TODO comment in `modules2.rs` to clarify that the `use` items are private and require re-exporting. This addresses the confusion reported in issue #1346.

### DIFF
--- a/exercises/10_modules/modules2.rs
+++ b/exercises/10_modules/modules2.rs
@@ -3,6 +3,7 @@
 
 mod delicious_snacks {
     // TODO: Add the following two `use` statements after fixing them.
+    // Hint: private imports require re-exporting.
     // use self::fruits::PEAR as ???;
     // use self::veggies::CUCUMBER as ???;
 


senekor:
I don't think this helps. The `modules1` exercise already covers public / private, so learners should be primed about that when seeing the "blabla is private" compiler error. The actual exercise hint also points to https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#re-exporting-names-with-pub-use, which is a pretty good explanation.

foxfromworld:
CI `cargo test --workspace` failed, but it passes locally.
The failure seems unrelated to this comment change.

foxfromworld:
> I don't think this helps. The `modules1` exercise already covers public / private, so learners should be primed about that when seeing the "blabla is private" compiler error. The actual exercise hint also points to https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#re-exporting-names-with-pub-use, which is a pretty good explanation.

Understood, I see your point — a redesign would make more sense.

PR #2400, opened 2026-05-17
Add additional comment to explain `03_if/if3` solution

### LINKED ISSUE #2211: Add additional comment to explain `03_if/if3` solution
I think `03_if/if3.rs` is a bit ambiguous. The code to be modified looks like this:

```rust

// TODO: Fix the compiler error in the statement below.
    let identifier = if animal == "crab" {
        1
    } else if animal == "gopher" {
        2.0
    } else if animal == "snake" {
        3
    } else {
        "Unknown"
    };

```

which is intuitively indicating to change "Unknown" to an unused identifier, however it still won't work because the compiler will continue prompting for changes until `2.0` is replaced with `2`.

Personally it took me a while to figure it out, and that's not the problem, but when I checked the official solution:

```rust

let identifier = if animal == "crab" {
        1
    } else if animal == "gopher" {
        2
    } else if animal == "snake" {
        3
    } else {
        // Any unused identifier.
        4
    };

```

the only change mentioned is from `Unknown` to `4`, but there's no explanation that `2.0` must also be changed to `2`. I think adding a comment to clarify the necessary type consistency would be very helpful.

wyf027:
Adds hints that all `if` branches must share one type and that unknown animals need an integer fallback. Closes #2211

### DIFF
--- a/exercises/03_if/if3.rs
+++ b/exercises/03_if/if3.rs
@@ -1,5 +1,10 @@
 fn animal_habitat(animal: &str) -> &str {
     // TODO: Fix the compiler error in the statement below.
+    //
+    // Hint: every `if` / `else if` / `else` branch must produce the same type.
+    // A floating-point literal, a string, and integers cannot be mixed in one chain.
+    // For unknown animals, use another integer identifier (for example `4`) instead of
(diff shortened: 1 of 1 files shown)


senekor:
slop

PR #2401, opened 2026-05-20
fix: remove ambiguous strings4 into example

### LINKED ISSUE #2190: strings4 can pass checks even with a "useless_conversion" warning.
On [line 24 of exercises/09_strings/strings4.rs](https://github.com/rust-lang/rustlings/blob/1aec7c1152e9b57142d5efaee4a9d95072b760bf/exercises/09_strings/strings4.rs#L24), we are instructed to "replace `placeholder(…)` with either `string_slice(…)` or `string(…)` depending on what you think each value is.".  Presumably that's the only type of change we're meant to make.  So, when I mistakenly replaced the `placeholder` on that line to `string_slice` (and changed all the other placeholders to what they need to be), I nominally pass the exercise (getting a green "Exercise done ✓"), but I'm still getting a warning:

```
warning: useless conversion to the same type: `&str`
  --> exercises/09_strings/strings4.rs:24:18
   |
24 |     string_slice("nice weather".into());
   |                  ^^^^^^^^^^^^^^^^^^^^^ help: consider removing `.into()`: `"nice weather"`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
   = note: `#[warn(clippy::useless_conversion)]` on by default
```

Now, I think the right thing for me to have done here is to use `string(...)` instead of `string_slice(...)`, so arguably this is an intended piece of the puzzle.  But it passes, and that doesn't seem right.  I'm not certain what the correct solution to this is (otherwise I'd probably submit a PR directly)... A few thoughts:

* I know I can put `#[warn(clippy::useless_conversion)]` right above line 24, and that makes the warning go away even though I've chosen the wrong alternative.
* On the flip side, I could put `#[deny(clippy::useless_conversion)]` (and/or this could be put into the exercise file, perhaps with a `// do not remove this line` comment after it), thus forcing this to be an error that needs to be fixed (hopefully by switching from `string_slice` to `string`).
* Similarly, we could put `#![deny(clippy::useless_conversion)]` (and some appropriate comment) up at the top of the file (this maybe seems like the best solution to me, though see below), also forcing things.

The biggest uncertainty I have about those last two options is that doing those adds a note such as the following to the compiler output:

```
note: the lint level is defined here
  --> exercises/09_strings/strings4.rs:2:9
   |
2  | #![deny(clippy::useless_conversion)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Now, maybe that's fine, and that's the best solution?  I'm not sure.  So, hopefully someone more experienced (I'm running into this because I'm new to rust, naturally) can figure out what the best solution should be.

The one thing that I think is probably a "bug" is that as-is, I can pass the exercise while still getting compiler warnings.  It seems to me that it should either be allowed without warnings, or give an error, and the latter seems more in-line with what this particular exercise is hoping to teach.

Thanks!

wyf027:
## Summary
- remove the ambiguous `"nice weather".into()` prompt from `strings4`
- remove the corresponding solution explanation and `#[allow(clippy::useless_conversion)]` example
- keep `strings4` focused on values that have one clear `String` vs `&str` answer before conversion traits are introduced

Fixes #2190.

## Test Plan
- `cargo fmt --check`
- `cargo check --manifest-path dev/Cargo.toml --bin strings4_sol`
- `cargo clippy --manifest-path dev/Cargo.toml --bin strings4_sol -- -D warnings`
- `cargo test --manifest-path dev/Cargo.toml --bin strings4_sol`
- `cargo test`

### DIFF
--- a/exercises/09_strings/strings4.rs
+++ b/exercises/09_strings/strings4.rs
@@ -21,8 +21,6 @@ fn main() {
 
     placeholder("rust is fun!".to_owned());
 
-    placeholder("nice weather".into());
-
     placeholder(format!("Interpolation {}", "Station"));
 
     // WARNING: This is byte indexing, not character indexing.
--- a/solutions/09_strings/strings4.rs
+++ b/solutions/09_strings/strings4.rs
@@ -15,15 +15,6 @@ fn main() {
 
     string("rust is fun!".to_owned());
 
-    // Here, both answers work.
-    // `.into()` converts a type into an expected type.
-    // If it is called where `String` is expected, it will convert `&str` to `String`.
-    string("nice weather".into());
-    // But if it is called where `&str` is expected, then `&str` is kept as `&str` since no conversion is needed.
-    // If you remove the `#[allow(…)]` line, then Clippy will tell you to remove `.into()` below since it is a useless conversion.
-    #[allow(clippy::useless_conversion)]
-    string_slice("nice weather".into());
-
     string(format!("Interpolation {}", "Station"));
 
     // WARNING: This is byte indexing, not character indexing.


senekor:
Stop spamming us with useless AI garbage.

PR #2411, opened 2026-06-14
Auto-Move on

ActionsPerMinute:
feature to allow rustlings to automatically move on to the next exercise, so the flow is not interrupted

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -26,6 +26,10 @@ pub struct Args {
     /// Only use this if Rustlings fails to detect exercise file changes
     #[arg(long)]
     pub manual_run: bool,
+    /// Automatically move on to the next exercise after a correct solution.
+    /// Can be toggled at runtime with `a` in the watch mode
+    #[arg(long)]
+    pub auto_move: bool,
 }
 
 #[derive(Subcommand)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,7 +118,7 @@ fn main() -> Result<ExitCode> {
                 )
             };
 
-            watch::watch(&mut app_state, notify_exercise_names)?;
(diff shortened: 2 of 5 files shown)


senekor:
I'm sceptical if this is a good idea. It's easy to remove or comment out lines such that an exercise compiles, even though it is not yet solved.

ActionsPerMinute:
> I'm sceptical if this is a good idea. It's easy to remove or comment out lines such that an exercise compiles, even though it is not yet solved.

I mean, that's true for all rustlings, you can go through the whole thing commenting out the asserts. Whether it auto moves to the next exercise is orthogonal to that fact. You can do that with this feature or without this feature. If you wanted to solve that, you'd put the asserts in a different file from where the exercises were.

For those that want to do the exercises properly, this streamlines the experience.

senekor:
I mean accidentally. If the auto-move is turned on and the exercise compiles accidentally, or because the user is trying something out by "cutting" non-compiling lines and pasting them somewhere else, rustlings would move on to the next exercise, which is not what the user wants. Getting back to the previous exercise is more tedious.

I agree that we don't need to worry about people cheating at the exercises.

ActionsPerMinute:
So, you have to save the file; or at least the way I'm using rust you have to..  You'd do all the moving in the editor, then the saving of the file would be like the "submit" of the attempt. Rustlings only evaluates when the file is saved, not during the file edit. At least, it does for me, I don't know if other editors continually save all changes

mo8it:
Closing because I agree with @senekor about accidential skipping.

Another source of confusion:
You solve an exercise and Rustlings automatically moves to the next one but you don't notice it and start questioning why it is not solved yet. Some exercises are similar so they will have similar errors. They also have similar names. So it is easy to miss.

PR #2414, opened 2026-06-19
fix: standard English grammar/phrasing

hippietrail:
In `hashmaps3.rs` I always stumble and find the task hard to understand. Being a native English speaker and language nerd I find the nonnative wording in that particular task to be a further distraction from trying to grok the problem.

English noun compounds only use a plural for the adjunct in rare exceptions (though it's becoming more common). "train station", "car park", and "score table" sound more idiomatic than "trains station", "cars park", or "scores table". Though "table of scores" is perhaps even more idiomatic.

There's also a nonnative English use of the simple past referring to a topic we haven't dealt with yet that I may as well fix while I'm here.


senekor:
Can you also update the solution?

PR #2417, opened 2026-06-29
Add --auto-next flag to skip the post-success prompt in watch mode

electron224:
After passing an exercise, the watch loop currently waits for `n` before advancing. Add an opt-in `--auto-next` flag that renders the "Exercise done" screen for ~1.5s, then auto-advances to the next pending exercise and runs it. Auto-advance only fires on success a failing exercise still stops at the prompt.

The `n:next` key is hidden from the prompt when --auto-next is on, but the user can still press `n` to skip ahead manually. 
--auto-next is compatible with --manual-run: pressing `r` triggers a check and on success the loop auto-advances the same way.


senekor:
closing for the same reason as #2411

PR #2418, opened 2026-07-06
Clarify options2 exercise comments

wcy6457:
## Summary

- Clarify that the first TODO should bind the inner `Option` value to `word` with `if let`.
- Explain what `Vec::pop()` does and why popping from `Vec<Option<i8>>` produces nested `Option` values.
- Add a short description of the intended `while let` loop behavior.

## Testing

- `cargo fmt --all --check`

I also tried broader local checks, but they were blocked by unrelated/current local issues: `cargo dev check --require-solutions` stopped on `exercises/03_if/if1.rs` metadata, and `cargo test --workspace` on Windows stopped because `cmd::tests::test_run_cmd` could not find an `echo` program.

## Disclosure

This PR was prepared with assistance from OpenAI Codex.

### DIFF
--- a/exercises/12_options/options2.rs
+++ b/exercises/12_options/options2.rs
@@ -9,7 +9,8 @@ mod tests {
         let target = "rustlings";
         let optional_target = Some(target);
 
-        // TODO: Make this an if-let statement whose value is `Some`.
+        // TODO: Use an if-let statement to bind the value inside
+        // `optional_target` to `word`.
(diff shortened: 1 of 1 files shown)


senekor:
> This PR was prepared with assistance from OpenAI Codex.

Thanks for the heads-up. If it wasn't worth your time to write it, it's not worth our time to read it.

PR #2420, opened 2026-07-17
Clarify Append command description in quiz2

sobhanmohammadi-dev:
## Description

Clarifies the documentation for the `Append` command in quiz2.

The previous wording was slightly ambiguous about where the string "bar" is appended.

## Changes

- Updated the Append command description to specify that "bar" is appended to the end of the string.

## Testing

Not applicable (documentation-only change).


senekor:
At least in the context of programming, "append" always means "at the end" and "prepend" is for "at the start".

Issue #2424, opened 2026-07-31
`rustlings dev check` falsely reports "doesn't contain any tests" for exercises with CRLF line endings


SandmeyerX:
## Environment

- `rustlings --version`: 6.5.0
- OS: Windows 11

## Description

Running `rustlings dev check` fails with:
```
Running all exercises to check that they aren't already solved...
Error: The file exercises/09_strings/strings3.rs doesn't contain any tests. If you don't want to add tests to this exercise, set test = false for this exercise in the info.toml file
```
Even though exercises clearly contains some tests, and `cargo test`) works fine.

The culprit is a literal string match:
https://github.com/rust-lang/rustlings/blob/734461f2fb8c7bb8403f4a9bd1fc7f983d32860b/src/dev/check.rs#L121

On files with Windows line endings the actual content is `#[test]\r\n`, which never matches `#[test]\n`.
Rust itself accepts CRLF as a valid line terminator, so it's could be considered as a false positive, not an exercise problem.

## Proposed fix
Normalizing CRLF to LF before matching.

I'll open a PR with the implementation soon.

PR #2425, opened 2026-07-31
Fix `dev check` rejecting exercises with CRLF line endings

SandmeyerX:
## Problem

On Windows (or whenever an editor saves an exercise with CRLF line endings),
`rustlings dev check` fails with:
```
The file `exercises/XX/xxx.rs` doesn't contain any tests.
```
even though the exercise clearly contains `#[test]` functions.

## Solution

Normalize CRLF to LF when checking whether an exercise contains tests, and add some unit tests.

### DIFF
--- a/src/dev/check.rs
+++ b/src/dev/check.rs
@@ -416,7 +416,8 @@ mod tests {
 
     #[test]
     fn detects_tests_in_files_with_lf_line_endings() {
-        let content = "fn main() {}\n\n#[cfg(test)]\nmod tests {\n    #[test]\n    fn it_works() {}\n}\n";
+        let content =
+            "fn main() {}\n\n#[cfg(test)]\nmod tests {\n    #[test]\n    fn it_works() {}\n}\n";
         assert!(exercise_has_tests(content));
     }
 }
--- a/src/dev/check.rs
+++ b/src/dev/check.rs
@@ -118,7 +118,7 @@ fn check_info_file_exercises(info_file: &InfoFile) -> Result<HashSet<PathBuf>> {
             );
         }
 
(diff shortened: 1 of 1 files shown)


SandmeyerX:
Close #2424

senekor:
Can we just omit the line ending? i.e. just check for `#[test]` instead of `#[test]\n`. Not sure, haven't thought too much about it.

SandmeyerX:
> Can we just omit the line ending? i.e. just check for `#[test]` instead of `#[test]\n`. Not sure, haven't thought too much about it.

@senekor  Thanks for the review! I'd lean toward keeping the `\n` check - reasoning:
 
The trailing newline was actually added deliberately in [e74f2a4](https://github.com/rust-lang/rustlings/commit/e74f2a4274e22faf050a1229b4526ceb53b55479) — the check originally was a bare contains("#[test]"), and I suppose the `\n` was introduced to avoid false positives where `#[test]` merely appears in prose, e.g.:
```
// TODO: write a #[test] function here
```
With a bare `#[test]` match, that comment would make the exercise count as "containing tests".

Maybe "More means better".

senekor (approved):
I see, makes sense. Thanks for digging. Looks good to me.

mo8it:
Reverted because I don't like the solution with `replace`. It has to allocate a `String` and replace all occurrences. A better solution would be `find` and then checking whether the found occurrence is followed by a newline.
Anyway, although checking for a newline is more accurate, I think that just checking for `#[test]` should be enough (I added the newline check originally).

mo8it:
Fixed in https://github.com/rust-lang/rustlings/commit/90658603902e7588c650ceb5ac0b26e31aa543af

PR #2428, opened 2026-08-01
Allow processing for hints and messages in the info file

### LINKED ISSUE #2426: `dev check` fails to parse `info.toml` if a multi-line string contains a backslash
## Environment

- `rustlings --version`: 6.5.0
- OS: Windows 11

## Description
I'm the maintainer of [rustlings-zh-cn](https://github.com/SandmeyerX/rustlings-zh-cn), the Simplified Chinese translation of rustlings, and just spent way too long debugging this, so I figured I'd write it up in case it helps someone else.

I built the latest `main` locally (I had a PR open for the CRLF test check false positive, #2425) and ran `rustlings dev check`:

```
Error: Failed to parse the `info.toml` file

Caused by:
    TOML parse error at line 23, column 17
       |
    23 | final_message = """+----------------------------------------------------+
       |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    invalid type: string "+----------------------------------------------------+...", expected a borrowed string
```

It worked fine with the previous version rustlings, but not now.

The error message `expected a borrowed string` is pretty useless on its own — it doesn't tell you *where* or *why*. 

After digging around I found that commit [0cbcb89](https://github.com/rust-lang/rustlings/commit/0cbcb8964c434066826b65633f2b208d69c334d3) upgrade [toml crate](https://docs.rs/toml/latest/toml/) from `0.9` to `1.0` , and changed the `info.toml` fields to `&'static str`.
But with `toml 1.x` a string can only be borrowed if it doesn't need any processing. 

Multi-line strings like `"""..."""` get unescaped, so as soon as they contain something like `\\`, the value gets allocated and parsing fails. 

My `welcome_message` / `final_message` contain ASCII art with backslashes: 
<details>
<summary>click for details</summary>

```
'                      __  ___                 
'     _______  _______/ /_/ (_)___  ____ ______
'    / ___/ / / / ___/ __/ / / __ \\/ __ `/ ___/     ~
'   / /  / /_/ (__  ) /_/ / / / / / /_/ (__  )     ___
'  /_/   \\__,_/____/\\__/_/_/_/ /_/\\__, /____/     ['_']  
'                                /____/       /=) zh-CN (=\\  
'                                                 -----
'                                                 || ||
'                                                 ~~ ~~
```
```
+----------------------------------------------------+
        ~
      ['_']  
  /=) zh-CN (=\\  >>干得漂亮，休息一会吧~
      -----
      || ||
      ~~ ~~
```

</details>
so mine hits this immediately.

## Workaround

Workaround that works for us: write those two messages as `'''...'''` literal strings. 

Literal strings never process escapes, so the backslashes stay literal and everything parses again.

## Proposed fix

My two cents on fixing it properly (either would be great):

1. Make parsing accept escaped multi-line strings. E.g. deserialize the message fields through a helper that falls back to owned strings when borrowing isn't possible (or use `Cow<'static, str>`). That would keep old `info.toml` files working, including what `dev new` generates.

2. At the very least, improve the error and document the breaking change. The [CHANGELOG](https://github.com/rust-lang/rustlings/blob/main/CHANGELOG.md) or [Website](https://rustlings.rust-lang.org/community-exercises/) doesn't mention this change, and the error message gives no hint about what to do — I only found the workaround by reading the `toml` source.

Also, maybe worth adding more tests. Right now CI can't catch this because the official `info.toml` never has backslashes in its strings.

I know community-exercises developer aren't the main audience, but this is the second time in a day that `dev check` broke (first the CRLF thing, now this). 
A little test coverage on the `info.toml` parsing side would go a long way for community repos.

Happy to help with any of the above if maintainers want.

mo8it:
Closes https://github.com/rust-lang/rustlings/issues/2426

### DIFF
--- a/rustlings-macros/src/lib.rs
+++ b/rustlings-macros/src/lib.rs
@@ -16,16 +16,8 @@ struct InfoFile<'a> {
 
 #[proc_macro]
 pub fn include_files(_: TokenStream) -> TokenStream {
-    // Remove `\r` on Windows
-    let info_file = String::from_utf8(
-        include_bytes!("../info.toml")
-            .iter()
-            .copied()
-            .filter(|c| *c != b'\r')
-            .collect(),
-    )
-    .expect("Failed to parse `info.toml` as UTF8");
-    let exercises = toml::de::from_str::<InfoFile>(&info_file)
+    let info_file = include_str!("../info.toml");
+    let exercises = toml::de::from_str::<InfoFile>(info_file)
         .expect("Failed to parse `info.toml`")
         .exercises;
 
--- a/src/app_state.rs
+++ b/src/app_state.rs
@@ -54,7 +54,7 @@ pub struct AppState {
     exercises: Vec<Exercise>,
     // Cache the number of done exercises to avoid iterating over all exercises every time.
     n_done: u32,
-    final_message: &'static str,
+    final_message: String,
     state_file: File,
     // Preallocated buffer for reading and writing the state file.
     file_buf: Vec<u8>,
@@ -66,8 +66,8 @@ pub struct AppState {
 
 impl AppState {
     pub fn new(
-        exercise_infos: Vec<ExerciseInfo>,
-        final_message: &'static str,
+        exercise_infos: Vec<ExerciseInfo<'static>>,
+        final_message: String,
         editor: Option<Editor>,
         vs_code_term: bool,
     ) -> Result<(Self, StateFileStatus)> {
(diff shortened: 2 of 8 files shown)


rustlings-macros/src/lib.rs
```diff
@@ -16,16 +16,8 @@ struct InfoFile<'a> {
 
 #[proc_macro]
 pub fn include_files(_: TokenStream) -> TokenStream {
-    // Remove `\r` on Windows
-    let info_file = String::from_utf8(
-        include_bytes!("../info.toml")
-            .iter()
-            .copied()
-            .filter(|c| *c != b'\r')
```
senekor, line 24: I suppose this isn't needed, because the `toml` crate already handles different line endings, right?

senekor (approved):
IIUC this is a refactor preparing for future changes. Looks good.

rustlings-macros/src/lib.rs
mo8it: This was required because deserialization with `\r\n` allocates a `String` and would raise an error when trying to deserialize to `&'a str`. Now that we have changed it to `String`, `toml` takes care of it.

mo8it:
> IIUC this is a refactor preparing for future changes.

Not preparing for future changes. What suggests that?

senekor:
I guess then I don't understand the purpose of this change? Is there a change in behavior? What's the meaning of the commit message "Allow processing for hints and messages"?

senekor:
Oh sorry, I missed the referenced issue. I didn't read carefully enough... ignore my questions :)

PR #2429, opened 2026-08-02
Rework Check-All

### LINKED ISSUE #2394: Text User Interface: keep session open after "check all" completes
When using the Rustlings terminal interface, pressing c (check all) currently closes the session once all exercises are checked.
I’d like to suggest keeping the interface open after completion so users can immediately review progress.
Additionally add a prompt that returns to the main Rustlings interface.

mo8it:
Closes https://github.com/rust-lang/rustlings/issues/2394

Reworked the visualization of the check-all progress. Instead of showing the exercise numbers, an exercise is shown with "done" or "pending" after being checked. To see the output, run `cargo r -- check-all`.

I also removed the manual input `c` to trigger check-all. It was offered as an option to avoid hitting `n` multiple times when the state file is out of sync (e.g. user continues on a different machine without syncing the state file). Instead of a manual trigger, check-all is now triggered automatically when the user hits `n` and the next exercise is done although it was not marked as such. This means that at least one exercise is done and not marked as such and check-all will fix possible others.

Rather orthogonal, but I did it on the way: I added a timeout to all commands in case a user writes a program which does not terminate. Currently, it is set to 30sec. What do you think about that value? @senekor

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -441,6 +441,7 @@ dependencies = [
  "shlex",
  "tempfile",
  "toml",
+ "wait-timeout",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


senekor (approved):
The timeout for exercises seems reasonable... but practically, how useful are 30 seconds? I expect most people to hit `Ctrl+C` before that. Making the timeout too tight seems risky, I don't know how long we should expect compilation to take on weak / old hardware... especially considering we want to add exercises with dependencies like tokio.

mo8it:
We might need to add either a global configuration option or even allow one for every exercise, especially for community exercises. But I would wait for someone to complain :P

PR #2446, opened 2026-09-06
Improve Windows build process in build.rs

gugu8intel-i9:
Update the error handling for copying Cargo.toml on Windows.


mo8it:
I think the error was ignored because it occurs when the file already exists or something like that. I don't see a need to improve it here since you will get an error later anyway.

senekor:
`std::fs::copy` will overwrite existing files, so the error must be something else. @gugu8intel-i9 should've said what their actual problem was though.

PR #2452, opened 2026-09-21
fix parsing of cargo locate-project output

senekor:
Deserializing the output of `cargo locate-project --workspace` failed for a windows user, with the error message from serde that `root` is a string and not a path. Hopefully, using a string directly will also work on Windows.


senekor:
Hi @mo8it, opening this as draft, since I haven't (yet) tested on windows if this actually fixes anything. (I've had a student of my course run into this problem.) Just wondering if you have some early thoughts on this.

This `root` value is only used in ways where a string also works, so this compiles (and works on linux). That being said, it's used in a place where `AsRef<Path>` is expected, so there will still be a conversion from string to path, just a bit later. Maybe it will fail there 🤔

I'll try to setup a windows VM to test this.

mo8it:
Cargo seems to convert it to a string anyway:
https://github.com/rust-lang/cargo/blob/cc74001909309fff5322b4d91c29c31ef6139df7/src/bin/cargo/commands/locate_project.rs#L54-L61

While reading about the command, `--message-format plain` makes more sense here. This avoids serialization and deserialization, a possible source of deviation. Can you use it in this PR?

senekor:
> While reading about the command, `--message-format plain` makes more sense here.

Agreed, this is nicer. The only stumbling block was the need to trim the whitespace from the end.

src/init.rs
```diff
@@ -66,15 +60,15 @@ pub fn init() -> Result<()> {
...
         }
 
-        let workspace_manifest =
-            serde_json::de::from_slice::<CargoLocateProject>(&locate_project_output.stdout)
-                .context(
-                    "Failed to read the field `root` from the output of `cargo locate-project …`",
-                )?
-                .root;
+        let workspace_manifest = {
+            let mut stdout = String::try_from(locate_project_output.stdout)
```
mo8it, line 64: `locate_project_output.stdout.pop()` should be enough for trimming here, right? The newline is from the println macro in Cargo which adds `\n`.

mo8it (changes requested):

src/init.rs
senekor: Maybe? Could be `\r\n` on Windows, right? Using `trim_end` seems a little more robust to me.

senekor:
I was able to confirm that this actually fixes the issue on Windows using a VM.

